PR #228, opened 2026-04-14
Detect BYOS source identity drift and S3 partition-key cutover

nethalo:
Closes #196, closes #198.

## Summary

Two sibling integrity gaps in the BYOS agent — both cause events to be silently misattributed to the wrong `bintrail_id`:

- **#196**: `SourceIdentity` was captured once at agent startup. If the source MySQL restarted with a regenerated `auto.cnf`, failed over behind a VIP, or its DSN resolved to a different instance, the agent kept stamping the stale `@@server_uuid` on every `MetadataRecord`. The SaaS then resolved events to the prior `bintrail_servers` row and blended two servers under one `bintrail_id` with no operator signal.
- **#198**: `serverIDStr` used as the S3 partition key is the UUID `bintrailID` when `--index-dsn` is set and the numeric `--server-id` otherwise. An upgrade across that cutover silently splits payload objects across two S3 prefixes; queries that span the upgrade miss pre-upgrade data.

### #196 — periodic identity re-capture with hard-fail
- `byosFlushConfig.sourceIdent` is now `*atomic.Pointer[byos.SourceIdentity]` (concurrent-safe between the stream goroutine that re-captures and the flush goroutine that stamps records).
- `byosStreamLoop` runs a 60s `identityTicker` that calls the new `checkSourceIdentity`:
  - UUID stable → refresh the pointer (host/port/user may shift under CNAME changes).
  - UUID changed → return an actionable error; stream loop exits, agent exits non-zero.
  - Transient DB error → `slog.Warn` and let the next tick retry; do not tear down the stream.

### #198 — partition-key freeze marker
- New `byos.EnsurePartitionKey(ctx, backend, serverID)`. On first run it writes `.bintrail-partition-key` (JSON: `server_id` + `first_seen`) to the storage backend. On subsequent runs it reads the marker: match → no-op, mismatch → hard-fail with a message that explains the cutover and the operator's options (migrate S3 objects to the new prefix, or restore the prior config).
- Wired into `runAgent` right after the S3 backend is constructed, before `NewPayloadWriter`.
- Additional `slog.Warn` in the BYOS-without-`--index-dsn`+S3 path so the chosen partition key appears in the startup banner (defense in depth when no marker exists yet).

### Tests
- `cmd/bintrail/agent_identity_test.go` (sqlmock): stable, change-detected, transient-error-tolerated, host-only-update.
- `internal/byos/partition_key_test.go` (reuses the in-memory `memBackend` from `payload_test.go`): first-run, match no-op, mismatch error.

## Out of scope
- Retry-with-backoff inside `loadSourceIdentity` — called out in #196 as an independent improvement.
- Sentinel "identity discontinuity" records for the SaaS side (requires coordinated change in `nethalo/dbtrail`).
- Automatic migration of S3 objects between prefixes — the error message points the operator at the options.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./... -count=1` (full unit suite, all packages green)
- [x] New sqlmock tests for `checkSourceIdentity` pass
- [x] New `memBackend` tests for `EnsurePartitionKey` pass
- [ ] Manual #198 repro (optional): run agent with `serverID=A` against S3 → marker written; restart with `serverID=B` → agent hard-fails at startup with the mismatch message
- [ ] Manual #196 repro (optional): run agent with a short `identityInterval`, `rm auto.cnf` + restart source MySQL → agent exits cleanly with the change-detected error within one tick

### DIFF
--- a/cmd/bintrail/agent.go
+++ b/cmd/bintrail/agent.go
@@ -639,32 +639,39 @@ func byosStreamLoop(ctx context.Context, events <-chan parser.Event, buf *buffer
 	}
 	identityFailures := 0
 
-	flushBatch := func() {
+	// flushBatch drains the pending batch to the in-memory buffer and, if
+	// sinks are configured, to the metadata/payload destinations.
+	//
+	// Returns an error only for programmer errors detected during the sink
+	// flush (nil sourceIdent pointer). Transient sink failures are logged
+	// inside flushToSinks and surfaced via flushPipelineState.
+	flushBatch := func() error {
 		if len(batch) == 0 {
-			return
+			return nil
 		}
 
-		// Always insert into buffer.
 		buf.Insert(batch)
 		slog.Debug("BYOS batch flushed to buffer", "events", len(batch), "buffer_size", buf.Len())
 
 		// Split and flush to sinks if configured. Skip on shutdown —
 		// the sinks would fail immediately with context.Canceled and
 		// produce misleading error logs.
+		var flushErr error
 		if fc != nil && (fc.metaClient != nil || fc.payloadWriter != nil) && ctx.Err() == nil {
-			flushToSinks(ctx, batch, fc)
+			flushErr = flushToSinks(ctx, batch, fc)
 		}
 		if fc != nil && fc.state != nil {
 			fc.state.setBufferStats(buf.Len(), buf.ApproxBytes(), buf.SizeEvictions())
 		}
 
 		batch = batch[:0]
+		return flushErr
 	}
 
 	for {
 		select {
 		case <-ctx.Done():
-			flushBatch()
+			_ = flushBatch()
 			return nil
 
 		case <-evictTicker.C:
@@ -677,18 +684,19 @@ func byosStreamLoop(ctx context.Context, events <-chan parser.Event, buf *buffer
 			}
 
 		case <-flushTicker.C:
-			// Timer-based flush to bound metadata latency.
-			flushBatch()
+			if err := flushBatch(); err != nil {
+				return err
+			}
 
 		case <-identityTickerC:
 			if err := checkSourceIdentity(ctx, fc.sourceDB, fc.sourceIdent, &identityFailures); err != nil {
-				flushBatch()
+				_ = flushBatch()
 				return err
 			}
 
 		case ev, ok := <-events:
 			if !ok {
-				flushBatch()
+				_ = flushBatch()
 				return nil
 			}
 
@@ -699,7 +707,9 @@ func byosStreamLoop(ctx context.Context, events <-chan parser.Event, buf *buffer
 
 			batch = append(batch, ev)
 			if len(batch) >= batchSize {
-				flushBatch()
+				if err := flushBatch(); err != nil {
+					return err
+				}
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


nethalo:
Addressed review feedback in `f94bfc7`. Changes organized by the concern each resolves.

## Bugs fixed
- **DSN errors misclassified as transient**: `checkSourceIdentity` now queries `@@server_uuid` directly instead of routing through `loadSourceIdentity`. Host/port/user come from the startup DSN flag and can't change at runtime, so re-parsing on every tick was dead weight that also hid real DSN-format errors behind a WARN-forever loop.
- **Unbounded transient-error tolerance**: added `maxIdentityFailures=10` consecutive-failure cap. At the default 60s cadence, ~10 minutes of contiguous DB failures aborts the stream — catches revoked-creds / dropped-user without being trigger-happy on a network blip.
- **Nil-pointer silent fallback in `flushToSinks`**: replaced with hard-fail (Error log + pipeline marked degraded). A nil `sourceIdent` is a programmer error, not a runtime condition.
- **Misleading startup warning**: downgraded to Info and rephrased. Old text told the operator objects "will not be queried unless manually migrated", implying the agent runs — but `EnsurePartitionKey` actually hard-fails on a mismatched marker, so the scenario the warning described can't happen silently.

## Partition-key marker hardening (#198)
- Added `Version int` field for forward compatibility — the marker lives in customer S3 forever.
- Separated four distinct failure modes, each with actionable errors that explicitly warn against `rm`-as-a-fix (which would re-arm the #198 cutover):
  - corrupt/unparseable marker
  - empty-body marker (partial upload)
  - marker with empty `server_id`
  - real `server_id` mismatch
- Malformed markers are preserved for operator inspection — no silent overwrite.
- Read cap via `io.LimitReader(rc, 64*1024)`.
- Unexported `partitionKeyMarker` constant (never referenced outside the package).

## Comment rot
- Dropped the "flush goroutine" references on `byosFlushConfig` and `flushToSinks`: flush runs synchronously inside `byosStreamLoop` today; the atomic is defensive for a future refactor, not for a current race.
- Removed the dead `"(via resolveServerIdentity rule 2)"` cross-reference — `serverutil.go` has no numbered rules.

## Test coverage gaps closed
- **`TestByosStreamLoop_IdentityChangeAbortsLoop`** — the ticker integration was entirely untested. Exercises the `identityTickerC` arm end-to-end with a real buffer, asserts pending events are drained via `flushBatch` before the loop returns, and that the error surfaces to the caller.
- **`TestCheckSourceIdentity_BoundedRetryAborts`** — 10 failing queries, 11th aborts.
- **`TestCheckSourceIdentity_ContextCanceledSilent`** — shutdown path doesn't log spurious warnings or count against the failure budget.
- **`TestCheckSourceIdentity_NilPointerIsBug`** — uninitialized pointer hard-fails.
- **`TestEnsurePartitionKey_MalformedMarker`** — corrupt marker preserved, not overwritten.
- **`TestEnsurePartitionKey_EmptyMarker`**, **`TestEnsurePartitionKey_EmptyServerIDInMarker`**, **`TestEnsurePartitionKey_BackendErrors`** (Exists / Get / Put via a `failingBackend` wrapper).
- `TestEnsurePartitionKey_FirstRun` now locks the marker key to `.bintrail-partition-key` (wire contract).
- Dropped `TestCheckSourceIdentity_HostUpdateStable` — it exercised behavior that can't happen at runtime.

## Deferred to follow-ups
- `byosFlushConfig` identity sub-struct refactor (type-design #3).
- Two-agent first-write race via S3 `If-None-Match`.
- `defer rc.Close()` error logging.

Full unit suite green (22 packages).

nethalo:
Addressed second-pass review in `d0e2c40`.

## Behavior changes
- **`flushToSinks` now tears down on nil `sourceIdent`** (silent-failure-hunter, type-design). Previously the two nil-guard branches logged Error + marked degraded and let the stream keep running — inconsistent with `checkSourceIdentity`'s BUG path. `flushToSinks` returns `error`, `byosStreamLoop`'s `flushBatch` closure propagates, and select arms tear down on programmer errors. Shutdown paths (`ctx.Done`, events channel close) explicitly discard with `_ = flushBatch()` because we're already exiting.
- **`partitionKeyRecord.Version` is now load-bearing** (type-design). `EnsurePartitionKey` refuses markers written by a future agent (`rec.Version > partitionKeyMarkerVersion`) with a message directing the operator to upgrade. Previously the field was aspirational.

## Documentation
- **Flapping-source limitation documented** in `checkSourceIdentity` godoc (silent-failure-hunter). The consecutive-failure counter resets on any success; rolling-window detection is out of scope and deferred until a real failure mode demands it.

## Tests added
- `TestFlushToSinks_NilSourceIdentPointer` — nil `*atomic.Pointer` branch.
- `TestFlushToSinks_NilIdentSnapshot` — pointer exists but no Store yet; asserts the distinct "not initialized" message.
- `TestCheckSourceIdentity_DeadlineExceededSilent` — mirrors the Canceled test against `context.DeadlineExceeded`; catches a regression that would drop the `DeadlineExceeded` short-circuit.
- `TestEnsurePartitionKey_NewerVersionRejected` — locks in the new Version guard.

## Not addressed (skipped as nits)
- `FirstSeen` inline "always UTC on write" comment (comment-analyzer) — low value, leaving.
- `io.LimitReader(+1 byte)` explicit truncation detection (silent-failure-hunter) — current error is acceptable.
- `byosFlushConfig` identity sub-struct (type-design) — single call site, deferred until a third consumer or invariant appears.

Full unit suite green. Three commits total on the branch.

Issue #297, opened 2026-05-23
docs(time-travel-sql): rule count drift — doc says 3 routing rules, proxysql-config emits 4 since #294


nethalo:
## Summary

`docs/time-travel-sql.md` repeatedly refers to "three" routing rules (`990001-990003`) for the ProxySQL config. Since PR #294 (`feat(shim): accept /*+ DBTRAIL_AT='<ts>' */ hint-comment form`), `bintrail proxysql-config` emits **four** rules — the fourth (`990004`) matches the `/*+ DBTRAIL_AT=... */` hint-comment shape.

`CLAUDE.md` has the same drift: it documents `proxysql-config` as emitting rule ids `990001-990003`.

## Repro

Generated `proxysql-setup.sql` with `bintrail v0.7.6` (commit e2c91fa) and inspected the live rules after applying:

```sh
mysql -u admin -padmin -h 127.0.0.1 -P 6032 -e \
  "SELECT rule_id, match_pattern, destination_hostgroup FROM runtime_mysql_query_rules WHERE rule_id BETWEEN 990001 AND 990005;"
```

Output:
```
rule_id  match_pattern             destination_hostgroup
990001   \\b_flashback\\.          991
990002   \\b_diff\\.               991
990003   \\b_snapshot\\.           991
990004   /\\*\\+\\s*DBTRAIL_AT     991
```

## Doc locations to fix

`docs/time-travel-sql.md`:
- Line ~132 (paragraph before "Verify ProxySQL accepted the config"): mentions "rule IDs (990001-990003)" — update to `990001-990004`.
- Line ~268 (Troubleshooting `_flashback.t doesn't exist`): SQL example queries `BETWEEN 990001 AND 990003` — update to `990004`.
- Line ~313 (Troubleshooting "Operator already has users in hostgroup 990"): mentions "rule IDs (990001-990003)" — update.

`CLAUDE.md`:
- The `proxysql-config` row in the command table mentions `rule_ids 990001-990003` — update.

The generated `proxysql-setup.sql` itself has the correct header listing all four rules; only the docs lag.

## Why this matters

Operators who re-apply `proxysql-config` and then inspect rules per the doc's Troubleshooting recipe would see 4 rows instead of the documented 3 and wonder whether something is misconfigured. Low impact but easy fix.

## Suggested labels

`documentation`

nethalo:
PR up at #301

Issue #276, opened 2026-05-07
shim: support full-table reconstruction in _flashback/_snapshot (no PK filter required)


nethalo:
## Context

`bintrail shim` currently rejects `_flashback.<table>` and `_snapshot.<table>` queries that don't filter by primary key. The parser at `internal/shim/parser.go:83` requires the shape:

```
SELECT * FROM _flashback.<table> AS OF '<ts>' WHERE <pk_col> = <value>
```

Anything without `WHERE <col> = <value>` returns *"malformed time-travel query"* to the client.

This was the right MVP scope — point-lookups are the common operational case ("what was this row at time X?") and they cost one indexed event-log query. But it leaves a real gap that surfaces the moment anyone tries the obvious thing:

```
mysql> SELECT * FROM _flashback.orders AS OF '2026-04-15 09:30';
ERROR ...: malformed time-travel query
```

Surfaced when a careful reader (PZ on the announcement thread) walked through the architecture and asked "how do I get correct full-table AS OF results?". The honest answer today is "you don't, via the shim". This issue closes that gap.

## Use cases this unlocks

1. *"What did the entire orders table look like before the bad migration ran?"* — full-table reconstruction at a known good timestamp.
2. *"Diff today's orders against last Monday's orders."* — feed both into a CSV / `EXCEPT` query / etc.
3. *"Restore a table to a prior state."* — already supported offline via `bintrail recover`, but operators ask for the interactive equivalent.

## Why the live MySQL is irrelevant

Worth pinning explicitly because it's the natural confused intuition: live MySQL represents *now*, not the past. A row that exists now but didn't exist at AS OF would be incorrectly included; a row deleted before AS OF would be missed. Reconstruction must happen entirely against `binlog_events` (plus archived Parquet), never merged with the live table.

## Algorithm

For each `(schema, table)` pair, walk every PK that ever existed in `binlog_events` with `event_timestamp <= AS OF`. For each PK, find the latest event ≤ AS OF. Emit rows where the latest event is INSERT or UPDATE (skip DELETE — the row didn't exist at AS OF).

In SQL terms, roughly:

```sql
SELECT row_after
FROM (
  SELECT
    pk_values,
    row_after,
    event_type,
    ROW_NUMBER() OVER (
      PARTITION BY pk_values
      ORDER BY event_timestamp DESC, event_id DESC
    ) AS rn
  FROM binlog_events
  WHERE schema_name = ? AND table_name = ?
    AND event_timestamp <= ?
[...]

PR #303, opened 2026-05-23
fix(shim): validate WHERE column matches table PK (#296)

### LINKED ISSUE #296: shim: WHERE on non-PK column returns wrong row (silent data-correctness bug)
## Summary

The shim's `_flashback` / `_snapshot` point-lookup path silently returns the **wrong row** when the user filters on a non-PK column whose value coincidentally matches a PK value. This is a correctness bug, not just a UX gripe — the wire response carries data the user did not ask for and may act on.

Discovered during an E2E walkthrough of `docs/time-travel-sql.md` against RDS MySQL 8.0.46 + bintrail v0.7.6 (commit e2c91fa) + ProxySQL 2.6.6.

## Repro

Source table:
```sql
CREATE TABLE myapp.orders (id INT PRIMARY KEY AUTO_INCREMENT, customer_id INT, status VARCHAR(32), ...);
INSERT INTO orders (customer_id, status, total) VALUES
  (100, 'pending', 19.99),  -- id=1, customer_id=100
  (200, 'shipped', 49.50),  -- id=2, customer_id=200
  (300, 'pending', 5.00);   -- id=3, customer_id=300
```

`bintrail_index.binlog_events.pk_values` after stream indexed the events:
```
pk_values event_type
"1"       INSERT
"2"       INSERT
"3"       INSERT
...
```

Through ProxySQL :6033 → shim:

```sql
-- The user expects: 0 rows (or 1064) — no order has customer_id=1
SELECT * FROM _flashback.orders AS OF '2026-05-23 18:20:13' WHERE customer_id=1;
```

Actual response:
```
id  customer_id  status  total  created_at           updated_at
1   100          paid    19.99  2026-05-23 18:20:06  2026-05-23 18:20:10
```

The shim returned the row where **id=1**, not where customer_id=1. The user's WHERE column was silently ignored; only the literal value `1` was used as a lookup key against `pk_values`.

## Root cause

`internal/shim/parser.go` captures the WHERE column into `PKColumn` (groups m[3]/m[4]/m[5] across the three regexes) and the value into `PKValue`. In `internal/shim/handler.go:367-405` (`runPointInTime`), `PKColumn` is used only to dispatch point-lookup vs full-table (handler.go:368). `PKValue` is then passed verbatim as `query.Options.PKValues=q.PKValue` (handler.go:387) and joined against `binlog_events.pk_values`, regardless of what column name the user typed.

A user writing `WHERE customer_id=1` gets the row with `pk_values="1"` — i.e. the row whose actual PK is 1, which on this table happens to have `customer_id=100`. The doc's Limitations section already promises "No JOINs, aggregations, or non-PK WHERE filters inside the shim", but the parser doesn't enforce that promise.

## Why this is dangerous, not just ugly

- The wire response is **schema-valid and plausible-looking** — there is no indicator that the WHERE was misinterpreted. An app or a human operator using this for forensics will treat it as authoritative.
- It only fails silently. There is no log line, no warning, no hint in the resultset metadata that the WHERE column was discarded.
- The collision condition is common: tables with INT PK starting at 1 will collide with any other INT column containing small numbers (foreign keys, counts, status enums, etc.).

A noisier variant — `WHERE customer_id=999999` (no collision) — returns an empty resultset, which an operator might also misinterpret as "no orders for that customer".

## Expected behavior

Two reasonable options, either matching the doc's intent:

1. **Reject the query with `ER_PARSE_ERROR` (1064)** when the WHERE column is not the table's PK. Error message: `"WHERE column must be the primary key of <schema>.<table> (expected <pk_col>, got <user_col>)"`. The shim already has access to PK via `schema_snapshots` (used by recovery).
2. **Map the user's column to the actual PK column at parse time** — only if the column matches the resolved PK column name, accept the query.

(1) is the simpler, safer choice and matches the existing "type, don't translate" stance of the parser elsewhere (e.g. 3-part naming → 1064).

## Notes on `_diff`

When I tried to repro on `_diff` with a non-default `SELECT` shape (`SELECT event_id, event_type FROM _diff.orders ... WHERE customer_id=2`), the parser returned 1064 — but only because the regex requires `SELECT *` literally. Reproing `_diff` with `SELECT *` likely has the same bug; I didn't test that explicitly.

## Environment

- bintrail v0.7.6 (commit e2c91fa, linux/amd64)
- ProxySQL 2.6.6 (Ubuntu 22.04 jammy repo)
- RDS MySQL 8.0.46, custom parameter group with `binlog_format=ROW`, `binlog_row_image=FULL`
- Index hosted on the same RDS instance, schema `bintrail_index`
- Streaming via `bintrail stream --server-id 4242 --start-file mysql-bin-changelog.000049 --start-pos 544 --schemas myapp`

## Suggested labels

`bug` `priority/p1` `area/shim`

nethalo:
## Summary

The shim parsed `_flashback` / `_snapshot` / `_diff` queries into a `TimeTravelQuery` whose `PKColumn` captured the user-supplied WHERE column, but only used that name to dispatch (single-row lookup vs full-table). The literal value (`PKValue`) was joined against `binlog_events.pk_values` regardless of column name, so `WHERE customer_id=1` against a table PK'd on `id` returned the row with **id=1** — a schema-valid resultset for a question the user never asked.

Repro is in #296. Discovered during the E2E walkthrough that also produced #297/#298/#299/#300.

This PR wires `validatePKColumn()` into `HandleQuery` once after `Parse` returns nil, so all four shapes share the validation site:

- `_flashback`, `_snapshot`, `_diff`
- the hint-comment form `SELECT /*+ DBTRAIL_AT='<ts>' */ * FROM …`

`validatePKColumn` reuses the existing `resolverCache` — same TTL, same sticky-fallback, same `Debug`/`Warn` log split as `columnOrderFor` — so production paths share one snapshot lookup.

## Design

**Reject branches (all 1064, single parse-error site):**
- WHERE column ≠ single-column PK → `"WHERE column must be the primary key of <schema>.<table> (expected <pk_col>, got <user_col>)"`
- Composite PK → `"<schema>.<table> has a composite primary key (<col1>, <col2>); WHERE <col> = <value> shape supports only single-column PKs"`
- Table has no PK declared → `"<schema>.<table> has no primary key declared in the indexed snapshot; cannot resolve <type> by PK"`

**Permissive branches (preserve `columnOrderFor`'s graceful-degradation contract — broken snapshot lookup must not turn a working query into a parse error):**
- `q.PKColumn == ""` (full-table reconstruction path, #276)
- `resolverFn == nil` (legacy `&Handler{}` tests)
- Resolver load fails (`ErrNoSnapshots` → Debug, other errors → Warn — same as `columnOrderFor`)
- Table not in latest snapshot (post-snapshot tables are common and self-fix on next `bintrail snapshot`)

Fixes #296

## Test plan

- [x] `go test ./internal/shim/... -count=1` — PASS (existing + 11 new subtests)
- [x] `go vet ./internal/shim/...` — clean
- [x] `go test ./internal/shim/... ./cmd/bintrail/... -count=1` — PASS

New `TestValidatePKColumnRejectsNonPKWhere` covers:
- `accept_pk_column_passes_validation`
- `reject_non_pk_column_returns_1064_with_actionable_message`
- `reject_composite_pk_naming_all_pk_columns`
- `reject_table_with_no_pk_declared`
- `permissive_when_table_missing_from_snapshot`
- `permissive_when_resolver_load_fails`
- `permissive_when_no_pk_column_in_query`
- `permissive_when_resolverFn_nil`
- `validates_snapshot_shape`
- `validates_diff_shape`
- `validates_hint_comment_shape`

### DIFF
--- a/internal/shim/handler.go
+++ b/internal/shim/handler.go
@@ -285,6 +285,17 @@ func (h *Handler) HandleQuery(qstr string) (*mysql.Result, error) {
 
 	q, perr := Parse(qstr, currentDB)
 	if perr == nil {
+		// Cross-cut PK validation (#296). Applied here, not inside each
+		// runX, so all four parsed shapes (TypeFlashback, TypeSnapshot,
+		// TypeDiff, and the hint-comment form which Parse normalises
+		// into TypeFlashback) share one validation site. Returns a
+		// *mysql.MyError typed at ER_PARSE_ERROR when the user's WHERE
+		// column does not match the table's real PK — the bug the
+		// issue describes was the shim silently using the literal value
+		// against pk_values regardless of column name.
+		if verr := h.validatePKColumn(q); verr != nil {
+			return nil, verr
+		}
 		switch q.Type {
 		case TypeFlashback, TypeSnapshot:
 			return h.runPointInTime(q)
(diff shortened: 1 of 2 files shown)


nethalo:
## E2E validation run

Built this PR's binary (commit 016e579, `v0.7.6-pr303`) on linux/amd64 via Docker, deployed to a fresh Ubuntu 22.04 EC2 + RDS MySQL 8.0.46 + ProxySQL 2.6.6 stack. Reproduced the original #296 false-positive and ran 7 cases through the patched shim via ProxySQL :6033 → shim :3308:

| # | Query | Expected | Got |
|---|-------|----------|-----|
| 1 | `WHERE id=1` (correct PK) | row with id=1 | ✅ row with id=1 |
| 2 | `WHERE customer_id=1` (the original bug) | 1064 | ✅ `ERROR 1064: WHERE column must be the primary key of myapp.orders (expected id, got customer_id)` |
| 3 | `WHERE customer_id=100` | 1064 | ✅ same error |
| 4 | `_snapshot ... WHERE customer_id=100` | 1064 | ✅ same error |
| 5 | `_diff ... WHERE customer_id=100` | 1064 | ✅ same error |
| 6 | `/*+ DBTRAIL_AT=... */ ... WHERE customer_id=100` | 1064 | ✅ same error |
| 7 | full-table `_flashback` (no WHERE) | 3 rows | ✅ 3 rows |

Tests cover: regression on the correct-PK path, the original repro (customer_id=1 collides with id=1), the same-named-value case across all four shapes, and the no-WHERE full-table path which the fix intentionally leaves untouched.

Source MySQL was a fresh RDS with `binlog_format=ROW`, `binlog_row_image=FULL` (via custom parameter group), populated with 3 orders, indexed via `bintrail stream` from current binlog position. Index schema snapshot has `myapp.orders.PKColumns=[id]`, so the validator's reject path was the one exercised — the permissive paths (no snapshot, table missing) have unit-test coverage and weren't hit on this run.

PR #321, opened 2026-05-24
fix(query): handle NULL binlog_file in result Scan (#318)

### LINKED ISSUE #318: [BUG P1] bintrail query: 'converting NULL to string is unsupported' on binlog_file column — Explorer in dbtrail dashboard broken
## Summary

`bintrail query` (which the dbtrail SaaS Explorer shells out to via the agent) fails when scanning any row where `binlog_file` is NULL in the index:

\`\`\`
failed to scan result row: sql: Scan error on column index 1, name \"binlog_file\": converting NULL to string is unsupported
\`\`\`

Symptom in the SaaS dashboard: clicking \"Run Query\" in the Explorer returns:
\`\`\`
Data plane error: {\"error\":\"failed to scan result row: sql: Scan error on column index 1, name \\\"binlog_file\\\": converting NULL to string is unsupported\"}
\`\`\`

## Where

`internal/query/query.go:385`:

\`\`\`go
if err := rows.Scan(
    &r.EventID, &r.BinlogFile, &r.StartPos, &r.EndPos, &r.EventTimestamp,
    &gtid, &connID, &r.SchemaName, &r.TableName, &r.EventType, &r.PKValues,
    &changedCols, &rowBefore, &rowAfter, &r.SchemaVersion,
); err != nil {
    return nil, fmt.Errorf(\"failed to scan result row: %w\", err)
}
\`\`\`

`r.BinlogFile` is declared as plain `string` at line 138 (`type ResultRow struct { ... BinlogFile string ... }`). When the column is NULL, the scan fails.

Other nullable columns in the same Scan already use `sql.NullString` / `sql.NullInt64` (`gtid`, `connID`). `binlog_file` was missed.

## Why NULL appears

The `bintrail_index.binlog_events` table allows NULL `binlog_file` for events that don't have a binlog source — snapshot baseline rows, schema_change records propagated from other sources, manually-imported events.

## Fix (1-line on the struct + minimal scan changes)

Either:

**Option A — change the struct field**:
\`\`\`go
type ResultRow struct {
    ...
    BinlogFile sql.NullString
    ...
}
\`\`\`
…and JSON-marshal it with a custom MarshalJSON or accessor.

**Option B — use a local NullString and copy on success** (matches the pattern already used for `gtid`):
\`\`\`go
var r ResultRow
var binlogFile sql.NullString
var gtid sql.NullString
// ...
if err := rows.Scan(&r.EventID, &binlogFile, &r.StartPos, ...); err != nil { ... }
if binlogFile.Valid {
    r.BinlogFile = binlogFile.String
}
\`\`\`

B is the smaller change and matches the surrounding code style.

## Impact

P1 — the dbtrail Explorer is unusable for any tenant with NULL `binlog_file` rows in their index. Every customer running real traffic hits this immediately once they touch the Explorer.

Cross-ref: [nethalo/dbtrail#1484](https://github.com/nethalo/dbtrail/issues/1484).

Caught while validating the Percona Live 2026 demo.

nethalo:
## Summary

- `internal/query/query.go` scanned `binlog_file` directly into a plain `string`, which crashed every query the moment the index contained a row with `NULL binlog_file` (snapshot baseline rows, schema_change records propagated from other sources, manually-imported events).
- Fix follows Option B from the issue: scan into a local `sql.NullString` and copy on `Valid`, mirroring the existing pattern for `gtid` and `connection_id`. `ResultRow.BinlogFile` stays plain `string` so JSON / CSV / table output is unchanged (empty string instead of `null`).
- Adds a `sqlmock`-based unit regression test covering the failing Scan path and a second test confirming the JSON formatter emits `""` for NULL rows.

Closes #318

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/query/... -count=1`
- [ ] `go test -tags integration ./internal/query/... -count=1` (not run — no MySQL container available in this worktree; unit test fully exercises the fix via sqlmock)

### DIFF
--- a/internal/query/query.go
+++ b/internal/query/query.go
@@ -378,16 +378,24 @@ func scanRows(rows *sql.Rows) ([]ResultRow, error) {
 	var results []ResultRow
 	for rows.Next() {
 		var r ResultRow
+		// binlog_file is nullable in practice: snapshot baseline rows,
+		// schema_change records propagated from other sources, and manually
+		// imported events have no binlog origin. Scan into NullString and copy
+		// on Valid so the empty string is preserved as "" in ResultRow / JSON.
+		var binlogFile sql.NullString
 		var gtid sql.NullString
 		var connID sql.NullInt64
 		var changedCols, rowBefore, rowAfter []byte
 
 		if err := rows.Scan(
-			&r.EventID, &r.BinlogFile, &r.StartPos, &r.EndPos, &r.EventTimestamp,
+			&r.EventID, &binlogFile, &r.StartPos, &r.EndPos, &r.EventTimestamp,
 			&gtid, &connID, &r.SchemaName, &r.TableName, &r.EventType, &r.PKValues,
 			&changedCols, &rowBefore, &rowAfter, &r.SchemaVersion,
 		); err != nil {
 			return nil, fmt.Errorf("failed to scan result row: %w", err)
+		}
+		if binlogFile.Valid {
+			r.BinlogFile = binlogFile.String
 		}
 		if gtid.Valid {
 			r.GTID = &gtid.String
--- a/internal/query/query_test.go
+++ b/internal/query/query_test.go
@@ -2,12 +2,14 @@ package query
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/DATA-DOG/go-sqlmock"
 	"github.com/dbtrail/bintrail/internal/parser"
 )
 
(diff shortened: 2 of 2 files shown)


nethalo:
Superseded by #319, which landed on `main` while this PR was being reviewed and addresses #318 more comprehensively:

- `internal/query/query.go` (`scanRows`) — same fix as here
- `internal/archive/archive.go` (rotate path) — same crash in the partition-archive Scan that this PR did not touch
- `internal/parquetquery/parquetquery.go` (DuckDB read-back) — same crash one layer down on archive read
- `cmd/bintrail/reconstruct.go` (gap-detection) — silent-degradation site exposed once the upstream Scan stopped crashing
- `internal/testutil/testutil.go` — repaired `connection_id` column drift in the test fixture

Closing as redundant. The post-merge review of this PR did flag the parquetquery and recover-side risks as follow-up suggestions, but #319 already shipped them as part of the same fix. No further action needed on #318.

PR #358, opened 2026-05-29
feat(shim): baseline-aware _snapshot point-in-time queries

### LINKED ISSUE #355: shim: make _snapshot AS OF reconstruct full tables via baseline lookup (today it is binlog-only)
## Summary

The shim's time-travel virtual schemas (`_flashback`, `_snapshot`, `_diff`) operate **only over the set of rows that have binlog activity** within the retained/indexed window. None of them reconstructs a full table. `_snapshot` is currently a literal synonym for `_flashback` — it does **not** apply deltas on top of a baseline.

As a result, a row that existed at the AS OF instant but was never touched within the indexed binlog window **silently does not appear** in the result. A user running `SELECT * FROM _flashback.orders AS OF '...'` gets the subset of rows that changed within retained binlog history, *not* a complete table snapshot.

True full-table point-in-time reconstruction already exists, but only in the offline `bintrail reconstruct` command (`internal/reconstruct/fulltable.go`), which seeds from a mydumper baseline Parquet snapshot and merges deltas via DuckDB. The shim never reaches that code path.

## Evidence (code-grounded)

- `internal/shim/handler.go:324-326` — `TypeFlashback` and `TypeSnapshot` both dispatch to the same `runPointInTime`.
- `internal/shim/parser.go:30-33` — comment confirms `_snapshot` is currently identical to `_flashback`, intended to integrate baseline lookups "for now they share an implementation."
- `runFullTable` (`handler.go:573-608`) issues the identical `FetchMerged` query with empty `PKValues` + `LimitPerPK:1`. `extractFullTableImages` (`handler.go:614-629`) builds the resultset only from `row_after` of non-DELETE events — no baseline seed.
- `internal/query/fetchmerged.go:131-200` — data sources are `binlog_events` (live) + archived `binlog_events` partitions only. The baseline-aware `query.FetchSnapshot` (`internal/query/snapshot.go`) is called only from `cmd/bintrail/query.go:323`, never from `FetchMerged` / the shim.
- `internal/reconstruct/fulltable.go` `ReconstructTable` is the real baseline+delta merge — reachable only from `cmd/bintrail/reconstruct.go`. The shim has zero references to the `reconstruct` package.

## Proposed work

Wire the `reconstruct` baseline-lookup path into the shim so that `_snapshot AS OF <ts>` (and the full-table `SELECT *` form) can answer for rows that never appeared in binlog events:

1. In `runFullTable` / `runPointInTime`, when a baseline is available (`reconstruct.FindBaseline` / `--baseline-dir` / `--baseline-s3` configured for the tenant), seed the row universe from the baseline Parquet and merge-apply deltas (reuse `internal/reconstruct/fulltable.go` logic rather than duplicating it).
2. Decide `_snapshot` vs `_flashback` semantics: keep `_flashback` as delta-only (current behavior), make `_snapshot` baseline-aware. Document the distinction.
3. Surface a clear signal when no baseline is configured (so a user knows the result is delta-only, not a complete table).
4. Tests: full-table AS OF over a baseline + deltas, including a row present in baseline that was never modified (must appear), and a row deleted after the baseline (must not appear).

## Caveats to carry over from `reconstruct`

- Full-table mode requires a PK; restricted PK types (no FLOAT/BLOB/BINARY/BIT/JSON).
- PK-changing UPDATEs are documented as not handled correctly (`fulltable.go:55-60`).
- Baselines written before `MetaKeyCreateTableSQL` was added need a re-dump.

Related: #187 (offline full-table reconstruction).

nethalo:
closes #355

## Summary
- `_snapshot` now diverges from `_flashback`. With `bintrail shim --baseline-dir`/`--baseline-s3` (the snapshots from `bintrail baseline`), a **single-row** `_snapshot` query seeds the row state from the baseline at-or-before `AS OF` and applies post-snapshot binlog events on top — so a row that existed at `AS OF` but was **never touched** within the retained binlog window now resolves. `_flashback` stays binlog-only, giving the two virtual schemas distinct, observable semantics.
- Pipeline mirrors `bintrail reconstruct`'s single-row path: `FindBaseline → ReadBaselineRow → ApplyAt`. New `internal/shim/snapshot.go`; dispatch split in `HandleQuery` (`_flashback → runPointInTime`, `_snapshot → runSnapshot`).
- Graceful, signalled fallback to the binlog-only path when: no baseline source configured, no baseline found (`ErrNoBaseline`), resolver unavailable, or the PK type isn't safe for the string-param baseline match.
- **PK-type guard** (`baselinePKStringMatchable`) scoped to types whose Parquet form coerces cleanly from a string literal: integer, `YEAR`, `DECIMAL`/`NUMERIC`, and string (`CHAR`/`VARCHAR`/`TEXT`/`ENUM`/`SET`). `DATETIME`/`TIMESTAMP`/`DATE` deliberately fall back (see follow-up #359).
- `shim.Config` gains `BaselineDir`/`BaselineS3`; `cmd/bintrail/shim.go` exposes the flags. Docs + parser comment updated.

## Scope
Single-row only by design (agreed split). **Full-table `_snapshot`** stays binlog-only — tracked as follow-up. Root-cause fix to enable temporal PKs is **#359**.

## Test plan
- [x] Integration (`go test -tags integration ./internal/shim/...`): divergence (never-touched row appears under `_snapshot`, absent under `_flashback`); merge cases (UPDATE wins, DELETE absent, post-baseline INSERT appears); `HandleQuery` dispatch wiring; non-integer supported PK (`DECIMAL`); `DATETIME`-PK fallback; no-baseline fallback.

## Review
Reviewed via `advisor` + a `general-purpose` code-reviewer. Both flagged the datetime-PK silent-miss (root cause: DuckDB reads the Parquet timestamp logical type as `TIMESTAMP WITH TIME ZONE` and casts the string param in the session timezone, missing on non-UTC hosts). Addressed conservatively here by excluding temporal types from the guard; the `SET TimeZone='UTC'` root-cause fix (which also fixes the same latent bug in `bintrail reconstruct --pk`) is tracked in #359.


nethalo:
## Ultrareview: bug_001 fixed ✅

**Empty PK value bypasses pk_values filter** (normal) — confirmed and fixed in `b615538`.

`WHERE name = ''` against a NOT-NULL string PK is a documented-legitimate shape, but `Options.PKValues == ""` makes `buildQuery` (`internal/query/query.go:241`) drop the `pk_values` filter entirely, and the snapshot path set no row bound — so `FetchMerged` returned every event for the table and `ApplyAt` folded an unrelated PK's latest image onto the baseline row (silent wrong answer).

Fix: route the empty value through `PKValuesIn: []string{q.PKValue}`, which emits `pk_values IN (?)` (including the empty-string predicate), while keeping the `pk_hash` fast path for the common non-empty case:

```go
if q.PKValue != "" {
    opts.PKValues = q.PKValue
} else {
    opts.PKValuesIn = []string{q.PKValue}
}
```

Added `TestSnapshotBaseline_EmptyStringPK` as a regression test: `WHERE name = ''` now returns exactly the empty-PK baseline row (asserting on the discriminating `note` column), not the leaked `carol` event. (The empty-string PK value itself may round-trip through Parquet/DuckDB as NULL — a benign representation detail orthogonal to the filter bug.)

PR #364, opened 2026-06-02
fix(stream): scope FK cascade pre-flight away from bintrail-internal schemas

### LINKED ISSUE #347: bintrail stream: FK cascade check scans ALL source schemas, not just --schemas filter — fatal-fails when source has unrelated bintrail-internal tables
## Symptom

`bintrail stream` exits with status 1 (FATAL) on the following pre-flight check:

```
WARN msg="FK cascade constraint found" schema=bintrail_index constraint=fk_access_rules_profile delete_rule=CASCADE update_rule=RESTRICT
{"error":"1 FK cascade constraint(s) found in indexed schemas; reversal SQL from `recover` may not correctly handle cascade side-effects"}
```

systemd then loops the failure (`Scheduled restart job, restart counter is at N`) until it hits `StartLimitBurst` and gives up with `Start request repeated too quickly`.

## Repro on 2026-05-25

A single source MySQL (Percona 5.7.44, demo box `10.42.1.10`) is being indexed by **two** bintrail agents simultaneously — that's a legitimate topology:
- A BYOS-mode agent (legacy, registered earlier).
- A new hosted-mode agent reached via SSH tunnel, just provisioned by a SaaS plan upgrade.

The BYOS agent had earlier created its local index DB **on the source MySQL** (`bintrail_index` schema, with `access_rules` / `profiles` tables and a `fk_access_rules_profile` CASCADE constraint).

The new hosted agent's `bintrail stream` invocation was generated WITHOUT a `--schemas` filter (separate issue on the SaaS side). On startup, the FK pre-flight scanned **every schema on the source MySQL**, found the BYOS agent's `bintrail_index.fk_access_rules_profile`, and fatal-failed — even though the hosted stream was supposed to index only `iotcore` (per the dashboard registration).

Verified by querying `information_schema.REFERENTIAL_CONSTRAINTS` on the source with `CONSTRAINT_SCHEMA='bintrail_index'` — exactly one match, the BYOS agent's FK.

## Root cause

The FK cascade pre-flight in `bintrail stream` reads from `information_schema.REFERENTIAL_CONSTRAINTS` without filtering on `CONSTRAINT_SCHEMA IN (...)` corresponding to the schemas the stream is about to index. So unrelated bintrail-internal tables left by other agents (or, more generally, **any** schema with a cascade FK that the user didn't ask us to index) become a hard blocker.

## Hotfix applied (operator-side workaround)

Edited the systemd unit to add `--schemas=iotcore` and restarted. Stream then reported `Source: no FK cascades ✓` and began streaming normally. Not a fix — just a workaround that depends on the operator knowing to filter schemas.

## Proposed fix

Two parts, both worth doing:

**1. Scope the FK check to the schemas being indexed.**
When `--schemas` is passed, filter the `REFERENTIAL_CONSTRAINTS` query to `CONSTRAINT_SCHEMA IN (...)`. When `--schemas` is omitted, behavior should still scope to schemas the stream will actually consume (filter out `mysql`, `performance_schema`, `information_schema`, `sys`, plus any matching `bintrail_index` / `bt_%` pattern — those are bintrail's own internal schemas and shouldn't be policed by another agent).

**2. Improve the log message clarity.**
Change `schema=bintrail_index` to something like `source_schema=bintrail_index` (or `schema=source:bintrail_index`) so operators can distinguish the source-side schema from the index-side one. Cost me an incorrect hotfix attempt (dropped FK on the LOCAL `bt_<prefix>.access_rules` first) before I realized the check was on the source.

**3. (Optional, lower priority)** Add a `--allow-fk-cascade` override flag so power users can opt into streaming despite cascade risks, with `recover` carrying the matching warning.

## Test plan
- Unit: pre-flight check with `--schemas=iotcore` against a source that has a cascading FK in `bintrail_index` should pass (currently fails).
- E2E: a streaming session against a source containing both `iotcore` (no cascades) and `bintrail_index` (with cascades) starts successfully when `--schemas=iotcore` is provided.

## Impact / severity

**High** — when two bintrail agents share a source MySQL (a legitimate topology supported by the SaaS, e.g. BYOS + cloud SSH-tunnel against the same Percona for forensics demos), the second-installed agent's stream is permanently blocked at startup. Without operator-side debugging this is invisible: the dashboard shows "Stream: Failed (N retries)" with the cryptic error message.

Discovered while working with the SaaS on 2026-05-25 against a dual-indexed Percona for PL2026 demo prep.

## Related (in `nethalo/dbtrail`)
A separate but linked issue exists there: the backend's `stream_start` endpoint doesn't propagate the `schemas` field from `registered_servers` to the agent's `stream_start` payload, so the agent regenerates the systemd unit without `--schemas`. Filing that one separately. Both bugs together produce this incident; fixing either independently would also have prevented it.

nethalo:
closes #347

## Summary
- `bintrail stream`/`index` without `--schemas` scanned **every** non-system schema on the source MySQL for CASCADE foreign keys. When two bintrail agents share one source (a supported BYOS + cloud-SSH-tunnel topology), the second agent's pre-flight tripped over the **first** agent's internal `bintrail_index` / `bt_<prefix>` FK cascade and fatal-failed (status 1 → systemd restart-loop → `StartLimitBurst`) on a schema it was never asked to index.
- Fix: in the unscoped branch, also exclude bintrail's own internal schemas — add `bintrail_index` to the `NOT IN (...)` list and `CONSTRAINT_SCHEMA NOT LIKE 'bt\_%'`. The explicit-`--schemas` branch is unchanged, so an explicitly named schema (even an internal one) is still policed.
- Extracted `buildFKCascadeQuery(schemas)` as a pure function so the query logic is unit-testable without a database.
- Log clarity (issue part 2): `schema=` → `source_schema=` and message → `FK cascade constraint found on source`, so operators don't confuse the source-side schema with the local index schema (which cost the reporter an incorrect hotfix attempt).

Out of scope (per issue): the optional `--allow-fk-cascade` override flag (lower priority; the incident is fully resolved by the above), and the related backend `stream_start` schema-propagation bug tracked separately in `nethalo/dbtrail`.

## Test plan
- [x] New DB-free unit tests on `buildFKCascadeQuery`: unscoped query excludes the 4 system schemas + `bintrail_index` + `bt\_%`; explicit `--schemas` uses a parameterized `IN (?)` list and adds **no** internal exclusions.
- [x] Integration test against real MySQL 8.0 (`-tags=integration`): a CASCADE FK in a `bt_`-prefixed schema is **ignored** when unscoped, but **still caught** when that schema is named explicitly — confirming the `bt\_%` LIKE-escape behaves as intended on the server.

### DIFF
--- a/cmd/bintrail/cmd_integration_test.go
+++ b/cmd/bintrail/cmd_integration_test.go
@@ -277,8 +277,8 @@ func TestValidateNoFKCascades_otherSchemaIgnored(t *testing.T) {
 // bt_<prefix> convention; testutil names every test DB bt_<TestName>). When no
 // --schemas filter is given, FK cascades inside such a schema must be ignored —
 // otherwise a second bintrail agent sharing the source MySQL fatal-fails on the
-// first agent's internal cascade (#347). This also exercises the LEFT(...)='bt_'
-// prefix match against real MySQL.
+// first agent's internal cascade (#347). This also exercises the
+// LEFT(CONSTRAINT_SCHEMA, 3) <> 'bt_' prefix match against real MySQL.
 func TestValidateNoFKCascades_internalSchemaIgnoredWhenUnscoped(t *testing.T) {
 	db, dbName := testutil.CreateTestDB(t)
 
@@ -298,6 +298,29 @@ func TestValidateNoFKCascades_internalSchemaIgnoredWhenUnscoped(t *testing.T) {
 	// But when the operator explicitly names that schema, we still police it.
 	if err := validateNoFKCascades(db, []string{dbName}); err == nil {
 		t.Fatalf("expected error when %q is explicitly targeted despite being bt_-prefixed", dbName)
+	}
+}
+
+// The inverse of the exclusion, and the load-bearing direction: an unscoped
+// scan must still CATCH a real CASCADE FK in an ordinary user schema (no bt_
+// prefix, not an index-DB name). This guards against a regression that
+// broadens the exclusion until the unscoped branch matches nothing. testutil
+// only makes bt_-prefixed DBs, so the non-internal schema is created by hand.
+func TestValidateNoFKCascades_userSchemaCaughtWhenUnscoped(t *testing.T) {
+	db, _ := testutil.CreateTestDB(t)
+
+	const userSchema = "fkcascade_user"
+	testutil.MustExec(t, db, "DROP DATABASE IF EXISTS `"+userSchema+"`")
+	testutil.MustExec(t, db, "CREATE DATABASE `"+userSchema+"`")
+	t.Cleanup(func() { _, _ = db.Exec("DROP DATABASE IF EXISTS `" + userSchema + "`") })
+
+	testutil.MustExec(t, db, "CREATE TABLE `"+userSchema+"`.parents (id INT PRIMARY KEY)")
+	testutil.MustExec(t, db, "CREATE TABLE `"+userSchema+"`.children ("+
+		"id INT PRIMARY KEY, parent_id INT NOT NULL, "+
+		"CONSTRAINT fk_parent FOREIGN KEY (parent_id) REFERENCES `"+userSchema+"`.parents(id) ON DELETE CASCADE)")
+
+	if err := validateNoFKCascades(db, nil); err == nil {
+		t.Fatalf("expected unscoped scan to catch the CASCADE FK in non-internal schema %q, got nil", userSchema)
 	}
 }
 
--- a/cmd/bintrail/cmd_integration_test.go
+++ b/cmd/bintrail/cmd_integration_test.go
@@ -277,8 +277,8 @@ func TestValidateNoFKCascades_otherSchemaIgnored(t *testing.T) {
 // bt_<prefix> convention; testutil names every test DB bt_<TestName>). When no
 // --schemas filter is given, FK cascades inside such a schema must be ignored —
 // otherwise a second bintrail agent sharing the source MySQL fatal-fails on the
-// first agent's internal cascade (#347). This also exercises the `bt\_%` LIKE
-// escape against real MySQL.
+// first agent's internal cascade (#347). This also exercises the LEFT(...)='bt_'
+// prefix match against real MySQL.
 func TestValidateNoFKCascades_internalSchemaIgnoredWhenUnscoped(t *testing.T) {
 	db, dbName := testutil.CreateTestDB(t)
 
--- a/cmd/bintrail/cmd_integration_test.go
+++ b/cmd/bintrail/cmd_integration_test.go
@@ -273,6 +273,34 @@ func TestValidateNoFKCascades_otherSchemaIgnored(t *testing.T) {
 	}
 }
 
+// A bt_-prefixed schema is one of bintrail's own internal schemas (the SaaS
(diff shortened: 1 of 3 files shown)


nethalo:
## Review round (xhigh multi-agent review) — hardening applied in 430b391

The review surfaced one real gap and a few correctness/clarity items, now fixed:

- **Wrong default name (the important one):** the exclusion hardcoded `bintrail_index`, but the codebase's documented default index DB name is `binlog_index` everywhere (config.go, init.go, status.go, agent.go, rotate.go, doctor.go, upload.go). An unscoped pre-flight against a default-named, co-located index DB would still fatal-fail on bintrail's own `access_rules` cascade. Now excludes **both** `binlog_index` and `bintrail_index`.
- **sql_mode-fragile prefix match:** replaced `NOT LIKE 'bt\_%'` with `LEFT(CONSTRAINT_SCHEMA, 3) <> 'bt_'`. A backslash LIKE escape is a trap — `ESCAPE '\'` syntax-errors in the default mode, and the bare pattern under-matches under `NO_BACKSLASH_ESCAPES`. `LEFT()` has nothing to escape and is mode-independent (verified against MySQL 8.0).
- **Misleading error:** "found in indexed schemas" → "found on source" (on the unscoped path nothing was explicitly indexed; this is what `doctor` surfaces).
- Tests assert `binlog_index` exclusion + the escape-free prefix match; integration test still green against real MySQL.

**Known limitation (deliberate, documented in-code):** the name heuristic can't tell bintrail's own `bt_*`/`binlog_index` schema from a user schema that happens to share the name — so such a user schema's real cascades are silently skipped, and a *custom-named* index DB still re-triggers #347. The robust fix (recognise internal schemas by their signature tables) is filed as **#365**.

nethalo:
## Comprehensive review round (code-reviewer + silent-failure-hunter + comment-analyzer + pr-test-analyzer) — applied in a3674c6

- **code-reviewer:** no findings above threshold — confirmed SQL-injection safety (parameterized IN; literal-only unscoped branch), boolean precedence, and that the renamed log key/error string break no existing assertions.
- **silent-failure-hunter (HIGH):** the unscoped scan silently carved bintrail-internal schemas out of a *safety* gate with no disclosure. Unlike the system-schema exclusions, this is a name heuristic that can wrongly skip a real user schema. Added one `slog.Info` on the unscoped path naming the excluded schemas, so a "no FK cascades" pass no longer hides that some schemas weren't scanned.
- **pr-test-analyzer (sev 8):** the load-bearing *catch* direction (unscoped scan must still FAIL on a CASCADE in a non-internal user schema) was only string-asserted. Added `TestValidateNoFKCascades_userSchemaCaughtWhenUnscoped` (real MySQL, hand-created non-`bt_` schema) — guards against a regression that broadens the exclusion until it matches nothing.
- **comment-analyzer (2 factual fixes):** (1) corrected an inaccurate comment — a bare `LIKE 'bt\_%'` does *not* syntax-error in default mode (only an explicit `ESCAPE '\'` would); it under-matches under `NO_BACKSLASH_ESCAPES`. (2) Softened the `binlog_index`/`bintrail_index` distribution claim (both are used widely; the incident used `bintrail_index`) and cited #365 by number. Fixed the integration-test comment operator (`<>`, not `=`).

All unit tests + FK integration tests pass against MySQL 8.0; full `go test ./...` green.

PR #366, opened 2026-06-02
fix(stream): detect bintrail-internal schemas structurally, not by name

### LINKED ISSUE #365: FK cascade pre-flight: detect bintrail-internal schemas structurally, not by name
Follow-up to #347 / PR #364.

## Context

The #347 fix scopes the FK cascade pre-flight (`validateNoFKCascades` / `buildFKCascadeQuery` in `cmd/bintrail/index.go`) so that, when no `--schemas` is given, it skips bintrail's own internal schemas. It does this with a **name heuristic**: exclude `binlog_index`, `bintrail_index`, and any `bt_`-prefixed schema.

That heuristic resolves the reported incident and the common default-named deployments, but it has two inherent holes (both surfaced and verified during PR #364 review):

1. **Over-exclusion (silent safety hole).** A *real user* schema that happens to match these names — e.g. an application schema literally named `bt_orders` or `binlog_index` — has its genuine `ON DELETE/UPDATE CASCADE` FKs silently skipped. The pre-flight prints `Source: no FK cascades ✓` and `recover` may later emit reversal SQL that omits cascade side-effects, with **no warning**. This is a hole punched in a correctness gate.

2. **Under-exclusion (re-triggers #347).** A bintrail index DB with a **custom name** (operator-chosen via `--index-dsn`, not `binlog_index`/`bintrail_index`/`bt_*`) is *not* excluded. If co-located on the source, an unscoped pre-flight finds bintrail's own `access_rules → profiles` CASCADE (`init.go`) and fatal-fails — exactly the #347 failure mode, just for a renamed index DB.

A name can't distinguish "bintrail's own schema" from "a user schema that looks like one", so no name list closes both holes.

## Proposed robust fix

Recognise a bintrail-internal schema by its **signature tables** rather than its name: a schema is bintrail-internal iff it contains the index tables bintrail creates (`binlog_events`, `schema_snapshots`, `stream_state`, `access_rules`, `profiles`, …). Exclude exactly those schemas from the unscoped FK-cascade scan. This kills both the over-exclusion (a user `bt_orders` with no bintrail tables is no longer skipped) and the under-exclusion (a custom-named index DB *is* recognised). Optionally also exclude the agent's own index schema derived from `--index-dsn` DBName as a cheap belt-and-suspenders signal.

## Related cleanups (low priority, can fold in or split)

- **Exclusion-list divergence.** The unscoped FK scan now excludes bintrail-internal schemas, but `checkSchemaVisibility` (`doctor.go`) and `metadata.TakeSnapshot` (`internal/metadata/metadata.go:252`) still exclude only the 4 MySQL system schemas. A schema the FK check skips can still be snapshotted/indexed — so a name-colliding user schema is indexed while its cascades go unchecked. Structural detection makes all three agree if they share the helper.
- **Duplicated system-schema list.** `('mysql','information_schema','performance_schema','sys')` is inlined in ~8 places (`index.go`, `doctor.go`, `agent.go`, `agent_validate.go`, `internal/agent/handler.go`, `internal/metadata/metadata.go`). Consider a single shared constant.

## Acceptance
- Unscoped FK pre-flight skips bintrail's own index schema regardless of its name.
- A user schema named `bt_*` / `binlog_index` that contains a real CASCADE FK and is *not* a bintrail index schema is still flagged.
- Test against real MySQL covers both (the current integration test can't: `testutil` prefixes every test DB `bt_`).

nethalo:
closes #365

## Summary
Follow-up to #347. Replaces the name heuristic in the unscoped FK-cascade pre-flight (`buildFKCascadeQuery`) with **structural detection**: a schema is treated as bintrail-internal only if it contains all of bintrail's signature index tables — `binlog_events`, `schema_snapshots`, `stream_state` — via a `NOT IN` subquery over `information_schema.TABLES` with `HAVING COUNT(DISTINCT TABLE_NAME) = 3`.

This closes both holes the `binlog_index`/`bintrail_index`/`bt_` name heuristic left open:

- **Under-exclusion fixed:** a custom-named index DB (operator-chosen via `--index-dsn`) is now recognised and skipped — it no longer re-triggers the #347 fatal pre-flight on bintrail's own `access_rules→profiles` cascade.
- **Over-exclusion fixed:** a real user schema that merely shares a bintrail-ish name (e.g. `bt_orders`) is no longer silently skipped — without the signature tables it is scanned and its genuine cascades flagged.

The signature tables are created *before* `access_rules` in `bintrail init`, so any schema carrying that cascade necessarily carries the signature too. The explicit-`--schemas` path is unchanged (an explicitly named schema is always policed).

Occam'd scope: one SQL change + comment/disclosure updates + test updates. **Deferred** (separate concern, as the issue allowed splitting): the `doctor`/`metadata.TakeSnapshot` exclusion-list divergence and the duplicated system-schema literal across ~8 files.

## Test plan
- [x] Unit assertions updated to the structural subquery (signature tables + `HAVING = 3`); explicit-`--schemas` path asserts no subquery.
- [x] Integration test against real MySQL 8.0: `audit_index` (signature tables + cascade, **non-bintrail name**) is skipped unscoped (under-exclusion fix); `fkcascade_user` (cascade, **no signature tables**) is still caught (over-exclusion fix); explicitly named schemas still policed.

### DIFF
--- a/cmd/bintrail/cmd_integration_test.go
+++ b/cmd/bintrail/cmd_integration_test.go
@@ -364,6 +364,34 @@ func TestValidateNoFKCascades_userSchemaCaughtWhenUnscoped(t *testing.T) {
 	}
 }
 
+// A schema with only SOME of the signature tables (here 2 of 3) is NOT a
+// bintrail index — its real CASCADE must still be CAUGHT when unscoped. This
+// pins the exactness of HAVING COUNT(DISTINCT TABLE_NAME) = 3: a regression to
+// >= 1 (or a shorter IN-list) would silently skip this real user cascade,
+// reopening the #347-class silent-skip bug. (The existing "caught" test above
+// uses zero signature tables, so it would not detect such a loosening.)
+func TestValidateNoFKCascades_partialSignatureCaughtWhenUnscoped(t *testing.T) {
+	db, _ := testutil.CreateTestDB(t)
+
+	const userSchema = "fkcascade_partial"
+	testutil.MustExec(t, db, "DROP DATABASE IF EXISTS `"+userSchema+"`")
+	testutil.MustExec(t, db, "CREATE DATABASE `"+userSchema+"`")
+	t.Cleanup(func() { _, _ = db.Exec("DROP DATABASE IF EXISTS `" + userSchema + "`") })
+
+	// Two of the three signature names, but not all three.
+	testutil.MustExec(t, db, "CREATE TABLE `"+userSchema+"`.binlog_events (id INT PRIMARY KEY)")
+	testutil.MustExec(t, db, "CREATE TABLE `"+userSchema+"`.stream_state (id INT PRIMARY KEY)")
+	// ...plus a genuine cascade that must be caught.
+	testutil.MustExec(t, db, "CREATE TABLE `"+userSchema+"`.parents (id INT PRIMARY KEY)")
+	testutil.MustExec(t, db, "CREATE TABLE `"+userSchema+"`.children ("+
+		"id INT PRIMARY KEY, parent_id INT NOT NULL, "+
+		"CONSTRAINT fk_parent FOREIGN KEY (parent_id) REFERENCES `"+userSchema+"`.parents(id) ON DELETE CASCADE)")
+
+	if !unscopedFKCascadeSchemas(t, db)[userSchema] {
+		t.Fatalf("expected partial-signature schema %q (2 of 3 signature tables) to be flagged by the unscoped scan, but it was not", userSchema)
+	}
+}
+
 // ─── ensureResolver ──────────────────────────────────────────────────────────────────
 
 func TestEnsureResolver_autoSnapshot(t *testing.T) {
--- a/cmd/bintrail/cmd_integration_test.go
+++ b/cmd/bintrail/cmd_integration_test.go
@@ -273,31 +273,40 @@ func TestValidateNoFKCascades_otherSchemaIgnored(t *testing.T) {
 	}
 }
 
-// A bt_-prefixed schema is one of bintrail's own internal schemas (the SaaS
-// bt_<prefix> convention; testutil names every test DB bt_<TestName>). When no
-// --schemas filter is given, FK cascades inside such a schema must be ignored —
-// otherwise a second bintrail agent sharing the source MySQL fatal-fails on the
-// first agent's internal cascade (#347). This also exercises the
-// LEFT(CONSTRAINT_SCHEMA, 3) <> 'bt_' prefix match against real MySQL.
-func TestValidateNoFKCascades_internalSchemaIgnoredWhenUnscoped(t *testing.T) {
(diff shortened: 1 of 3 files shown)


nethalo:
## Focused review round (4-angle workflow) — applied in e6a70e0

The structural-detection logic reviewed clean — **no confirmed logic bugs**. The NOT-IN-NULL trap was checked and refuted (`TABLE_SCHEMA` is never NULL; an empty subquery excludes nothing, which is correct), and the detection-correctness angle (init order, partitioned table, case-sensitivity) found nothing real. Two PLAUSIBLE, **test-only** findings, both fixed:

- **Flaky nil-assertion (the sharp one):** the old `validateNoFKCascades(db, nil) == nil` check was anti-monotonic, and this PR *widened* its flaky surface — under the old name heuristic a leaked `bt_*`-cascade schema from a crashed run was excluded, but the new structural rule *catches* it (cascade, no signature tables). Replaced with a targeted `unscopedFKCascadeSchemas` helper that asserts `audit_index` specifically is excluded, immune to unrelated/leftover cascades on a shared server.
- **Partitioned-fixture gap:** production `binlog_events` is RANGE-partitioned; the fixture used a plain table. Partitioned it in the fixture — empirically confirms the subquery's `TABLE_TYPE = 'BASE TABLE'` filter matches a partitioned table (audit_index still correctly excluded).

All unit + FK integration tests pass against MySQL 8.0; full `go test ./...` green.

nethalo:
## Comprehensive review round (code-reviewer + silent-failure-hunter + comment-analyzer + pr-test-analyzer) — applied in 83ff07e

- **code-reviewer:** no findings above threshold. Independently verified the load-bearing ordering invariant *across git history* — `stream_state` (f19c02c) predates `profiles`/`access_rules` (14cd6f0) — so an in-place upgrade of a long-lived shared MySQL cannot trip the pre-flight on bintrail's own index DB.
- **comment-analyzer:** all comments accurate; confirmed the "signature created before access_rules" claim against `init.go`. No changes.
- **silent-failure-hunter (HIGH):** the unscoped skip-disclosure log asserted skipped schemas were "bintrail's own" and named nothing — for a runtime-dynamic match, that would falsely reassure an operator if a user schema replicated all three signature table names. **Reworded** to state the rule and name the signature tables (`binlog_events`, `schema_snapshots`, `stream_state`), so a (rare) collision is self-diagnosable rather than misattributed. I chose the honest reword over adding a per-startup probe query to enumerate skipped schemas — disproportionate given the trigger requires a user schema replicating bintrail's three signature names.
- **pr-test-analyzer (sev 7):** added `TestValidateNoFKCascades_partialSignatureCaughtWhenUnscoped` — a schema with only 2 of 3 signature tables + a real cascade must still be **caught**, pinning `HAVING COUNT(DISTINCT TABLE_NAME) = 3` as exact. The existing "caught" test uses zero signature tables, so it wouldn't detect a loosening to `>= 1`.

All unit + FK integration tests pass against MySQL 8.0; full `go test ./...` green.

PR #412, opened 2026-06-06
control plane F4: multi-source hardening — per-source metrics, replica detection, richer states

### LINKED ISSUE #402: control plane F4: multi-source hardening (per-source metrics, replica detection, richer states)
v0.8.1 shipped the control plane (#398/#399/#400): "+ Add server" in `bintrail up --console` runs doctor → provisions a per-source index DB → starts a supervised stream, with boot reconcile, advisory-lock double-daemon protection, and crash-loop backoff. N sources already work; this issue tracks the deliberately-punted hardening (phase 4 of the approved blueprint, `drafts/control-plane-blueprint.md`):

- [ ] **Per-source Prometheus metrics**: `internal/observe` metrics are process-global today; with N streams they aggregate misleadingly. Add a `bintrail_id`/entry label and one metrics endpoint on the supervisor.
- [ ] **Replica-of-monitored detection**: adding a replica of an already-monitored server double-indexes silently in position mode. GTID-mode detection (compare `server_uuid`/executed sets) → amber warn card in the add-server doctor flow (warn-only per the approved decision, never hard-block).
- [ ] **Richer monitor states**: split `stalled` (connected but no events/checkpoint advance for N minutes) and `lost_position` (unfillable gap auto-advance happened — events were lost) out of the generic `failed`/`running`. Both are silent today: a stalled stream shows RUNNING, and a gap auto-advance is only a log line.
- [ ] **Supervisor restart-storm guard**: backoff exists per entry; consider a circuit-breaker state (`failed` permanent after K hours of crash-looping) so a misconfigured source doesn't retry forever.
- [ ] **Empty-schema start UX**: starting monitoring on a schema with no tables fails the initial snapshot ("no columns found") and retry-loops; the doctor could pre-check and the error could say "create at least one table first".

Each item is independently shippable.

nethalo:
closes #402

## Summary

Phase 4 of the approved control-plane blueprint — all five issue items plus the `pending`-state fix from the #407 postmortem comment:

- **Per-source Prometheus metrics** — every `bintrail_stream_*` metric now carries a `source` label (`observe.ForSource`: entry ID when supervised, resolved `bintrail_id` for standalone `stream`). Under `up --console` the **daemon serves one `/metrics` endpoint** for all supervised streams; the main stream's per-process bind is cleared there to avoid a double-bind. Unlabeled gauges previously clobbered each other last-writer-wins with N streams.
- **Replica/duplicate detection (warn-only, amber card)** — the supervisor's doctor reads the candidate's `@@server_uuid` / `@@gtid_executed` and compares against every other monitored entry's recorded identity (`bintrail_servers`) and accumulated executed set (`stream_state.gtid_set`) — no connection to the other sources needed. Catches *replica-of-monitored*, *primary-of-monitored-replica*, and *same-server-added-twice*. Needs `gtid_mode=ON`; skipped otherwise. Never blocks (per the approved decision).
- **Richer monitor states** — `pending` now covers launch→first checkpoint (flip happens via stream liveness hooks: first successful checkpoint save or batch flush); `stalled` (alive but no progress for 5 min — the checkpoint ticker fires even with zero events, so an idle-healthy source never trips it) and `lost_position` (gap auto-advance skipped purged events; sticky until Stop+Start) are **derived at read time** — the stored state machine stays `pending|running|failed|stopped`.
- **Circuit breaker** — continuous crash-looping for 6 h → permanent `failed (gave up …)`, goroutine exits, advisory lock released; Start (or daemon restart) re-arms. Also: the stream seeds its first checkpoint with the resolved start position instead of persisting an empty file/pos 0.
- **Empty-schema UX** — doctor's schema-visibility check now distinguishes *"schema exists but has no tables yet → create at least one table first"* from *"not visible → grants/typo"*; `TakeSnapshot`'s no-columns error gives the same guidance.

UI: new `STALLED` / `LOST POSITION` badges (amber/red), state-aware tooltips, and the Stop/Start verb logic treats the new states as live.

## Test plan
- [x] Unit tests pass (`go test ./... -count=1`, 23 packages)
- [x] Integration suite green against Docker MySQL (`go test -tags integration ./cmd/bintrail/ ./internal/metadata/`), including `TestIntegrationMonitorSupervisor` with the new pending→running semantics
- [x] New unit coverage: derived states + precedence, pending-flip hooks, circuit breaker (gave-up + clean-stop), `gtidSetContainsUUID` / `classifyReplicaOverlap` (incl. lowercase-UUID gotcha), per-source metric isolation
- [x] Console frontend verified with a real headless-Chrome render (state classification, badge CSS, modal flow — screenshot in session)

### DIFF
--- a/cmd/bintrail/doctor.go
+++ b/cmd/bintrail/doctor.go
@@ -523,6 +523,22 @@ func checkSchemaVisibility(ctx context.Context, db *sql.DB, schemas []string) ch
 		if len(schemas) > 0 {
 			filter = strings.Join(schemas, ", ")
 		}
+		// Zero tables has two very different causes: the schema genuinely
+		// has no tables yet (fix: create one) vs. bintrail cannot see it at
+		// all (fix: grants / schema-name typo). Telling the operator to
+		// GRANT when the schema is simply empty sends them down the wrong
+		// path (#402).
+		if n, err := countVisibleSchemas(ctx, db, schemas); err == nil && n > 0 {
+			return checkResult{
+				Name:   "Schema visibility",
+				Status: statusFail,
+				Detail: "schema visible but contains no tables yet: " + filter,
+				Remediation: "Bintrail snapshots table schemas when monitoring starts and cannot monitor\n" +
+					"an empty schema. Create at least one table first:\n\n" +
+					"  CREATE TABLE <schema>.<table> (id INT PRIMARY KEY, ...);\n\n" +
+					"Then start monitoring again.",
+			}
+		}
 		return checkResult{
 			Name:   "Schema visibility",
 			Status: statusFail,
@@ -531,7 +547,8 @@ func checkSchemaVisibility(ctx context.Context, db *sql.DB, schemas []string) ch
 				"Grant minimum read access:\n\n" +
 				"  GRANT SELECT ON *.* TO <bintrail-user>;\n\n" +
 				"Or scope to the schemas you want indexed:\n\n" +
-				"  GRANT SELECT ON <schema>.* TO <bintrail-user>;",
+				"  GRANT SELECT ON <schema>.* TO <bintrail-user>;\n\n" +
+				"Also double-check the schema names for typos.",
 		}
 	}
 	return checkResult{
@@ -539,6 +556,29 @@ func checkSchemaVisibility(ctx context.Context, db *sql.DB, schemas []string) ch
 		Status: statusPass,
 		Detail: fmt.Sprintf("%d tables across %d schemas", tableCount, schemaCount),
 	}
+}
+
+// countVisibleSchemas reports how many of the requested schemas exist and are
+// visible to the doctor's connection (all non-system schemas when the filter
+// is empty). It distinguishes "schema is empty" from "schema is invisible"
+// in checkSchemaVisibility.
+func countVisibleSchemas(ctx context.Context, db *sql.DB, schemas []string) (int, error) {
+	var query string
+	var args []any
+	if len(schemas) > 0 {
+		placeholders := strings.TrimRight(strings.Repeat("?,", len(schemas)), ",")
+		query = fmt.Sprintf(`SELECT COUNT(*) FROM information_schema.SCHEMATA
+			WHERE SCHEMA_NAME IN (%s)`, placeholders)
+		for _, s := range schemas {
+			args = append(args, s)
+		}
+	} else {
+		query = `SELECT COUNT(*) FROM information_schema.SCHEMATA
+			WHERE SCHEMA_NAME NOT IN ('information_schema','performance_schema','mysql','sys')`
+	}
+	var n int
+	err := db.QueryRowContext(ctx, query, args...).Scan(&n)
+	return n, err
 }
 
 func checkIndexConnection(ctx context.Context, dsn, dbName string) checkResult {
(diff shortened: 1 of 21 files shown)


nethalo:
**Adversarial review pass (3 agents) — findings fixed in `2b2d1a7`:**

- **HIGH — `lost_position` evaporated on daemon restart**: the auto-advance persisted its checkpoint but the data-loss signal was in-memory only, so a restart showed a clean RUNNING badge over permanently lost events. Now durable: `stream_state.gap_lost_at/_detail` (DDL + `EnsureSchema` migration), re-hydrated by `Start`, cleared only by an explicit `Stop` (operator acknowledgment; daemon shutdown never calls Stop). Integration test covers record → restart → `lost_position` → ack → clean running.
- **Warn cards were computed but never shown when `started=true`** (both the save flow and the row Start button rendered the doctor report only on failure) — a replica-of-monitored warn started the stream and silently discarded the card. Started-with-warnings now keeps the form open with the amber cards. Note: this surfaces *all* warnings (retention, FK cascades too), matching what console.md already claimed. The replica remediation no longer promises a "press Start again" step that never existed.
- **`/metrics` bind is now synchronous + fail-fast** (was async log-only — an operator passing `--metrics-addr` could get a silently absent endpoint), symmetric with the console bind.
- **Replica detection is honest about unparseable GTID sets**: malformed candidate set → visible skip card; malformed peer set → counts toward "(N could not be verified)".
- Comment/docs accuracy: `Since` semantics for derived states, `errors_total{source,type}`, lost-position cause wording.

Follow-up commit also adds `TestIntegrationReplicaOverlapSQL` — the detection SQL (registry → source gtid_mode gate → `loadPeerIdentity` against a seeded per-source DB) now runs against real MySQL; the container is `gtid_mode=OFF` so the full warn path stays covered by the pure-function tests.

nethalo:
**Second review round (5 agents: code, silent-failures, comments, tests, type-design) — fixed in `0065362`:**

- **Data race on `monitorJob.lockDB`** (introduced by the lost_position fix in `2b2d1a7`): `Stop()` read `lockDB` unlocked against `Start()`'s provisioning-window write; a Stop on an exited job also Exec'd on a closed pool, silently failing the gap-record clear. `Stop` now uses its own short-lived connection via an immutable `job.indexDSN`; `lockDB` is back to Start→run only (happens-before safe). `-race` clean.
- **`Start` idempotency** now gates on the stored state machine, not a hand-enumerated list of derived presentation strings — adding a derived state can no longer cancel a live stream by omission.
- **Hydration breadcrumb**: real DB errors during gap-record re-hydration are logged (ErrNoRows stays quiet); previously a transient error silently un-surfaced a recorded data loss.
- **Breaker testability**: backoff/healthy-reset lifted to vars; new test pins that a flapping-but-healthy stream never trips the breaker (false trip = monitoring silently stops — that direction was untested).
- **Replica-check seam**: pure `evaluateReplicaOverlap` (warn-card assembly + unverified accounting) now unit-tested — the GTID-ON warn path was untestable on the gtid_mode=OFF container; `loadCandidateIdentity` extracted so the integration test exercises production queries instead of an inline copy.
- **`checkSchemaVisibility`** empty-vs-invisible branch (the #402 UX fix) gets sqlmock coverage.
- **app.js**: `editServer` reports whether the form opened; started-with-warnings degrades to a toast instead of rendering the replica warning into a hidden form on concurrent delete. Re-verified in headless Chrome.

Remaining non-blocking notes (accepted): breaker re-arms on daemon restart (documented, arguably desired), stale per-source metric series after Stop, `Since` reflects the stored-state transition for derived states (documented).

Issue #212, opened 2026-04-11
Full-table reconstruct: robust PK encoding for non-integer and mutated primary keys


nethalo:
## Summary

Follow-up to #187. The v1 full-table reconstruct (PR #211) has two documented limitations in the PK-handling path that are correct for the common case (integer and string PKs, stable) but silently produce wrong output for the edge cases.

Both limitations stem from the same design choice: the in-memory change map in `internal/reconstruct/fulltable.go::mergeBaselineIntoWriter` is keyed by the `pk_values` string that was pre-computed at indexing time by `parser.BuildPKValues(pkCols, row)`, and the baseline side rebuilds the same key from DuckDB-scanned row values using the same function. Byte-identical keys are needed for the match to work.

## Limitation 1: Non-integer PK types may not round-trip

`parser.BuildPKValues` uses `fmt.Sprintf("%v", row[col.Name])` to format each PK value. For INT/BIGINT PKs this is stable: go-mysql returns `int32`/`int64`, DuckDB returns `int32`/`int64`, and `%v` produces the same decimal string on both sides. For VARCHAR PKs the same holds (both return `string`).

For **DATETIME/TIMESTAMP** PKs go-mysql may return a formatted string (`"2026-04-11 14:30:45"`) while DuckDB `parquet_scan` returns `time.Time` — `%v` on a `time.Time` produces `"2026-04-11 14:30:45 +0000 UTC"`. The keys will not match and every event will miss the baseline. For **DECIMAL** PKs the situation is worse because go-mysql may use a third-party decimal type while DuckDB uses `string` for decimal-backed-as-string or `float64`. **ENUM/SET** PKs and **BINARY/VARBINARY** PKs are also at risk.

Effect: events for rows in tables with non-integer PKs silently become "new inserts" in the final dump, duplicating rows (if the baseline also has them) or silently missing deletions.

## Limitation 2: UPDATE events that mutate the primary key

The change map is keyed by `ev.PKValues`, which for UPDATE events is built from the **before** image (see `internal/parser/parser.go:359`: `PKValues: BuildPKValues(pkCols, before)`). When an UPDATE changes the PK itself (e.g. `id` mutates from `P1` → `P2`), the event is stored with key `P1`. The baseline row `P1` is located and its `row_after` (carrying `P2`) is written. So far so good.

[...]

PR #474, opened 2026-06-11
fix(shim): map ENUM/SET ordinals back to labels in time-travel results

### LINKED ISSUE #472: shim: map ENUM ordinals back to labels in time-travel results
## Problem

Live rows and time-travel rows render ENUM columns differently through the same ProxySQL connection:

```sql
SELECT status FROM orders WHERE id = 1;
--> 'pending'

SELECT status FROM orders WHERE id = 1 AS OF '1 minute ago';
--> '3'    -- the ordinal for 'shipped'
```

Binlog ROW images store ENUMs as their 1-based ordinal; the shim materializes historical rows from the stored `row_before`/`row_after` images verbatim, never mapping the ordinal back to the label. Every time-travel path is affected (`_flashback`, `_snapshot`, `_diff`, bare `AS OF`, `DBTRAIL_AT` hint). Verified live on the released `bintrail-demo:0.13.2`; `docs/demo.md` currently documents it as a troubleshooting row.

## Fix

The schema snapshot already has what's needed: `schema_snapshots.column_type` stores the full `information_schema.COLUMNS.COLUMN_TYPE` (e.g. `enum('pending','processing','shipped','delivered','cancelled')`) since #212, loaded into `metadata.Column.ColumnType`. Map ordinal → label at the shim's row-materialization point.

Edge cases to cover:

- ordinal `0` = MySQL's invalid-value empty string `''`
- `NULL` stays `NULL`
- out-of-range ordinal (enum shrunk by a later ALTER) → leave the number as-is, don't error
- pre-#212 snapshots with empty `column_type` → leave as-is (current behavior)
- JSON round-trip means the ordinal arrives as `float64`
- `SET` columns have the sibling problem (bitmask instead of labels) — include if it falls out naturally from the same mapping point, otherwise leave for a follow-up

Once fixed, drop the ENUM troubleshooting row from `docs/demo.md`.

## Acceptance

A time-travel query over an ENUM column returns the same label representation as the live row; shim tests cover the mapping + the edge cases above.

nethalo:
closes #472

## Summary

- Binlog ROW images store ENUMs as 1-based ordinals and SETs as bitmasks, so every time-travel path (`_flashback`, bare `AS OF`, `DBTRAIL_AT` hint, `_snapshot`, `_diff`) returned `'3'` where a live `SELECT` returns `'shipped'` — verified live on the released demo image during the v0.13.2 verification pass.
- New `internal/shim/enumlabel.go`: parses `enum(...)`/`set(...)` member labels from the schema snapshot's `column_type` (captured since #212) and maps ordinals→labels in row images. Nil-safe no-op mapper when the snapshot is unavailable or the table has no ENUM/SET columns.
- Applied at all five materialization points: `runPointInTime`, `runFullTable`, `runDiff` (both image JSONs), `runSnapshotPointInTime` (post-`ApplyAt`), `runSnapshotFullTable` (before `fullTableTextCell`, which would otherwise hide the ordinal as a text cell).
- Conservative by design: out-of-range ordinals (enum shrank post-event), non-integral values, SET masks with unknown bits, and values already strings (the `_snapshot` baseline carries labels from mydumper) pass through unchanged. Ordinal `0` → MySQL's empty-string invalid-entry sentinel. This also makes baseline+delta `_snapshot` rows internally consistent (labels vs ordinals in one resultset before this fix).
- `columnOrderFor`'s resolver lookup extracted into `tableMetaFor` so the mapper shares the same graceful-degradation contract (broken/absent snapshot never turns a working query into an error).
- Drops the now-obsolete "Historical ENUM shows a number" troubleshooting row from docs/demo.md.

## Test plan

- [x] New unit tests: `parseEnumSetLabels` (escaped quotes, commas in members, empty members, malformed input, pre-#212 empty column_type) and `enumLabelMapper` (ordinal/bitmask mapping, 0-sentinels, out-of-range/non-integral/string/NULL pass-through, nil safety)
- [x] New integration test `TestRunPointInTime_EnumSetOrdinalsMapToLabels`: real index DB, snapshot with `column_type`, UPDATE event with ordinal images → labels on the wire for both `_flashback` and `_diff`
- [x] Full unit suite green (`go test ./... -count=1`)
- [x] Full shim integration suite green (`go test -tags integration ./internal/shim/ -count=1`)


nethalo:
**Second review wave applied** (comprehensive multi-agent: test-coverage + type-design on top of the earlier code/silent-failure/comments pass):

- **Critical catch (test-coverage agent, reproduced live):** `schema_snapshots.column_type` was `VARCHAR(128)` since #212 — a mundane 10-member ENUM renders a 136+-char `COLUMN_TYPE`, and under strict mode the 1406 **aborts the entire snapshot transaction**, so `bintrail snapshot` failed outright for schemas with realistic enums. This PR now widens it to `TEXT` (DDL + idempotent `ensureColumnWidened` MODIFY migration in `EnsureSchema`; values preserved; resolver already COALESCEs). Commit 5162f05.
- New behavioral coverage for every render path that had none: both `_snapshot` paths (baseline-label pass-through, post-`ApplyAt` delta mapping, and the map-BEFORE-`fullTableTextCell` ordering constraint), `runFullTable`'s hoisted loop, a full-chain test (real `CREATE TABLE` → `TakeSnapshot` with a 189-char enum + backslash member → labels on the wire — the test that would have caught the 1406), the `EnsureSchema` widening migration, and the `enumMapperFor` composition.
- Type-design agent: no changes warranted (encapsulation 8/10, invariants 7-9/10); follow-up note recorded on #476 — when the mapper is reused, move the file to `internal/metadata` rather than duplicating the parser.

PR #482, opened 2026-06-13
fix(console): four monitor/index UX bugs that made the UI look broken

nethalo:
Surfaced while debugging a live `+Add server` monitored source that wouldn't start. The infra was fine — these console bugs **compounded** into a UI that looked completely broken (white buttons, vanishing Start, a wall of `Unknown database` errors, the whole monitor surface disappearing).

## Bugs fixed

**1. Primary button invisible on hover** (`style.css`)
`.btn:hover` sets `background: var(--surface-2)`. `.btn-primary:hover` (equal specificity, later in source) only set `filter`/`box-shadow`, so the gray background won → **white text on light gray**. Re-assert the gradient in `.btn-primary:hover`.

**2. A broken selected server hid the WHOLE control plane** (`reconstruct.go`)
`handleCapabilities` resolved the selected bundle first and **502'd** when its index was unreachable (a monitored source whose per-source index isn't provisioned yet). The frontend's `gateCapabilities` then degraded to `{}`, hiding the **Start button** and the **"+Add server" monitor copy**. `Monitor`/`Auth` are **process-level** — report them unconditionally; `Reconstruct` (per-server) stays gated on a resolvable bundle.

**3. Raw `Unknown database 'bintrail_idx_…'` error wall** (`app.js`)
`renderError` now detects MySQL 1049 and shows an **actionable empty state**: the index DB is created when monitoring starts and **never lives on the source** (the exact source-vs-index confusion users hit), with a *Manage servers* action — instead of a red error wall.

**4. Optional "bring your own index" section auto-expanded** (`app.js`)
Editing a monitored source round-trips its derived index DSN as host/dbname, so a failed-preflight error card opened a per-source index the operator **never typed**. Keep Advanced collapsed for source entries.

## Tests
- `TestHandleCapabilitiesMonitorSurvivesUnresolvableServer` — caps degrade gracefully (200 + Monitor) for an unresolvable selection.
- `TestResolveHeader` — moved its dead-server **502** assertion to `/api/status`; the data endpoints carry that signal now, not `/api/capabilities`.

## Verification
Headless Chrome (playwright-core, `channel:'chrome'`) against a source-less `watch` daemon with a monitored entry whose index is absent: Start button present + monitor copy shown, `+Add server` keeps its gradient on hover, Advanced collapsed for the source entry, and `renderError` renders the friendly empty state naming the index DB and clarifying it never lives on the source. Zero JS errors.

> Related to the same investigation: #480 (Test probe "not provisioned yet") and #481 (monitor-start preflight no longer a silent failure).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,3 +93,59 @@ jobs:
       # the sole writer during its run.
       - name: Integration tests (-tags=integration)
         run: go test -tags=integration -race -p 1 -count=1 ./...
+
+  console-e2e:
+    # Headless-Chrome regression guard for the console SPA: go test never
+    # renders assets/*, so CSS/DOM/presentation bugs (invisible buttons, a
+    # form that auto-expands, a raw error wall, the control plane vanishing on
+    # a broken server selection) are invisible to the Go suite. Each scenario
+    # pins a bug that shipped in 0.13.3.
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      # Same manually-named container as the integration job (8.4 = the bundled
+      # index engine, caching_sha2_password default).
+      - name: Start MySQL (bintrail-test-mysql, 8.4)
+        run: |
+          docker run -d --name bintrail-test-mysql \
+            -e MYSQL_ROOT_PASSWORD=testroot \
+            -p 13306:3306 \
+            mysql:8.4 \
+            --binlog-format=ROW --binlog-row-image=FULL --log-bin=binlog --server-id=1
+          for i in $(seq 1 60); do
+            if docker exec bintrail-test-mysql mysqladmin ping -u root -ptestroot --silent 2>/dev/null; then
+              echo "MySQL is ready."; exit 0
+            fi
+            sleep 2
+          done
+          echo "MySQL did not become ready in time" >&2
+          docker logs bintrail-test-mysql >&2 || true
+          exit 1
+
+      - name: Install playwright chromium (+ system deps)
+        run: |
+          npm --prefix test/console-e2e install --no-audit --no-fund
+          npx --prefix test/console-e2e playwright install --with-deps chromium
+
+      # PW_CHANNEL unset → use the chromium installed above (not system Chrome).
+      - name: Console E2E (headless Chrome)
+        run: make console-e2e
+
+      - name: Upload failure screenshot
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: console-e2e-failure
+          path: |
+            ${{ runner.temp }}/console-e2e-failure.png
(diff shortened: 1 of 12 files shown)


nethalo:
## Added: a headless-Chrome E2E regression guard (so this never recurs silently)

These four bugs were invisible to `go test` — it never renders the embedded SPA. To stop that class from shipping again, this PR now also adds `test/console-e2e/`: a real headless-Chrome harness that drives the console through the exact lifecycle state that triggered the bugs (a source-less `watch` daemon with a monitored source whose per-source index is **unprovisioned**), pinning each fix as a scenario:

- boot: monitor capability reported (caps cascade)
- control plane: Start button present + monitor copy visible (caps cascade)
- button: primary keeps its gradient on hover (the white-on-light bug)
- form: advanced section collapsed for a source entry (auto-expand bug)
- error: a 1049 renders the friendly empty state, not a raw wall (×3 assertions)
- no uncaught JS errors

`make console-e2e` runs it locally (Docker + Node; `PW_CHANNEL=chrome` to use system Chrome). A new `console-e2e` job in `ci.yml` runs it on every PR with the playwright-managed chromium and uploads a screenshot on failure.

**Verified: 10/10 pass on this branch; the same guard catches 7/8 on the pre-fix tree** (i.e. it would have caught these bugs). This makes the PR self-contained — fixes + the test that proves them.

nethalo:
## Multi-agent review (code / silent-failure / comments / tests) — applied

Ran 4 specialized reviewers over `main...HEAD`. Verdict: solid, no blockers. Findings addressed in `7b2afa5`:

**Applied**
- **Silent gap in `handleCapabilities`** (silent-failure-hunter, MEDIUM): the `resolve` error was dropped with zero trace, so an *unexpected* failure (a genuine connection error, not just a bad header) would silently strip Time-travel from the UI. Now logs `slog.Warn` for anything that isn't `ErrUnknownServer`/`errNoServers`, mirroring `connManager.Resolve`'s Debug/Warn split.
- **E2E 1049 scenario sidestepped the integration path** (pr-test-analyzer, HIGH): it fabricated the error string. Now it queries `/api/status` against the real unprovisioned default server and feeds the **actual backend error** to `renderError` — proving the 1049 reaches the frontend with the index db name preserved through `scrubDSNError`. A backend format change or over-aggressive scrub now fails the test.
- **Uncovered `byoIndex` arm** (pr-test-analyzer, MEDIUM): added a scenario for a no-source BYO-index entry (must auto-**expand** advanced) — the complement of the collapse-for-source scenario.
- **Hover check hardened**: also asserts the text stays white (visibility is contrast, not just background).
- **Broken doc sentence** (comment-analyzer): `reconstruct.go` "which resolveOr properly" → "which resolveOr 502s properly".

**Noted, not changed (LOW / by design)**
- `renderError`'s 1049 copy ("isn't indexing yet") is slightly off for the rare *index-existed-then-dropped* case; left as-is (still points the user the right way; changing it churns the E2E assertions).
- `TestHandleCapabilitiesMonitorSurvivesUnresolvableServer`'s `Reconstruct==false` sub-assertion is over-determined (no baseline anyway); the load-bearing `Monitor==true` assertion is what discriminates, and it's confirmed to fail on the pre-fix tree.

Harness now **14/14** green against the fixed build.

PR #494, opened 2026-06-14
fix: index UNSIGNED integers with the high bit set as their correct value (#490)

### LINKED ISSUE #490: Data corruption: UNSIGNED integers with high bit set are indexed as negative values
## Summary

UNSIGNED integer column values with the high bit set are indexed as **negative** numbers. `go-mysql` decodes integer columns as signed `int64`, and bintrail never applies the `TABLE_MAP` `SIGNEDNESS` bitmap (nor the column's `unsigned` type from the snapshot) to reinterpret them. The stored value is wrong.

## Evidence

- **go-mysql v1.13.0** `decodeValue` (`replication/row_event.go`) decodes `MYSQL_TYPE_TINY/SHORT/INT24/LONG/LONGLONG` as signed (`ParseBinaryInt*`). `RowsEvent.SignednessBitmap` / `UnsignedMap()` are parsed but used **only** in the `Dump`/pretty-printer, never in `decodeImage`/`decodeValue`.
- **bintrail** reads the decoded value verbatim: `internal/parser/parser.go` (`handleRows`, `emitInserts/emitUpdates/emitDeletes`) → `internal/indexer/indexer.go` (`marshalRow`). No unsigned handling anywhere in the parse/index path.
- **Reproduced** against MySQL 8.0.46: `BIGINT UNSIGNED = 18446744073709551615` is decoded and stored as `int64(-1)`. (Repro lives in an internal audit test, `TestAudit_UnsignedBigint`.)

## Impact

Any unsigned value `>= 2^(N-1)` is stored incorrectly: `BIGINT UNSIGNED > 2^63`, `INT UNSIGNED > 2^31`, etc. (`BIT(64)` with the top bit set is the same problem). Below that threshold the bit pattern is identical, so there is no loss.

The sharp edge is **unsigned primary keys**: an unsigned PK value with the high bit set produces a negative `pk_values` → wrong `pk_hash` → recovery `WHERE` clauses target the wrong rows (or none), and PK lookups fail.

## Suggested fix

Apply `rowsEv.Table.UnsignedMap()` in `handleRows` to reinterpret signed → unsigned for flagged columns. The signedness is also derivable from `metadata.ColumnMeta.ColumnType` (`"... unsigned"`) captured in the snapshot — and that is the more robust source, because the `TABLE_MAP` `SIGNEDNESS` bitmap is only present when `binlog_row_metadata=FULL`.

## References

- Found in the 2026-06-14 binlog-fidelity audit (3-layer: article → go-mysql → bintrail).
- Readyset "Replication Internals: Decoding the MySQL Binary Log — Part 7: TABLE_MAP_EVENT" (the `SIGNEDNESS` optional metadata; type codes do not encode signedness).

nethalo:
closes #490

## Summary
- go-mysql decodes every INT-family column as **signed** and never applies the TABLE_MAP `SIGNEDNESS` bitmap, so an `UNSIGNED` value with the high bit set was indexed as a negative number (e.g. `BIGINT UNSIGNED 18446744073709551615` → `int64(-1)`). For unsigned PKs this also corrupted `pk_values`/`pk_hash`.
- Fix reinterprets the value at the single choke point `metadata.MapRow`, via a new `coerceUnsigned(v, col)`:
  - **Signedness** from the snapshot's `ColumnType` (`"... unsigned"`) — not the TABLE_MAP bitmap, which is only present under `binlog_row_metadata=FULL` (not the default); `ColumnType` is always loaded by the resolver.
  - **Width** from `DataType`: `tinyint→uint8`, `smallint→uint16`, `mediumint→uint32 & 0xFFFFFF`, `int→uint32`, `bigint→uint64`. MEDIUMINT is masked to 24 bits because go-mysql returns it as `int32`.
- No-op when the column isn't unsigned, when `ColumnType` is empty (pre-#212 snapshots), or when the value isn't a signed integer (NULL/string/decimal/…).

## Scope / notes
- Fixing `MapRow` also fixes **unsigned primary keys**, since `pk_values` is built from its output.
- **Out of scope:** `BIT(64)` top-bit (distinct type, not declared `unsigned`); read-side `float64` precision for values >2^53 (separate finding). This PR makes the *stored* value correct — fully correct end-to-end for the common range and strictly better for the huge range.

## Test plan
- [x] New unit tests: `TestCoerceUnsigned` (per-width boundaries incl. the MEDIUMINT 24-bit case, signed unaffected, empty-ColumnType no-op, NULL no-op) and `TestMapRow_unsignedCoercion`
- [x] New integration test `TestParseFile_unsignedIntegers` (real binlog): `BIGINT/INT/MEDIUMINT UNSIGNED` and an unsigned PK index to the correct value
- [x] `internal/parser` + `internal/metadata` integration tests pass (no regression)

### DIFF
--- a/internal/metadata/metadata.go
+++ b/internal/metadata/metadata.go
@@ -199,9 +199,63 @@ func (r *Resolver) MapRow(schema, table string, row []any) (map[string]any, erro
 	}
 	named := make(map[string]any, len(row))
 	for i, col := range tm.Columns {
-		named[col.Name] = row[i]
+		named[col.Name] = coerceUnsigned(row[i], col)
 	}
 	return named, nil
+}
+
+// coerceUnsigned reinterprets an integer value decoded by go-mysql into the
+// correct unsigned value. go-mysql always decodes INT-family columns as SIGNED
+// (int8/int16/int32/int64) — it parses the TABLE_MAP SIGNEDNESS bitmap but never
+// applies it (UnsignedMap is used only in its Dump output). So a column declared
+// UNSIGNED whose value has the high bit set comes back negative (e.g.
+// BIGINT UNSIGNED 18446744073709551615 → int64(-1)). Left uncorrected, the wrong
+// value lands in the index and, for unsigned PKs, corrupts pk_values/pk_hash.
+//
+// Signedness is taken from the snapshot's ColumnType ("... unsigned"), NOT the
+// TABLE_MAP bitmap, because the bitmap is only present under
+// binlog_row_metadata=FULL (not the default), whereas ColumnType is always
+// loaded by the resolver. Width is taken from DataType so the reinterpretation
+// masks to the column's real size — MEDIUMINT is 3 bytes but go-mysql returns it
+// as int32, so it must be masked to 24 bits (else MEDIUMINT UNSIGNED -1 would
+// become 2^32-1 instead of 2^24-1).
+//
+// No-op when the column is not unsigned, when ColumnType is empty (pre-#212
+// snapshots can't express signedness), or when the value is not a signed integer
+// (NULL, string, decimal, time, etc. are returned unchanged). BIT columns are
+// intentionally not handled here — BIT is a distinct type, not declared
+// "unsigned", and has its own representation.
+func coerceUnsigned(v any, col ColumnMeta) any {
+	if !strings.Contains(strings.ToLower(col.ColumnType), "unsigned") {
+		return v
+	}
+	var signed int64
+	switch x := v.(type) {
+	case int8:
+		signed = int64(x)
+	case int16:
+		signed = int64(x)
+	case int32:
+		signed = int64(x)
+	case int64:
+		signed = x
+	default:
+		return v // not a signed integer (NULL/string/decimal/...) — leave as-is
+	}
+	switch strings.ToLower(col.DataType) {
+	case "tinyint":
+		return uint8(signed)
+	case "smallint":
+		return uint16(signed)
+	case "mediumint":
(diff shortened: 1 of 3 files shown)


nethalo:
### Review applied (multi-agent: code / tests / silent-failure / comments)

Follow-up commit `4494896`:
- **Comment fix (factual):** the original doc comment claimed the TABLE_MAP `SIGNEDNESS` bitmap is only present under `binlog_row_metadata=FULL` — that's wrong (it's logged under the default `MINIMAL` too). Corrected: the real reason to use the snapshot's `ColumnType` is that `coerceUnsigned` runs in `MapRow`, which never sees the live `TableMapEvent`.
- **Fail-loud for stale snapshots:** `NewResolver` now emits one `slog.Warn` when a loaded snapshot predates `column_type` (#212), so the fix can't silently no-op an entire pre-#212 snapshot.
- **Tests:** locked the load-bearing DataType-default no-op for `ENUM('signed','unsigned')` (substring false-match, neutralized by the width switch); added an UPDATE before-image assertion (the unsigned PK `recover` keys on).

### Known caveat (out of scope for this PR)
`recover`/reconstruct read `row_before`/`row_after` JSON and unmarshal numbers to `float64`, so a `BIGINT UNSIGNED` value > 2^53 is rounded in the regenerated SQL. This PR makes the **stored** value correct and is fully correct end-to-end for `TINYINT/SMALLINT/MEDIUMINT/INT UNSIGNED` (all < 2^53). Full end-to-end correctness for `BIGINT UNSIGNED` also needs the read-path fix (`json.Decoder` + `UseNumber()`), tracked separately. `BIT(64)` top-bit corruption is likewise separate.

PR #498, opened 2026-06-14
fix: advance GTID checkpoint only at transaction commit boundaries (#491)

### LINKED ISSUE #491: Data loss: GTID-mode mid-transaction checkpoint skips the transaction on restart
## Summary

In **GTID checkpoint mode**, `bintrail stream` can persist a transaction's GTID into the durable `gtid_set` **before that transaction's row events have been fully indexed**. If a checkpoint fires mid-transaction (ticker or shutdown), the saved set already claims the transaction. On restart, GTID auto-positioning makes MySQL **skip** that transaction → its un-indexed remainder is **silently lost**. Root cause: the GTID is accumulated at the `GTID_LOG_EVENT` (the *first* event of the transaction), and no `XID_EVENT` (commit) boundary is honored.

## Evidence

- `internal/streamrun/streamrun.go:672-678`: `state.accGTID.Update(ev.GTID)` runs for every event carrying a GTID — including the leading `EventGTID`, which the parser emits at the start of the transaction, before any row events.
- `checkpoint()` fires on the ticker (`:659`, default `--checkpoint 10s`), on `ctx.Done()` (`:656`), and on channel close (`:664`) — none gated on a transaction boundary.
- No `XID_EVENT` handling exists (the symbol appears only in comments).
- Resume reads and trusts the saved set: `resolveStartPosition` (`streamrun.go:267-273`) parses the persisted `gtid_set` and resumes via GTID auto-position, so MySQL omits any GTID already in the set.

## Impact

Deterministic (not just a race) for any DML transaction whose row events span more than the checkpoint interval — common for bulk operations on RDS/managed MySQL, which is exactly where GTID mode (`--start-gtid`) is the norm. The worst case: the ticker fires right after the `GTID_LOG_EVENT` but before any rows → the **entire** transaction is claimed and then skipped on restart.

## Scope

**GTID mode only.** Position mode is safe: it persists `binlogPos = ev.EndPos` of the last *flushed* event, which is exact — no loss, no duplication.

## Suggested fix

Advance the durable GTID set only at **transaction boundaries** — `XID_EVENT`, or the `QUERY_EVENT` carrying `COMMIT` / `Q_DDL_LOGGED_WITH_XID` — and make flush + checkpoint atomic at those boundaries. This requires handling `XID_EVENT` in the stream dispatch.

## References

- Found in the 2026-06-14 binlog-fidelity audit; accumulation timing and resume path both verified by reading the code.
- Readyset "Replication Internals" Part 5 (GTID_LOG_EVENT — first event of the transaction) and Part 9 (XID_EVENT — never checkpoint mid-transaction).

nethalo:
closes #491

## Summary
- **Bug:** in GTID mode, `accGTID` was updated on every event including the leading `EventGTID` (transaction **start**). A checkpoint firing mid-transaction persisted a GTID for a not-yet-fully-indexed transaction → on restart, GTID auto-position made MySQL **skip** it → silent data loss. No `XID_EVENT` (commit) boundary was processed.
- **Fix:** the `StreamParser` now emits a new `EventCommit` at each `XID_EVENT` (and at a `COMMIT` `QUERY_EVENT`), and `streamLoop` advances the durable GTID set **only** at commit boundaries (`EventCommit`, and `EventDDL` which auto-commits). The leading `EventGTID` is attribution-only.
- **Semantics:** converts the stream from *silently losing data* to **at-least-once** — no loss; an ungraceful crash may re-stream a not-yet-checkpointed transaction (duplicate rows), strictly better than loss. Exactly-once (atomic flush+checkpoint) is tracked separately (follow-up).
- Position mode is unchanged (`accGTID` nil; `EventCommit` is a no-op there and is only emitted for GTID-bearing sources). The agent BYOS loop skips `EventCommit` like other non-row events.

## Files
- `internal/parser/parser.go` — new `EventCommit` event type.
- `internal/parser/stream.go` — `emitCommit` on `XIDEvent` / `COMMIT` query.
- `internal/streamrun/streamrun.go` — `advanceGTID` closure; accumulate at commit boundaries only.
- `cmd/bintrail/agent.go` — skip `EventCommit`.

## Test plan
- [x] Unit tests pass (`go test ./... -count=1`, 28 pkgs)
- [x] `TestStreamLoop_gtidCheckpointAtCommit` (deterministic): a mid-transaction checkpoint does **not** persist the GTID and the row **is** indexed (no loss); a committed transaction **does** advance it.
- [x] `TestStreamLoop_gtidAdvancesOnCommit_live` (end-to-end): with `gtid_mode` enabled (toggled + restored), the parser emits `EventCommit` and `streamLoop` advances `gtid_set` to contain the source UUID.
- [x] Updated `TestStreamLoop_gtidAccumulation` to the commit-boundary semantics.
- [x] `internal/streamrun` + `internal/parser` + `cmd/bintrail` integration + E2E pass; `go vet -tags integration ./...` clean.

### DIFF
--- a/cmd/bintrail/agent.go
+++ b/cmd/bintrail/agent.go
@@ -739,8 +739,8 @@ func byosStreamLoop(ctx context.Context, events <-chan parser.Event, buf *buffer
 				return nil
 			}
 
-			// Skip non-row events (GTID tracking, DDL).
-			if ev.EventType == parser.EventGTID || ev.EventType == parser.EventDDL {
+			// Skip non-row events (GTID tracking, DDL, commit boundaries).
+			if ev.EventType == parser.EventGTID || ev.EventType == parser.EventDDL || ev.EventType == parser.EventCommit {
 				continue
 			}
 
--- a/internal/byos/split.go
+++ b/internal/byos/split.go
@@ -94,8 +94,8 @@ func PKHash(pkValues string) string {
 // permitted (older callers / tests) and produces a record without source
 // identity fields.
 //
-// Returns an error for unsupported event types (DDL, GTID). Callers should
-// filter these before calling SplitEvent.
+// Returns an error for unsupported event types (DDL, GTID, commit boundaries).
+// Callers should filter these before calling SplitEvent.
 func SplitEvent(ev parser.Event, serverID string, ident SourceIdentity) (MetadataRecord, PayloadRecord, error) {
 	typeName, err := eventTypeName(ev.EventType)
 	if err != nil {
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -36,12 +36,14 @@ const (
 	// produces this type — it exists so baseline rows can flow through the
 	// same ResultRow pipeline as real binlog events.
 	EventSnapshot EventType = 6
-	// EventCommit marks a transaction commit boundary (XID_EVENT, or a COMMIT
-	// QUERY_EVENT). It carries no row data — only the GTID of the transaction
-	// that just committed. The stream consumer uses it to advance the durable
-	// GTID checkpoint ONLY after a transaction's rows are fully received, so a
-	// checkpoint can never claim a half-streamed transaction (#491). Emitted only
-	// by the StreamParser; the file parser does not produce it.
+	// EventCommit marks a transaction commit boundary, emitted by the StreamParser
+	// at an XID_EVENT and — as a catch-all for implicitly-committed statements that
+	// carry a GTID but have no XID and aren't table DDL (GRANT, CREATE DATABASE,
+	// CREATE VIEW, XA COMMIT, etc.) — when the next transaction's GTID_EVENT
+	// arrives. It carries no row data, only the committed transaction's GTID. The
+	// consumer advances the durable GTID checkpoint ONLY on this event (and on
+	// EventDDL), never on the leading EventGTID, so a checkpoint can never claim a
+	// half-streamed transaction (#491). The file parser does not produce it.
 	EventCommit EventType = 7
 )
 
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -37,13 +37,16 @@ const (
 	// same ResultRow pipeline as real binlog events.
 	EventSnapshot EventType = 6
 	// EventCommit marks a transaction commit boundary, emitted by the StreamParser
-	// at an XID_EVENT and — as a catch-all for implicitly-committed statements that
-	// carry a GTID but have no XID and aren't table DDL (GRANT, CREATE DATABASE,
-	// CREATE VIEW, XA COMMIT, etc.) — when the next transaction's GTID_EVENT
-	// arrives. It carries no row data, only the committed transaction's GTID. The
-	// consumer advances the durable GTID checkpoint ONLY on this event (and on
-	// EventDDL), never on the leading EventGTID, so a checkpoint can never claim a
-	// half-streamed transaction (#491). The file parser does not produce it.
+	// at an XID_EVENT (InnoDB DML) and — as a catch-all when the next transaction's
+	// GTID_EVENT arrives — for transactions that carry a GTID but emit no XID and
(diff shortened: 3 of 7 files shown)


nethalo:
### Review applied (multi-agent: code / silent-failure / tests / comments)

Follow-up commit `e130039` addresses a **real gap** the review caught, plus disclosure/test-hygiene items:

- **CRITICAL (fixed):** the original commit detection (XID + exact \`COMMIT\` + table DDL) missed GTID-bearing **implicitly-committed statements** with no XID and not table DDL — \`GRANT/REVOKE\`, \`CREATE/DROP DATABASE\`, \`CREATE/DROP VIEW/TRIGGER/PROCEDURE/FUNCTION\`, \`CREATE/DROP INDEX\`, \`ANALYZE/OPTIMIZE TABLE\`, \`XA COMMIT\`. Their GTID never advanced → endless re-streaming and eventually a **false permanent-data-loss gap alarm** for a zero-row transaction. Replaced the brittle string match with a robust catch-all: **commit the in-flight GTID when the next transaction's \`GTID_EVENT\` arrives** (plus immediate commit at XID, and \`EventDDL\` for table DDL). \`emitCommit\` clears \`currentGTID\` so it never double-commits; no statement enumeration, and it can't fire mid-transaction (SAVEPOINT/BEGIN no longer matter).
- **Test hygiene (fixed):** the live test now registers the \`gtid_mode\` restore \`t.Cleanup\` **before** the transition, so a partial enable can't leak \`gtid_mode\` to other tests on the shared server.
- **Comments (fixed):** \`emitCommit\` position-mode wording, \`EventCommit\` doc, \`byos.SplitEvent\` doc; corrected the stale \`emptyPayloadNoEffect\` premise.
- **New tests:** parser unit tests for XID commit, implicit-commit-via-next-GTID (GRANT), and compressed-transaction commit; a streamLoop DDL GTID-accumulation subtest.

### At-least-once disclosure
Re-streamed transactions on an ungraceful crash produce duplicate \`binlog_events\` rows with **new \`event_id\`s**, which \`query.MergeResults\` (dedup-by-\`event_id\`) does **not** collapse. This is pre-existing pipeline behavior (\`flush()\` on batch-full writes durably without checkpointing), not introduced here — this PR only removes a silent-**loss** window. Exactly-once is tracked as a follow-up.

All green: unit (28 pkgs), \`internal/streamrun\` + \`internal/parser\` + \`cmd/bintrail\` integration, E2E; \`gtid_mode\` restored to OFF after the suite.

nethalo:
### Second review round applied (commit \`ca5fa3d\`)

A re-review of the redesign (4 agents) found **no correctness bugs** — `code-reviewer` verified all paths safe with primary-source evidence (idempotent `MysqlGTIDSet.Update`, forward-monotonic `binlogPos`, fallback can't fire mid-transaction, compressed-path ordering load-bearing & correct). Applied precision + hardening:

- **Comments:** corrected the fallback statement-list — a normal InnoDB `COMMIT` ends in an XID and does **not** reach the next-GTID fallback; the fallback covers implicitly-committed DDL/DCL + no-XID explicit terminators (`XA COMMIT`, non-transactional `COMMIT`). Documented the trailing-transaction tail-lag (a last non-XID statement before idle/shutdown re-streams on restart — harmless, no rows).
- **Code:** `advanceGTID`'s `Update` failure escalated `slog.Warn` → `slog.Error` (it's the sole durable-progress path; a frozen-but-green checkpoint is the worst failure mode).
- **Tests:** assert the commit **GTID** on the two commit-emission unit tests; added compressed-transaction-with-rows (rows + XID in one payload), no-double-commit-after-XID, trailing-implicit-commit-not-emitted, position-mode-commit-boundary, and a **live `ANALYZE`-interleave** that proves the next-GTID fallback end-to-end against a real GTID binlog (robust set-coverage assertion).

All green: unit (28 pkgs; one unrelated `internal/shim` flake confirmed passing in isolation), `internal/streamrun` + `internal/parser` integration, E2E; `gtid_mode` restored to OFF.

PR #499, opened 2026-06-14
baseline: parse every row of multi-line mydumper INSERTs (fix #495 silent data loss)

### LINKED ISSUE #495: baseline: only the first row of each multi-row INSERT is parsed — silent data loss on mydumper >=1.x dumps
## Summary

`bintrail baseline` silently keeps only the **first row of each multi-row `INSERT` statement** when converting a mydumper SQL dump, dropping every continuation row. On a real-sized table this loses nearly all data while the command reports success.

## Repro (MySQL 8.4 source, 1,000,000-row table)

```sh
bintrail dump --source-dsn "$SRC" --output-dir /tmp/dump --tables app.orders
# mydumper (the default mydumper/mydumper:latest is now v1.0.3) writes multi-row INSERTs:
#   INSERT INTO `orders` (...) VALUES(row1)
#   ,(row2)
#   ,(row3)
#   ... ;
bintrail baseline --input /tmp/dump --output /tmp/baseline
#  -> "baseline complete   rows: 159"
```

The source table has 1,000,000 rows. The baseline parquet ends up with exactly **159** rows — which is the exact number of `INSERT` **statements** across the dump chunks (39+45+40+35). Verified:

```sh
bintrail reconstruct --sql "SELECT COUNT(*) FROM parquet_scan('/tmp/baseline/*/app/orders.parquet')"
# -> 159
```

## Root cause

`baseline`'s SQL parser reads the row on the `VALUES(...)` line and ignores the `,(...)` continuation lines of a multi-row `INSERT`. mydumper >= 1.0 emits multi-row INSERTs by default, so one INSERT statement carries thousands of rows but only the first is captured.

## Impact

Time-travel / `reconstruct` (single-row and full-table) is built on the baseline. A baseline missing ~99.98% of rows silently breaks reconstruction — single-row reconstruct of a non-captured row returns empty; full-table reconstruct produces a near-empty dump. The only signal is the `rows:` count in the log, which a user has no reason to cross-check against the source.

This is silent data loss in the foundation of the time-travel feature, on the default `bintrail dump` -> `bintrail baseline` pipeline at any real scale.

## Workaround

Force TSV output: `mydumper --format LOAD_DATA` (one row per line), which `baseline` parses correctly (1,000,015 rows in our test). Note the TSV path is not flawless either — it produced 15 NULL-`id` rows out of 1e6 (~0.0015%, likely a chunk-boundary / blank-line parse), enough to abort a full-table `reconstruct` with `canonicalizePKValue: nil PK value for column "id"`. So both baseline input parsers have fragility; the multi-row-INSERT one is catastrophic, the TSV one is minor.

## Suggested fixes

- `bintrail baseline` should parse multi-row `INSERT ... VALUES (...),(...),...;` tuples (not just the first), **or**
- `bintrail dump` should emit a baseline-compatible format by default (`--format LOAD_DATA`, or `--statement-size` forcing one row per INSERT), **and**
- at minimum, `baseline` should cross-check the parsed row count against mydumper's metadata and **fail loudly** on a large mismatch instead of reporting success.

## Environment

- bintrail 0.13.4 (linux/amd64 release tarball)
- MySQL 8.4.9 source (binlog_format=ROW, binlog_row_image=FULL)
- mydumper/mydumper:latest = v1.0.3

Found while benchmarking dbtrail vs MariaDB system-versioned tables for data-recovery performance.

nethalo:
closes #495

## Summary

`bintrail baseline` silently kept only the **first row of each multi-row `INSERT`** when converting a mydumper SQL dump. mydumper >= 1.0 emits one tuple per physical line:

```sql
INSERT INTO `orders` (...) VALUES(row1)
,(row2)
,(row3);
```

`ReadSQLFile` was line-oriented: for each line it required an `INSERT` prefix, found ` VALUES`, and parsed tuples **only from that one line**. The `,(rowN)` continuation lines don't start with `INSERT`, so they were skipped — a 1,000,000-row table produced a baseline with exactly one row per `INSERT` *statement* (159 in the report). Time-travel / `reconstruct` is built on the baseline, so this silently broke recovery while the command reported success.

## Fix

- **Stateful parse across lines** (`internal/baseline/reader_sql.go`): while inside a not-yet-terminated `INSERT`, continuation lines are parsed as further tuples until the `;` terminator is consumed. `parseSQLTuples` now returns `(terminated bool, err)` so `ReadSQLFile` knows when the (possibly multi-line) statement ends — this is exact regardless of where the `;` sits (own line, trailing a tuple, after whitespace), unlike a `HasSuffix(line, ";")` heuristic.
- Splitting on physical lines stays safe because mydumper escapes embedded newlines inside string values as the two chars `\n` (never a raw newline byte), so a line boundary always falls between tuples.
- **Fail loud on truncation**: a file that ends mid-statement (no `;`) now errors instead of reporting a silently short row count. This is defense-in-depth for genuine truncation — *not* the fix for #495 itself (the dropped rows came from terminated statements).

## Tests

New regression tests in `internal/baseline/baseline_test.go`, all **verified to fail on the pre-fix code**:
- `TestReadSQLRowMultiLineInsert` — the issue's exact format (2 statements × 4+2 tuples → 6 rows; old code returned 2).
- `TestReadSQLRowMultiLineLayouts` — leading-comma / trailing-comma / `;`-on-own-line / `VALUES`-then-tuples / single-line / two-statements all yield identical rows.
- `TestReadSQLRowTruncated` — unterminated file → error.

## Scope notes (from an adversarial probe of the parser)

- **Did NOT** implement the issue's suggested mydumper-metadata row-count cross-check: mydumper metadata has no reliable per-table count across versions, so a fail-on-mismatch would produce false failures. The parser fix + truncation detection address the silent-loss root cause directly.
- An adversarial sweep surfaced **separate, pre-existing** baseline-parser bugs unrelated to multi-row parsing — binary/`BLOB`/`BINARY(16)`/JSON columns are silently corrupted on default mydumper output (quote-blind value reader), and `REPLACE INTO` dumps are silently dropped. These are **out of scope here** and handled in a follow-up PR.
- The minor TSV `\N`-id fragility noted in the issue (the workaround path) is not addressed — not reproducible without the source data.

## Test plan
- [x] New regression tests proven to fail on pre-fix code
- [x] `gofmt` / `go vet` clean

### DIFF
--- a/internal/baseline/baseline_test.go
+++ b/internal/baseline/baseline_test.go
@@ -544,6 +544,133 @@ func TestReadSQLRowUnterminated(t *testing.T) {
 	}
 }
 
+func TestReadSQLRowMultiLineInsert(t *testing.T) {
+	// mydumper >= 1.0 emits one tuple per physical line with a leading comma.
+	// The old line-oriented parser captured only the first tuple of each
+	// INSERT and silently dropped every continuation row — issue #495. This
+	// asserts every row of a multi-line, multi-statement dump is returned.
+	const sqlData = "INSERT INTO `orders` (`id`,`name`) VALUES(1,'a')\n" +
+		",(2,'b')\n" +
+		",(3,'c')\n" +
+		",(4,'d');\n" +
+		"INSERT INTO `orders` (`id`,`name`) VALUES(5,'e')\n" +
+		",(6,'f');\n"
+	dir := t.TempDir()
+	path := filepath.Join(dir, "shop.orders.00000.sql")
+	if err := os.WriteFile(path, []byte(sqlData), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	var rows [][]string
+	if err := ReadSQLFile(path, func(values []string, nulls []bool) error {
+		rows = append(rows, append([]string(nil), values...))
+		return nil
+	}); err != nil {
+		t.Fatalf("ReadSQLFile: %v", err)
+	}
+	// Two statements carrying 4 + 2 = 6 tuples total.
+	if len(rows) != 6 {
+		t.Fatalf("got %d rows, want 6 (multi-line continuation rows were dropped)", len(rows))
+	}
+	for i, want := range []string{"1", "2", "3", "4", "5", "6"} {
+		if rows[i][0] != want {
+			t.Errorf("row %d id = %q, want %q", i, rows[i][0], want)
+		}
+	}
+	// Spot-check continuation-row values, not just ids.
+	if rows[3][1] != "d" || rows[5][1] != "f" {
+		t.Errorf("continuation values wrong: row3=%v row5=%v", rows[3], rows[5])
+	}
+}
+
+func TestReadSQLRowMultiLineLayouts(t *testing.T) {
+	// Every comma/terminator placement mydumper variants emit must yield the
+	// same three rows.
+	layouts := map[string]string{
+		"leading-comma":      "INSERT INTO t VALUES(1)\n,(2)\n,(3);\n",
+		"trailing-comma":     "INSERT INTO t VALUES(1),\n(2),\n(3);\n",
+		"semicolon-own-line": "INSERT INTO t VALUES(1)\n,(2)\n,(3)\n;\n",
+		"values-then-tuples": "INSERT INTO t VALUES\n(1),\n(2),\n(3);\n",
+		"single-line":        "INSERT INTO t VALUES(1),(2),(3);\n",
+		"two-statements":     "INSERT INTO t VALUES(1);\nINSERT INTO t VALUES(2)\n,(3);\n",
+	}
+	for name, sql := range layouts {
+		t.Run(name, func(t *testing.T) {
+			dir := t.TempDir()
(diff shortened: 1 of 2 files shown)


nethalo:
Updated after a review round (code-reviewer + silent-failure-hunter, both run on the committed diff). They independently converged on one hardening: the lenient `break` in `parseSQLTuples` on an unexpected token was a residual silent-loss path — with the new cross-line state it would swallow following statements as bogus continuations and drop their rows, while a later `;` hides the loss from the EOF guard.

Folded the fix in: an unexpected token where a tuple/terminator is expected is now a loud `unexpected token` error. It cannot fire on valid mydumper output (after a tuple only `,`, whitespace, or `;` appears). Added `TestReadSQLRowUnexpectedToken` covering both reviewers' repros, verified to fail on the pre-hardening code.

An adversarial probe (8 agents) of the parser also surfaced **separate pre-existing** bugs — `_binary`/JSON columns corrupted on default mydumper output, `REPLACE INTO` silently dropped, `\Z` escape lost. Those are out of scope for this PR (different root cause in `parseSQLValue`/dispatch) and are addressed in a follow-up.

PR #501, opened 2026-06-14
baseline: decode mydumper binary/JSON values + accept REPLACE INTO

nethalo:
Follow-up to #495 (stacked on its branch — base auto-retargets to `main` when #499 merges). Separate root cause from the multi-row INSERT fix.

## Problem (found by an adversarial probe of the parser)

On **default** mydumper/mysqldump output, the baseline SQL reader silently corrupted several common column types — verified against real mydumper v1.0.3:

- **BINARY/VARBINARY/BLOB/BIT/GEOMETRY** dump as `_binary "<bytes>"` (charset introducer). The value starts with `_`, not a quote, so it fell to the quote-blind default reader — a raw `,` or `)` byte inside the value (near-certain for any real blob, e.g. a `BINARY(16)` UUID) split the row into the wrong columns, or aborted the whole file.
- **JSON** dumps as `CONVERT("<json>" USING utf8mb4)` — stored as the literal wrapper text instead of the JSON document.
- **`REPLACE INTO`** (mysqldump `--replace` / mydumper `--replace`) was skipped entirely by the `INSERT`-only dispatch → every row silently dropped.
- The `\Z` (Ctrl-Z, `0x1A`) and `\0` escapes were missing from the double-quote parser → binary bytes `0x1A`→`'Z'`, `0x00`→`'0'`.

All are pre-existing (not introduced by #495); the probe surfaced them while verifying the multi-row fix.

## Fix (`internal/baseline/reader_sql.go`)

- `parseSQLValue`: parse MySQL charset introducers (`_binary`/`_utf8mb4`/…) and the `CONVERT("…" USING …)` JSON wrapper via the real quoted-string parsers, returning the decoded bytes / JSON. Falls back to the default reader when the structure doesn't match (numbers/keywords unaffected).
- `parseSQLString` / `parseSQLDoubleString`: add the `\0` and `\Z` escapes.
- `ReadSQLFile`: accept `REPLACE INTO` alongside `INSERT`.

## Verification

End-to-end with **real mydumper v1.0.3**: a table with VARBINARY/BLOB/BIT/JSON columns holding adversarial bytes (`,` `)` quotes, NUL, `0x1A`) round-trips **byte-exact** through `baseline` → Parquet → read-back:

| id | vb | bl | bt | js |
|---|---|---|---|---|
| 1 | `612C6229` | `2722` | `AA` | `{"m": "done) here", "n": 5}` |
| 2 | `001A0D0A5C` | `48656C6C6F` | `01` | `[1, 2, 3]` |

The captured dump is committed as a test fixture (`testdata/mydumper_v1_binary_json.sql`). Every new test was confirmed to **fail on the pre-fix parser**.

## Not addressed here
- `" VALUES"` substring false-match on an identifier containing an embedded space ending on a VALUES-aligned boundary (e.g. `` `Allowed Values` ``) — low realism (contrived naming), needs a quote-aware keyword scan. Noted for a separate change.

## Test plan
- [x] `go test ./... -count=1` green; new tests fail on pre-fix parser
- [x] Real-mydumper round-trip byte-exact (baseline → Parquet → read-back)
- [x] `gofmt` / `go vet` clean


nethalo:
**Merge note (stacked PR + squash):** this branch is based on `worktree-issue-495` and contains its commit. When #499 is squash-merged into `main`, GitHub auto-retargets this PR to `main` but does **not** rebase the branch, so the diff may show a conflict in `reader_sql.go` (the squash creates a new SHA for the #495 change). After #499 merges, drop the now-redundant first commit:

```
git checkout parser-hardening
git rebase --onto main worktree-issue-495 parser-hardening
git push --force-with-lease
```

Added a committed end-to-end test (`TestRunMydumperBinaryJSONRoundTrip`) that drives `baseline.Run` over the real mydumper fixture and asserts the Parquet bytes — so the binary/JSON round-trip is now permanent coverage, not just a one-time manual check (it also exercises `convertValue` + `MysqlToParquetNode`, the schema-mapping hop).

nethalo:
**Updated after a 4-agent review** (code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter). code-reviewer and silent-failure-hunter came back clean (no new silent-failure path; new code fails loud, falls back only on shapes mydumper never emits). Applied:

- Fixed a fabricated flag name: `mydumper --insert-into-replace` → **`mydumper --replace`** (verified against the `mydumper/mydumper:v1.0.3-1` binary; no such `--insert-into-replace` option exists). In the dispatch comment, the REPLACE test comment, and this PR body.
- Dropped the unverified `GEOMETRY` claim from the introducer comment (BINARY/VARBINARY/BLOB are fixture-verified; geometry isn't).
- Recorded the fixture's exact generation command in the test doc-comment (provenance).
- Added a single-quoted-inner-literal case to `TestReadSQLRowConvertJSON` (covers the `parseConvertExpr` `case '\''` branch).

Filed #503 for the **pre-existing, non-default** items the review surfaced (out of scope here): hex-blob `0x…` binary stored as ASCII, GEOMETRY type mapping, `convertValue` silent NULL-on-error, and a `WriteRow` column-count guard (with the generated-column caveat).

nethalo:
Auto-closed by GitHub when the stacked base branch `worktree-issue-495` was deleted on #499's merge (GitHub can't retarget/reopen a PR whose base branch is gone). Continued as #504 — same commit, rebased onto `main`.

PR #505, opened 2026-06-14
fix: preserve integers >2^53 on the row-image read path (#496)

### LINKED ISSUE #496: Data corruption: recover/query read path rounds integers >2^53 via JSON float64
## Summary

`row_before`/`row_after` are stored as JSON. The recover/query read path unmarshals them into `map[string]any`, which turns **every** JSON number into `float64`. Any integer with `|value| > 2^53` is silently rounded on read, so `recover`/`query` output — including the regenerated reversal SQL `WHERE`/`SET` — is wrong even when the stored value is exact.

## Evidence

- `internal/query/query.go:472-476` — `json.Unmarshal(rowBefore, &r.RowBefore)` into `map[string]any` (unmarshal error discarded). Standard `encoding/json` decodes numbers as `float64`.
- `internal/recovery/recovery.go` `FormatSQLValue` — the integer-valued-float branch is guarded by `~1e15`; values beyond that fall through to `strconv.FormatFloat`, emitting the rounded float. (`FormatSQLValue` already has `uint64`/`uint32` branches, but they are unreachable because the scan produces `float64`.)

## Impact

Tightly coupled to #490. A `BIGINT UNSIGNED` value with the high bit set is always `> 2^63 ≥ 2^53`, so **every** value #490 targets in the BIGINT case still regenerates **wrong** in `recover` SQL despite #490 fixing the stored value and `pk_values`. Also affects large **signed** `BIGINT > 2^53` regardless of sign.

The `reconstruct`/DuckDB path reads native `int64` and is unaffected — the bug is specific to the `binlog_events` JSON read path.

Example: a row with `id BIGINT UNSIGNED = 18446744073709551615` regenerates as `... = 18446744073709552000` (rounded) → matches no row.

## Suggested fix

Decode `row_before`/`row_after` with `json.Decoder` + `UseNumber()` wherever they're read for recover/query, and have `recovery.FormatSQLValue` handle `json.Number` (emit the integer literal directly; fall back to float formatting only for genuinely fractional values). The already-present `uint64`/`uint32` branches then become reachable.

## References

- 2026-06-14 binlog-fidelity audit (read-side precision finding).
- Sharpened by the #494 review: this is the gap that prevents #490 from fully closing `BIGINT UNSIGNED` end-to-end. Companion to the BIT(64) issue (which also hits this path for `BIT(64) > 2^53`).

nethalo:
closes #496

## Summary
- `row_before`/`row_after` JSON was unmarshalled into `map[string]any` with default `encoding/json` → every number became `float64`, silently rounding integers above 2^53 (`BIGINT UNSIGNED > 2^63`, large signed `BIGINT`). `recover` SQL and `query`/CSV output emitted the wrong value even though storage was exact — leaving #490 incomplete end-to-end for BIGINT.
- New `query.UnmarshalRowImage` (`json.Decoder` + `UseNumber()`) used at **both** read sites: `internal/query` (live MySQL) and `internal/parquetquery` (archived Parquet — `recover` merges both). Numbers now arrive as `json.Number` (the exact literal).
- Consumers updated:
  - `recovery.FormatSQLValue`: emit `json.Number` verbatim (**additive** — `float64` case stays for genuine DOUBLEs from the DuckDB reconstruct path).
  - `metadata.ordinalValue`: parse `json.Number` for ENUM/SET label resolution (full 64-bit, no 2^53 cap).
  - **shim**: `BuildSimpleTextResultset` rejects `json.Number`, so `resultsetValue` maps it to `int64`/`uint64`/`float64` (exact, incl. `BIGINT UNSIGNED > 2^63`) at the three resultset sites; `fullTableTextCell` renders the exact literal.
- Unaffected (verified): `query` JSON/CSV, MCP, console, and the archive **write** path use `json.Marshal` or raw-JSON passthrough — `json.Number` marshals correctly.

## Note
This is the **read** half. Combined with #490 (the merged **write** fix — `marshalRow` stores `uint64`→exact-integer JSON), `BIGINT UNSIGNED` is now correct end-to-end through `recover`. The integration test below relies on that storage guarantee.

## Test plan
- [x] Unit: `FormatSQLValue(json.Number)`, `UnmarshalRowImage` (>2^53 exact), `ordinalValue(json.Number)`, `resultsetValue(json.Number)`.
- [x] Integration: `BIGINT UNSIGNED` max → `recover --dry-run` emits the exact value (not the rounded `...616`).
- [x] Full unit (28 pkgs) + integration: `recovery`/`query`/`parquetquery`/`reconstruct`/`metadata`/`shim`/`console`/`cmd/bintrail`/`cmd/bintrail-mcp` + E2E — all green. (The shim path initially broke with "unsupport type json.Number"; fixed and covered.)

### DIFF
--- a/internal/metadata/enumlabel.go
+++ b/internal/metadata/enumlabel.go
@@ -2,7 +2,9 @@ package metadata
 
 import (
 	"database/sql"
+	"encoding/json"
 	"math"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -237,6 +239,16 @@ func (s *EnumMapperSource) memoized(key mapperKey, r *Resolver) *EnumLabelMapper
 // honest call.
 func ordinalValue(v any) (uint64, bool) {
 	switch n := v.(type) {
+	case json.Number:
+		// Row images decoded by query.UnmarshalRowImage keep numbers as
+		// json.Number (#496). Parse the exact integer literal — no 2^53 float
+		// limit, so SET masks using high bits resolve correctly. A negative,
+		// fractional, or out-of-range value fails ParseUint → pass-through.
+		u, err := strconv.ParseUint(n.String(), 10, 64)
+		if err != nil {
+			return 0, false
+		}
+		return u, true
 	case float64:
 		if n < 0 || n != math.Trunc(n) || n > 1<<53 {
 			return 0, false
--- a/internal/metadata/enumlabel_test.go
+++ b/internal/metadata/enumlabel_test.go
@@ -1,11 +1,30 @@
 package metadata
 
 import (
+	"encoding/json"
 	"errors"
 	"reflect"
 	"testing"
 	"time"
 )
+
+func TestOrdinalValue_jsonNumber(t *testing.T) {
+	// Row images decoded by query.UnmarshalRowImage carry numbers as json.Number
+	// (#496); ENUM/SET label resolution must still extract the ordinal.
+	if u, ok := ordinalValue(json.Number("3")); !ok || u != 3 {
+		t.Errorf("ordinalValue(json.Number(\"3\")) = (%d,%v), want (3,true)", u, ok)
+	}
+	// Full 64-bit range (a SET mask using a high bit) — no 2^53 float cap.
+	if u, ok := ordinalValue(json.Number("9223372036854775808")); !ok || u != 9223372036854775808 {
+		t.Errorf("ordinalValue(high-bit) = (%d,%v), want (9223372036854775808,true)", u, ok)
+	}
+	// Negative, fractional, and non-numeric → pass-through (!ok).
+	for _, bad := range []json.Number{"-1", "1.5", "abc"} {
+		if _, ok := ordinalValue(bad); ok {
+			t.Errorf("ordinalValue(json.Number(%q)) reported ok, want !ok", bad)
+		}
+	}
+}
 
 func TestParseEnumSetLabels(t *testing.T) {
 	tests := []struct {
--- a/internal/parquetquery/parquetquery.go
+++ b/internal/parquetquery/parquetquery.go
@@ -904,10 +904,10 @@ func scanRows(rows *sql.Rows) ([]query.ResultRow, error) {
 			_ = json.Unmarshal([]byte(changedCols.String), &r.ChangedColumns)
 		}
 		if rowBefore.Valid && rowBefore.String != "" {
-			_ = json.Unmarshal([]byte(rowBefore.String), &r.RowBefore)
+			r.RowBefore = query.UnmarshalRowImage([]byte(rowBefore.String))
 		}
 		if rowAfter.Valid && rowAfter.String != "" {
-			_ = json.Unmarshal([]byte(rowAfter.String), &r.RowAfter)
+			r.RowAfter = query.UnmarshalRowImage([]byte(rowAfter.String))
 		}
 		results = append(results, r)
 	}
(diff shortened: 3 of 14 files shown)


nethalo:
### Review applied (4 agents) — caught a CRITICAL regression I introduced

The review found two related shim issues (and the advisor pre-flight caught that I'd skipped `cmd`/MCP integration — ran them, green):

- **CRITICAL (code-reviewer, reproduced):** `resultsetValue` mapped `json.Number`→narrowest type (`int64` vs `float64`), so a **DOUBLE column with mixed integral/fractional values across rows** made `BuildSimpleTextResultset` reject the whole `_flashback` full-table query (`"row types aren't consistent"`). Pre-fix worked by accident (uniformly `float64`→DOUBLE).
- **MEDIUM (silent-failure):** `fullTableTextCell` rendered the raw literal (`1e+21`) while baseline-origin cells rendered decimal via `FormatTextValue` → same DOUBLE column diverged across baseline vs event rows in a `_snapshot` merge.

**Unified fix:** a shared `numberToText` recovers the exact numeric type then renders via go-mysql's `FormatTextValue` (the same path baseline cells take). Columns become uniformly `VAR_STRING` (no crash), event/baseline rows are byte-identical, and `BIGINT UNSIGNED > 2^63` stays exact. `resultsetValue` + `fullTableTextCell` both route through it.

- **Comments (comment-analyzer):** fixed three pre-existing `float64` comments this PR made stale (`recovery.go` ×2, `snapshot.go`) + the `UnmarshalRowImage` handler list.
- **Tests (pr-test-analyzer):** regression for the mixed-float column (no crash + exact), shim wire round-trip of `BIGINT UNSIGNED` max, baseline/event byte-identity for extreme doubles, the 2^63 boundary; plus the flagged coverage gaps — recover UPDATE SET+WHERE, parquetquery archive round-trip, and a `json.Number` cell in the mydumper matrix.

All green: unit (28 pkgs) + integration `recovery`/`query`/`parquetquery`/`reconstruct`/`metadata`/`shim`/`console`/`cmd/bintrail`/`cmd/bintrail-mcp` + E2E.

nethalo:
### Second review round applied (commit \`ac7e611\`)

A re-review of the \`numberToText\` redesign (4 agents) confirmed the **CRITICAL crash fix is solid and introduces no new critical** — \`[]byte\`→\`VAR_STRING\` is charset-33 utf8 (not BLOB), \`numberToText\` is exact, DOUBLE/DECIMAL byte-identical, all \`BuildSimpleTextResultset\` callers audited. Applied the accuracy + coverage items:

- **Comments:** the "byte-identical / both origins preserve exactly" claims overstated — true for INT and DOUBLE, but **FLOAT(32-bit)** baseline cells widen (DuckDB scans `float32` → `FormatTextValue` → `0.10000000149011612` vs the event side's `0.1`). Scoped the claim to INT/DOUBLE and named the FLOAT exception + the baseline `BIGINT UNSIGNED > 2^63` storage limitation — both **pre-existing, baseline-side**, tracked in **#506**.
- **Tests:** documented the FLOAT(32) divergence (locks it); added negative large-int / negative extreme-double + a beyond-`float64` literal-passthrough case; and an **end-to-end `_snapshot` full-table DOUBLE merge** test (baseline-origin integral + event-origin fractional in one column → no crash + baseline/event byte-identity through `HandleQuery`).

The `DOUBLE→VAR_STRING` wire-type change is intentional/benign (text protocol behind ProxySQL; no consumer inspects column types) — it's what buys exact `BIGINT UNSIGNED` rendering.

All green: unit (28 pkgs) + `internal/shim` integration. Follow-up **#506** for the baseline-write-path numeric gaps.

Issue #506, opened 2026-06-15
Baseline write-path numeric fidelity: BIGINT UNSIGNED > 2^63 loss + FLOAT(32) render divergence


nethalo:
## Summary

The **baseline write path** (`bintrail baseline` / mydumper → Parquet, `internal/baseline`) has two numeric-fidelity gaps that the read-path fix (#496/PR #505) surfaced but does not address — both are baseline-side and independent of that diff. They matter for full-table reconstruction (`bintrail reconstruct` and the shim's `_snapshot` full-table merge), where a never-touched baseline row is rendered straight from the Parquet baseline.

## Gap 1 — `BIGINT UNSIGNED > 2^63` cannot round-trip through a baseline

`internal/baseline/schema.go` maps `bigint` → **signed** `parquet.Int(64)`, and the writer converts with `strconv.ParseInt(raw, 10, 64)` — which **errors** on `18446744073709551615` (the exact class #490/#496 target). So a `BIGINT UNSIGNED` value above 2^63 either fails the baseline build or is stored wrong/negative. After #496 an *event-origin* row renders such a value exactly (json.Number), but a *baseline-origin* (never-touched) row cannot.

**Fix direction:** store `bigint unsigned` as Parquet unsigned 64 (or as a string, like DECIMAL already is) and convert via `ParseUint`. Mirror the `coerceUnsigned`/`ColumnType` "unsigned" detection used on the index path (#490).

## Gap 2 — FLOAT (32-bit) columns render inconsistently baseline-vs-event

A baseline `FLOAT` column is stored as Parquet 4-byte and scanned back by DuckDB as Go **`float32`**; `mysql.FormatTextValue(float32(0.1))` widens to `float64(float32(0.1))` = **`0.10000000149011612`**, while the event side renders the shortest float32 literal **`0.1`** (which is what MySQL itself displays). So in one full-table result, an unchanged FLOAT cell and a post-baseline-event FLOAT cell of the same value render differently. (DOUBLE is fine — both sides are float64. This is pre-existing — present identically before #505.)

**Fix direction:** in the shim's `fullTableTextCell` default branch (and the equivalent `reconstruct` mydumper path), special-case `float32` via `strconv.FormatFloat(float64(x), 'f', -1, 32)` so the baseline side matches the (correct) event side / MySQL display.

## Notes
[...]

PR #507, opened 2026-06-15
fix: index BIT(64) with the high bit set as its correct unsigned value (#497)

### LINKED ISSUE #497: Data corruption: BIT(64) with the high bit set is indexed as a negative value
## Summary

`BIT(64)` columns whose high bit (bit 63) is set are indexed as a **negative** value. go-mysql decodes `BIT(N)` as `int64` (`decodeBit` casts `Uint64`→`int64`), and bintrail stores it verbatim. This is the same class of corruption as #490 (UNSIGNED integers), but `BIT` is a distinct type that cannot be declared `UNSIGNED`, so the #490 fix (`coerceUnsigned`, gated on `ColumnType` containing "unsigned") intentionally does **not** cover it.

## Evidence

- go-mysql v1.13.0 `replication/row_event.go` `decodeBit` → returns `int64`.
- `internal/metadata/metadata.go` `coerceUnsigned` gates on `strings.Contains(ColumnType, "unsigned")`; a BIT column's `COLUMN_TYPE` is `bit(N)`, so BIT is correctly excluded there (documented in the function comment).

## Impact

`BIT(64)` values with bit 63 set are stored as a negative `int64`; recovery/forensic output for those columns is wrong. `BIT(1..63)` are unaffected (the top bit is never set within `int64` range). Narrow but real data-fidelity gap; named in #490 as "the same problem."

## Suggested fix

Handle BIT columns in the same `MapRow` coercion layer — reinterpret the decoded `int64` as `uint64` (BIT is inherently an unsigned bit string). Decide the desired representation in `row_before`/`row_after` and `recover` output (numeric `uint64`, or a `b'...'` literal). Note the read-path `float64` issue (companion) also affects `BIT(64) > 2^53`, so both fixes are needed for full end-to-end correctness.

## References

- 2026-06-14 binlog-fidelity audit; flagged again in the #494 review. Companion to the read-path precision issue.

nethalo:
closes #497

## Summary
- go-mysql decodes `BIT(N)` as a signed `int64`, so a `BIT(64)` with the high bit (bit 63) set comes back **negative** and was indexed verbatim.
- `coerceUnsigned` now reinterprets a BIT column's `int64` as `uint64` — identity for `BIT(1..63)` (always positive within `int64`), fixing `BIT(64)`. The BIT branch runs **before** the unsigned-substring gate (a BIT column's `ColumnType` is `bit(N)`, no "unsigned").
- Stored as numeric `uint64`: `recover` emits it as an integer literal (valid for a BIT column) and the shim renders the integer. Combined with **#496** (json.Number read path, now in `main`), a `BIT(64)` value > 2⁵³ survives **exactly** end-to-end.

## Test plan
- [x] Unit: `coerceUnsigned` BIT cases (all-bits-set → `uint64` max, high-bit-only, `BIT(8)` positive unchanged, NULL passthrough); existing unsigned cases unaffected.
- [x] Integration: index a real `BIT(64)`/`BIT(8)` row through the parser → `RowAfter` is `uint64` (not negative).
- [x] Full unit (28 pkgs) + integration `metadata`/`parser`/`recovery`/`shim` green.

### DIFF
--- a/internal/metadata/metadata.go
+++ b/internal/metadata/metadata.go
@@ -253,6 +253,9 @@ func coerceUnsigned(v any, col ColumnMeta) any {
 		if i, ok := v.(int64); ok {
 			return uint64(i)
 		}
+		// go-mysql always decodes BIT as int64, so this never fires today. If a
+		// future go-mysql/MariaDB path ever delivered BIT as []byte/string, leave
+		// it uninterpreted rather than mis-coerce — the original value, not a bug.
 		return v
 	}
 	if !strings.Contains(strings.ToLower(col.ColumnType), "unsigned") {
--- a/internal/metadata/metadata.go
+++ b/internal/metadata/metadata.go
@@ -242,10 +242,19 @@ func (r *Resolver) MapRow(schema, table string, row []any) (map[string]any, erro
 // snapshots can't express signedness — NewResolver warns once in that case), or
 // when the value is not a signed integer (NULL, string, and DECIMAL/FLOAT/DOUBLE
 // UNSIGNED — which go-mysql returns as string/float — are returned unchanged).
-// BIT is intentionally gated out: it is a distinct type, never declared
-// "unsigned". (BIT(64) with the high bit set has the same underlying corruption,
-// since go-mysql decodes BIT as int64, but is tracked separately.)
+// BIT is also reinterpreted here: go-mysql decodes it as a signed int64, so a
+// BIT(64) with the high bit set comes back negative; it's mapped to uint64 (#497).
 func coerceUnsigned(v any, col ColumnMeta) any {
+	// BIT is an unsigned bit string. go-mysql decodes BIT(N) as int64, so BIT(64)
+	// with the high bit set comes back negative; reinterpret as uint64 — identity
+	// for BIT(1..63) (always positive within int64). BIT's ColumnType is "bit(N)"
+	// (no "unsigned"), so handle it before the unsigned gate below.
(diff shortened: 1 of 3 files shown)


nethalo:
### Review applied (3 agents) — approve, no blocking issues

- **code-reviewer:** approve, no issues at confidence ≥80. Minimal, correctly placed before the unsigned gate, exact `DataType=="bit"` match, correct `uint64` for all widths; rides the proven uint64/json.Number path through every consumer (indexer write, recover SQL, shim, query, reconstruct); BIT-as-PK consistent.
- **pr-test-analyzer:** coverage adequate — BIT converges to `uint64` at `coerceUnsigned`, so downstream is already locked by #490/#496; the two new tests cover the only BIT-specific behavior (parse→uint64), and the integration test is type-strict.
- **silent-failure-hunter:** no blocking silent failure; storage round-trip, recover SQL, and shim wire all lossless for `BIT(64)` max. Applied its one cheap item: a comment asserting the BIT-is-always-int64 invariant on the non-int64 fallthrough (commit `4618f55`).

Pre-existing items routed to follow-up **#506**: baseline-origin BIT renders `X'hex'` vs the delta integer (both lossless-valid; baseline-write-path, same theme as the unsigned/FLOAT items). The pk_values representation change for a BIT-PK across the upgrade is inherent to the fix (new indexing only; BIT-PK is rejected by reconstruct / warned by the shim).

All green: unit (28 pkgs) + integration `metadata`/`parser`/`recovery`/`shim`.

PR #508, opened 2026-06-15
docs: document binlog_row_image=FULL (server-wide) + no-PARTIAL_JSON source requirements (#492)

### LINKED ISSUE #492: docs: document binlog_row_image=FULL (server-wide) and no-PARTIAL_JSON as hard source requirements
## Summary

`binlog_row_image=FULL` is a hard requirement for dbtrail, but it isn't documented prominently as a **server-wide** requirement, and the adjacent `binlog_row_value_options` constraint isn't documented at all. Sources running partial row images (`MINIMAL`/`NOBLOB`) or `binlog_row_value_options=PARTIAL_JSON` produce data bintrail cannot faithfully index — and today that degradation is silent.

## Why it matters

- Under `binlog_row_image=MINIMAL`, an UPDATE's before-image carries only the PK and the after-image only the changed columns. go-mysql pads the absent columns to full width with `nil`, so unchanged columns — **and the after-image PK** — come back as `NULL`. Reproduced on MySQL 8.0.46: `UPDATE m SET a='aNEW' WHERE id=1` yields `RowAfter = {id:nil, a:"aNEW", b:nil, c:nil}`. A later `recover` would NULL the unchanged columns and build `WHERE id = NULL` (matches nothing).
- `binlog_row_value_options=PARTIAL_JSON` makes UPDATEs carry only a JSON **diff**, not the full document — so there is no complete after-image to recover from.
- **`binlog_row_image` is session-settable**, so a server global of `FULL` does not by itself guarantee every event is FULL (see the companion enforcement-hardening issue).

This is **out of scope by design** — dbtrail does not take responsibility for data quality under non-FULL configs — but the requirement must be explicit so operators can comply, and so support triage can cite it.

## Action

Add an explicit "Source requirements" section (e.g. `docs/streaming.md` / `docs/indexing.md` / `docs/guide.md`) stating:

- `binlog_format = ROW`
- `binlog_row_image = FULL` — **server-wide**; per-session overrides to `MINIMAL`/`NOBLOB` are unsupported and produce inaccurate index data.
- `binlog_row_value_options` must **not** include `PARTIAL_JSON`.

Cross-reference `SUPPORT.md` for the support boundary.

nethalo:
closes #492

## Summary
Two source-config facts were missing from the docs and silently produce inaccurate captures (surfaced by the binlog-fidelity audit):

- **`binlog_row_image=FULL` must be server-wide.** The preflight/`doctor` read the *global*, but it's session-settable — a per-session `SET SESSION binlog_row_image=MINIMAL`/`NOBLOB` writes partial images that dbtrail indexes as if complete (unchanged columns and the after-image PK come back NULL → `recover` destroys data / matches nothing).
- **`binlog_row_value_options` must not include `PARTIAL_JSON`** (logs only a JSON diff, no full after-image) — was undocumented anywhere.

Added to the canonical operator spots only (no redundancy across the ~6 places FULL is mentioned in passing):
- `docs/deployment.md` "Source MySQL Requirements" (detailed)
- `docs/streaming.md` (the requirement note)
- `docs/guide.md` (pre-start checklist)
- `SUPPORT.md` — new "Source server configuration" section making partial images / `PARTIAL_JSON` explicitly **out of scope**, so triage can cite it.

## Test plan
- [x] Docs-only — no code changed; `go test ./... -count=1` passes (28 pkgs).
- [x] Cross-doc links resolve; `#2-source-mysql-requirements` anchor matches the heading.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -83,9 +83,31 @@ The boundary triage cites:
   and remove the bundled `index-init` + `index-mysql` services (see
   [docker.md](docs/docker.md)) — same ship-vs-operate split, your server.
 
+## Source server configuration (required for correct capture)
+
+dbtrail reads ROW-format binary logs from your **source** MySQL. Faithful capture
+requires the source to be configured **server-wide** (not just per-session):
+
+- `binlog_format = ROW`.
+- `binlog_row_image = FULL`. Partial images (`MINIMAL`/`NOBLOB`) — **including from
+  a per-session `SET SESSION binlog_row_image = MINIMAL`** while the global is `FULL`
+  — are **out of scope**: dbtrail indexes incomplete before/after images as if
+  complete, so `recover` emits NULLs for unchanged columns and its `WHERE` clause
+  matches nothing.
+- `binlog_row_value_options` must **not** include `PARTIAL_JSON` — partial JSON
+  updates log only a diff, leaving no complete after-image to recover from.
+
+dbtrail's startup preflight and `bintrail doctor` validate the **global**
+`binlog_row_image`; preventing per-session overrides is the operator's
+responsibility. Data captured while the source violated these requirements is out
+of scope — configure the source (see
+[deployment.md → Source MySQL Requirements](docs/deployment.md#2-source-mysql-requirements))
+and re-index.
+
 ## Reporting issues
 
 Bugs in dbtrail's binaries, schema, tooling, console, or docs: please open
 an issue with reproduction steps — those are always in scope. If your report
 is about the index MySQL server's own operation (disk, backups, upgrades,
-corruption), see the list above first.
+corruption), or about data captured under an unsupported source configuration
+(non-`FULL` row image, `PARTIAL_JSON`), see the lists above first.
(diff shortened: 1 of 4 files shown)


nethalo:
### Review applied (comment-analyzer) — caught a technical inaccuracy

Fixed in `f419fa0`:
- **NOBLOB ≠ MINIMAL:** I'd lumped `NOBLOB` with `MINIMAL`'s catastrophic consequence. Only `MINIMAL` omits the after-image PK (→ `recover` WHERE matches nothing); `NOBLOB` keeps the PK and all non-BLOB/TEXT columns, dropping only unchanged BLOB/TEXT (a reversal can NULL those — milder). Also removed the self-contradiction ("destroys data **and** WHERE matches nothing").
- **"reads the global":** the check is session-scope `SHOW VARIABLES` on bintrail's own connection (equals global since bintrail never `SET SESSION`s); per-session overrides escape because bintrail only sees its own session, not via a global read. Reworded across deployment.md / SUPPORT.md / streaming.md.

Verified accurate (no change): session-settability of `binlog_row_image`, the MINIMAL→WHERE-matches-nothing mechanism (against `recovery.go`), PARTIAL_JSON semantics, all cross-doc links + the `#2-source-mysql-requirements` anchor.

Issue #517, opened 2026-06-19
Epic: MariaDB-as-source — alpha → beta


nethalo:
Tracking issue for taking **MariaDB-as-source** from **alpha** (shipped in #515) to **beta**.

bintrail is a data-*recovery* tool, so the bar for beta is: the **silent-data-loss gaps closed**, the **advertised surfaces covered**, and **more versions/topologies tested**. Items are grouped by that priority.

## Data-safety blockers (non-negotiable for beta)

- [x] **Real MariaDB GTID gap detection.** ✅ Landed in v0.16.0 (#518; verified vs live MariaDB 11.4.12).
- [x] **Decode compressed row events** (`log_bin_compress=ON`). ✅ CLOSED via PR #930 ("index MariaDB compressed row events").
- [x] **MariaDB-specific fidelity audit.** ✅ #620 CLOSED via PR #989 — full corruption sweep run against MariaDB.
- [x] **Multi-domain GTID stance.** ✅ RATIFIED option (1) per-domain resume; validated live vs an interleaved two-domain stream and docs caveat closed via PR #1348 (closed #621). The validation test exposed a go-mysql v1.13.0 GTID-accumulation data race — fixed by bumping to v1.16.0 in the same PR. Original: **Tracked in #621.** PARTIAL: gap detection is already domain-aware (#518, `mariadbCheckpointCoversFloor`) and resume is structurally multi-domain (`gomysql.ParseMariadbGTIDSet` → `StartSyncGTID` carries the full per-domain map). Remaining: ratify option (1) per-domain-resume as the stance, validate against a live multi-domain cluster, close the `docs/mariadb.md:165` topology caveat.
- [x] **Position corruption under GTID groups** (found *after* this epic was written): MariaDB defers `end_log_pos` / emits `LogPos=0` intra-transaction, corrupting start_pos/end_pos and poisoning resume dedup. ✅ Fixed via PR #1180 (#1117; #986 was the same bug). Residuals tracked separately (P3).

## Coverage / surface (to claim "supported" honestly)

- [x] **Version matrix in CI** ✅ #622 closed via PR #1339 — 10.6 LTS + 10.11 + 11.4 matrix (fail-fast off), all three contexts now required by branch protection. Original: — **tracked in #622.** Add 10.6 LTS + 10.11 alongside 11.4 (alpha tests only 11.4). Unblocked: its deps #518/#520 are closed; pure YAML matrix work in `ci.yml`.
[...]

PR #518, opened 2026-06-19
feat: real MariaDB GTID gap detection (beta gate #1)

nethalo:
## Gate #1 of the MariaDB-as-source beta (#517)

Replaces the alpha **degrade-to-warn** for MariaDB GTID-mode resume with real purged-binlog gap detection, following the design verified against live MariaDB 11.4 in the #517 kickoff comment. Implemented TDD, against a live MariaDB, with zero change to the MySQL path.

### How MariaDB gap detection works
MariaDB has no `@@gtid_purged`, so the **purge floor** is derived from `BINLOG_GTID_POS(<oldest surviving binlog>, 4)` — the GTID state at the start of the oldest binlog the source still has; everything strictly before it is gone. The executed set comes from `@@gtid_binlog_pos`. New `detectMariaDBGTIDGap` runs three queries (`SHOW BINARY LOGS` → `BINLOG_GTID_POS` → `@@gtid_binlog_pos`) and returns the same `gapResult` the existing flavor-agnostic auto-advance + durable `gap_lost_at` machinery already consumes.

Decision tree: empty/NULL floor → caught-up or fillable; floor **not covered** by the checkpoint → **unfillable** (`PurgedGTIDSet = floor`, feeds auto-advance); else equal → no-gap / behind → fillable.

### Domain-global coverage (not go-mysql's `Contain`)
`mariadbCheckpointCoversFloor` compares the **max sequence per domain and ignores `server_id`**, because a MariaDB sequence is domain-global. This is faithful to the design's "per-domain `seq >= other.seq`" intent and avoids a spurious unfillable-gap alarm after a primary failover — go-mysql's set-level `MariadbGTIDSet.Contain` is keyed by `(domain, server)` and would false-alarm when the floor and checkpoint carry different `server_id`s in the same domain. (That error is in the safe direction — a false alarm, never silent loss — but the per-domain max avoids it.)

### MySQL stays byte-identical
The GTID auto-advance is now flavor-aware (`normalizeGTIDForFlavor` / `parseGTIDSetForFlavor` instead of the hardcoded `NormalizeGTIDSet` / `ParseMysqlGTIDSet`). For a MySQL source those reduce to the exact original calls — a structural no-op, confirmed by the unchanged MySQL gap test suite.

### Bug caught by empirical verification
`BINLOG_GTID_POS` returns SQL **NULL** (not `""`) on a source whose first binlog still exists — the common no-purge case. A plain-string `Scan` would error and make the stream **refuse to start**. Fixed with `sql.NullString` (NULL ≡ nothing purged), pinned by a deterministic unit test.

### Tests
- **Unit (sqlmock):** every branch — caught-up, fillable, unfillable, empty floor, **NULL floor**, multi-domain (fillable + unfillable), cross-server failover, domain-never-seen, empty-checkpoint guard, plus a **multi-server-per-domain** test that mutation-catches a broken `max()` aggregation (the silent-data-loss direction — added after an adversarial review found the original suite left it untested).
- **Integration (live MariaDB 11.4):** `TestDetectMariaDBGTIDGap_livePurge` manufactures a real `PURGE BINARY LOGS` gap, asserts unfillable detection, and **closes the auto-advance loop** — the floor re-parses with the MariaDB parser and a second detection clears the gap (clean re-sync). Runs in the `integration-mariadb-source` CI job.

### Docs
Dropped the alpha gap-detection limitation from `docs/mariadb.md` (incl. the now-unreachable troubleshooting row) and `docs/streaming.md`; reframed the remaining caveat to the genuinely-untested mid-capture primary-failover topology; fixed the MySQL-specific `--gap-timeout` flag help.

Refs #517 (beta epic), #515 (alpha).

### DIFF
--- a/cmd/bintrail/stream.go
+++ b/cmd/bintrail/stream.go
@@ -75,7 +75,7 @@ var (
 func init() {
 	streamCmd.Flags().StringVar(&strmIndexDSN, "index-dsn", "", "DSN for the index MySQL database (required)")
 	streamCmd.Flags().StringVar(&strmSourceDSN, "source-dsn", "", "DSN for the source MySQL server (required)")
-	streamCmd.Flags().StringVar(&strmFlavor, "source-flavor", "mysql", "Source database flavor: mysql or mariadb (MariaDB source support is alpha; prefer position mode for resume)")
+	streamCmd.Flags().StringVar(&strmFlavor, "source-flavor", "mysql", "Source database flavor: mysql or mariadb (MariaDB source support is alpha)")
 	streamCmd.Flags().Uint32Var(&strmServerID, "server-id", 0, "Unique replica server ID (required, must differ from all other servers)")
 	streamCmd.Flags().StringVar(&strmStartFile, "start-file", "", "Initial binlog file (mutually exclusive with --start-gtid)")
 	streamCmd.Flags().Uint32Var(&strmStartPos, "start-pos", 4, "Initial position within start file")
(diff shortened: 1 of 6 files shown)


nethalo:
## Multi-agent review pass applied

Ran a 4-specialist review (code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer) plus an earlier adversarial review workflow. All cleared the data-loss-safety invariant (no silent-loss path; every detection error fails the stream loud). Applied findings:

**Test guards added (both mutation-proven):**
- `TestDetectMariaDBGTIDGap_boundaryAtFloor` — pins the strict `cpMax < floorMax` boundary (cp=floorMax−1 → unfillable, cp=floorMax → fillable). An off-by-one here is the silent-data-loss direction; the prior tests all sat far from the boundary.
- `TestNormalizeGTIDForFlavor` — pins the byte-identical-for-MySQL guarantee of the now-flavor-aware auto-advance.
- `TestMariaDBCheckpointCoversFloor_multiServerPerDomain` (from the adversarial pass) — closes the untested multi-server-within-a-domain `max()` aggregation (silent-loss direction), using a 64-iteration loop to beat Go's randomized map order.

**Pre-existing MEDIUM fixed in this PR (per maintainer decision — affects MySQL too):**
- The shared auto-advance stamped `gap_lost_at` *after* advancing the checkpoint with a warn-only failure, so a transient index-DB error could leave an advanced checkpoint with no durable data-loss record (the #402 "healthy badge over a silent skip" scenario). Extracted `persistGapAutoAdvance`: stamp **before** advancing, **fail loud**. Pinned by `TestPersistGapAutoAdvance_stampsBeforeAdvance` + `_abortsOnStampFailure` (mutation-proven against an advance-first regression).

**Comment/doc accuracy:** 3 wording fixes (stale "recommended position mode" framing, an imprecise "GTID gap detection in position mode" phrase, and a confidence-boundary comment that understated its own test coverage).

All green locally: full unit suite (all packages), `go vet`, and the MariaDB integration selection (`-race -p 1`).

PR #521, opened 2026-06-20
fix: surface S3 footer-probe failures in reconcile --deep

### LINKED ISSUE #469: archive reconcile --deep: silently downgrades to non-deep when S3 footer probes fail
## Problem

When `s3ParquetRowCount` fails for an object (network, credentials — e.g. the duckdb aws-extension setup of #459/#464 failed on a role-only host), the per-object failure is `slog.Warn`'d, but `internal/archive/reconcile.go` then **skips the row-count comparison** for that file entirely: `--deep` quietly delivers non-deep verification and can report "no drift" on exactly the discrepancies it was asked to verify. No prune/repair risk (presence comes from the SDK listing), but the deep contract is silently weakened — and reconcile's dry-run is documented as a cron monitor, so a green run carries weight.

## Proposed fix

Count footer-probe failures and surface "N file(s) could not be deep-verified" in the reconcile summary; exit non-zero in dry-run mode when N > 0 (mirroring the existing drift-monitor contract), or at minimum flag the affected rows in the report.

Found during the PR #464 silent-failure audit.

nethalo:
## Problem (#469)

`archive reconcile --deep` was silently downgrading to non-deep verification when a Parquet-footer probe failed.

When the footer read fails for an object, the scan logs a `slog.Warn` and leaves `RowCount` as the zero-value `sql.NullInt64{}` (`Valid:false`). The deep row_count drift check in `internal/archive/reconcile.go` guards on `rowCount.Valid`:

```go
if opts.Deep && rowCount.Valid && (!row.RowCount.Valid || row.RowCount.Int64 != rowCount.Int64) { ... }
```

so an INVALID RowCount silently **skips** the comparison for that pair. A `--deep` dry-run could therefore report "no drift" — and exit 0 — on exactly the objects it was asked to deep-verify. A green cron run hid unverified objects.

The dry-run exit (`report.Err()`) is computed purely from diff Actions and structurally cannot see a silently-skipped row_count check (it produces no action).

## Fix (decision-layer count — closes local AND dual-backend)

The count is computed where the gate lives — at `archive.Diff`'s pass-1 loop, keyed on the **picked** `rowCount` (`pickMeta`), not on a per-S3-probe counter:

- `Report.DeepUnverified` is incremented when `opts.Deep` is true and the **picked** row_count is Invalid for a scanned pair.
- The command sources `deepUnverified` from `report.DeepUnverified`; `scanS3Archive` drops its footer-failure return value (the `slog.Warn` stays).
- **Surfaced** in both output modes: a `WARNING:` line in text (only when N > 0) and an always-present `deep_unverified` field in `--format json`.
- **Dry-run exits non-zero** when N > 0, via `reconcileDryRunErr`, wrapping the unchanged `report.Err()` (the count can't ride it — a skipped check produces no diff action).

Counting at the decision layer is strictly better than the raw per-S3-probe counter:

- **(a) catches local footer failures** — `scanLocalArchive` leaves `RowCount` Invalid on a footer read error with a `slog.Warn` but no counter; that path was previously a silent skip too.
- **(b) catches the dual-backend prefer-local blind spot** — `archive reconcile` accepts `--archive-dir` AND `--archive-s3` together; `archive.Diff` pairs the local+S3 file for one `(partition, bintrail_id)`, and `pickMeta` PREFERS the local count. A broken local footer + a readable S3 footer made `pickMeta` return the Invalid local count → deep check skipped → `footerFailures==0` → dry-run exit 0. The original PR's S3-only counter re-instantiated the exact silent downgrade it set out to kill.
- **(c) no over-count false-positive** — when a partition is genuinely deep-verified via the surviving backend (picked count Valid), it is NOT counted.

## Scope

The residual gap was **dual-backend (and local), not "S3-only"** — and it is now **closed**. The decision-layer count covers a failed local footer read, the dual-backend prefer-local case, and S3-only, in one place, without false-positiving on a pair verified via the other backend.

**Execute mode** (`--repair`/`--prune`) still *prints* the count but keeps its existing exit code — only the dry-run cron path is gated. This asymmetry is per-spec, unchanged by this fix.

## Tests

`internal/archive/reconcile_test.go` — `TestDiffDeepUnverified`:
- dual-backend prefer-local (local footer failed, S3 valid) → `DeepUnverified==1`, **no action, `Err()==nil`** (the silent-downgrade state)
- local-only footer failure → counted
- S3-only footer failure → counted (parity with the old per-probe counter)
- picked count valid → NOT counted (no false-positive); S3-only with a valid footer → NOT counted
- no `--deep` → count stays zero

`cmd/bintrail/archive_test.go`:
- `TestReconcileWiringFromReportDeepUnverified` — command wiring sources `deepUnverified` from `report.DeepUnverified`; zero-drift + `DeepUnverified>0` → non-zero dry-run + JSON field + text WARNING.
- existing `TestReconcileDryRunErrFlagsDeepUnverified` / `...JSONIncludes...` / `...TextSurfaces...` retained.

### Verification

```
CGO_ENABLED=1 go build ./...                                   # OK
CGO_ENABLED=1 go vet ./cmd/bintrail/... ./internal/archive/... # OK
CGO_ENABLED=1 go test ./internal/archive/... ./cmd/bintrail/... # ok (unit)
CGO_ENABLED=1 go test -tags integration \
  ./internal/archive/... ./cmd/bintrail/...                    # ok (MySQL :13306)
```

Closes #469

### DIFF
--- a/cmd/bintrail/archive.go
+++ b/cmd/bintrail/archive.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/fs"
@@ -111,6 +112,11 @@ func runArchiveReconcile(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 
 	var files []archive.ScannedFile
+	// deepUnverified counts S3 objects whose --deep footer probe failed:
+	// they were scanned but could NOT be deep-verified, so the row_count
+	// drift check silently skips them (#469). Surfaced in the report and,
+	// on the dry-run path, made to fail the cron monitor.
+	deepUnverified := 0
 	if arcDir != "" {
 		local, err := scanLocalArchive(arcDir)
 		if err != nil {
@@ -119,11 +125,12 @@ func runArchiveReconcile(cmd *cobra.Command, args []string) error {
 		files = append(files, local...)
 	}
 	if arcS3 != "" {
-		remote, err := scanS3Archive(ctx, arcS3, arcRegion, arcDeep)
+		remote, footerFailures, err := scanS3Archive(ctx, arcS3, arcRegion, arcDeep)
 		if err != nil {
 			return fmt.Errorf("scan --archive-s3: %w", err)
 		}
 		files = append(files, remote...)
+		deepUnverified += footerFailures
 	}
 
 	db, err := config.Connect(arcIndexDSN)
@@ -147,15 +154,16 @@ func runArchiveReconcile(cmd *cobra.Command, args []string) error {
 
 	executed, execErrs := executeReconcileActions(ctx, db, report.Actions, arcRepair, arcPrune)
 
-	if err := writeReconcileReport(os.Stdout, arcFormat, &report, executed, execErrs, arcRepair, arcPrune); err != nil {
+	if err := writeReconcileReport(os.Stdout, arcFormat, &report, deepUnverified, executed, execErrs, arcRepair, arcPrune); err != nil {
 		return fmt.Errorf("write report: %w", err)
 	}
 	if len(execErrs) > 0 {
 		return fmt.Errorf("%d reconcile action(s) failed", len(execErrs))
 	}
 	if !arcRepair && !arcPrune {
-		// Dry-run: non-zero exit on any drift (the cron monitor contract).
-		return report.Err()
+		// Dry-run: non-zero exit on any drift (the cron monitor contract),
+		// including objects --deep was asked to verify but couldn't (#469).
+		return reconcileDryRunErr(&report, deepUnverified)
 	}
 	// Execute mode: exit 0 ⟺ no unaddressed drift remains. EVERY action
 	// this invocation's flags didn't execute counts — --prune without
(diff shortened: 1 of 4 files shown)


nethalo:
**Review fix applied** (`dcf17d8`) — moved the deep-unverified accounting to the **decision layer**.

The 2-lens review flagged a dual-backend blind spot: the original S3-only `footerFailures` counter re-instantiated the exact silent `--deep` downgrade this PR exists to kill. `archive reconcile` accepts `--archive-dir` AND `--archive-s3` together; `archive.Diff` pairs a local+S3 file into one `(partition, bintrail_id)`, and `pickMeta` **prefers local**. So a failed *local* footer read (Invalid `RowCount` + `slog.Warn`, no counter) with a *successful* S3 probe → `pickMeta` returns the Invalid local count → the deep `row_count` check skips → `footerFailures==0` → dry-run exits 0. Silent downgrade, back again.

**Chose the decision-layer approach** (over the local-counter fallback): count "deep requested but the PICKED `rowCount` is Invalid" in `archive.Diff`'s pass-1 loop → `Report.DeepUnverified`. One place closes all three:
- local footer failures (was a silent skip too),
- the dual-backend prefer-local case,
- and it removes the over-count false-positive — a pair genuinely deep-verified via the *surviving* backend (picked count Valid) is **not** counted.

`scanS3Archive` drops its footer-failure return (the `slog.Warn` stays); the command sources `deepUnverified` from `report.DeepUnverified`. `report.Err()` untouched — a silently-skipped check produces no diff action, so the count rides a distinct struct field that still feeds the text WARNING, the JSON `deep_unverified`, and the non-zero dry-run exit.

New test `TestDiffDeepUnverified` pins the discriminating case: dual-backend prefer-local → `DeepUnverified==1` while `len(Actions)==0 && Err()==nil` (the silent state) — plus local-only, S3-only, the picked-valid no-false-positive, and no-`--deep` cases. `TestReconcileWiringFromReportDeepUnverified` pins the command wiring.

Green: build + vet + unit + integration (MySQL :13306) on `internal/archive/...` and `cmd/bintrail/...`. PR scope note reworded — the residual was dual-backend, not "local-only", and is now closed.

nethalo:
Applied the IMPORTANT review fix (`783a32d`): the execute-mode exit (`--repair`/`--prune`) was omitting `deepUnverified`, so `archive reconcile --deep --repair` printed the "could not be deep-verified" WARNING but returned nil (silent exit 0) — reintroducing the #469 green-run-hides-unverifiable-files failure in execute mode (and `--repair` can't fix a footer it can't read, so it's genuinely unaddressed drift).

- **FIX 1 (IMPORTANT):** extracted a sibling `reconcileExecuteErr(rep, deepUnverified, repair, prune)` holding the pending-sum drift rule + the shared `deepUnverified` guard, and routed the execute branch through it — so the deep-unverified signal can't be dropped on one path. Dry-run and execute remain distinct siblings (dry-run fails on ALL drift via `rep.Err()`; execute only on what its flags left unaddressed). The WARNING line already prints in both modes. Added `TestReconcileExecuteErrFlagsDeepUnverified` mirroring the dry-run test (`deepUnverified>0` → non-zero; `==0` → clean; pending drift still wins).
- **FIX 2 (suggestion):** two over-count-guard subtests on `TestDiffDeepUnverified` — local-VALID + s3-no-count → `DeepUnverified==0` (pins pickMeta-prefers-local); both-no-count → `DeepUnverified==1` (pins counted-once-per-key, not per-backend).
- **FIX 3 (suggestion):** one-line comment at the `insertAction` branch noting a footer-failed INSERT under `--deep` surfaces via `report.Inserts` (drift → non-zero exit), not `DeepUnverified` — placement is intentional.

Verify (CGO_ENABLED=1): `go build ./...` + `go vet ./cmd/bintrail/... ./internal/archive/...` clean. Unit `go test ./internal/archive/... ./cmd/bintrail/...` green; integration (`-tags integration`, MySQL :13306 up) green too.

Issue #503, opened 2026-06-14
baseline: harden SQL→Parquet path against non-default binary encodings + silent NULL-on-error


nethalo:
Follow-ups surfaced by the multi-agent review of #501 (binary/JSON parser fix). All are **pre-existing** and **not regressions** from #499/#501; #501 fixed the default mydumper path (`_binary "…"`, `CONVERT(…)`), these are the remaining edges. Medium priority.

> **Grooming note (2026-06-25):** item 3 below is **already fixed** — struck through. Open items: **1, 2, 4.** Item 1 is a P1 silent-binary-corruption bug and is the real core of this issue.

## 1. Hex-blob binary (`0x…` / `_binary 0x…`) stored as ASCII text — silent corruption

mydumper `--hex-blob` (and mysqldump `--hex-blob`) dump binary columns as `0x<hex>` instead of `_binary "…"`. The parser's pre-existing `0x` branch (`reader_sql.go`) returns the literal string `"0x612C6229"`; for a binary column `convertValue` (`writer.go:200`) then stores **10 ASCII bytes instead of the 4 bytes `a,b)`**. The introducer form `_binary 0x612C6229` is worse: `parseIntroducedString` finds no quote → falls back → captures the whole `"_binary 0x612C6229"` token.

Repro: `INSERT INTO t VALUES (1, 0x612C6229);` then read the `vb` column back from the baseline Parquet → ASCII `0x612C6229`, not bytes.

Fix is non-trivial because the parser is type-blind: hex-decoding belongs in `convertValue` (which knows the column is binary), but a binary column whose actual bytes are the ASCII text `0x..` would be mis-decoded. Needs a deliberate design (e.g. decode `0x…` only for binary-family columns, document the residual ambiguity).

## 2. GEOMETRY columns: no Parquet type mapping + unverified dump encoding

`MysqlToParquetNode` (`schema.go`) has no `geometry` case → it falls to the STRING node, so WKB (non-UTF8) bytes would land in a String-typed column. Also unverified whether mydumper emits geometry as `_binary "…"` (would route through #501's introducer path) or some other form. Needs a geometry column in a fixture to confirm end-to-end. (#501 removed the unverified GEOMETRY claim from its code comment for now.)

## ~~3. `convertValue` silently stores NULL on conversion error~~ — ✅ FIXED

[...]

PR #522, opened 2026-06-20
fix: baseline UNSIGNED integers silently lost to NULL

### LINKED ISSUE #506: Baseline write-path numeric fidelity: BIGINT UNSIGNED > 2^63 loss + FLOAT(32) render divergence

nethalo:
## Problem

The mydumper-schema parser (`internal/baseline/schema.go`) captured only the bare type token (`\w+`), **dropping the `UNSIGNED` attribute for every integer width**. Consequently:

- `BIGINT UNSIGNED` in `[2^63, 2^64-1]` failed `ParseInt(raw, 10, 64)`
- `INT/INTEGER UNSIGNED` > 2147483647 failed `ParseInt(raw, 10, 32)`

Both errors hit the **silent NULL-on-error fallback** in `WriteRow` (the bare `store NULL` whose comment claimed "to avoid data loss" while *causing* it). The value was dropped from the baseline snapshot with no error — silent data loss in a data-recovery tool.

## Fix (two coupled parts)

**#506 — UNSIGNED:**
- `colRe` now captures the trailing `unsigned` attribute, matched **only in the type tail** (after the token + optional display width like `int(10)`), so a column literally named `is_unsigned` or a `COMMENT 'unsigned…'` never trips it (the name is isolated in backticks, group 1).
- `Column` gains an `Unsigned bool`. `INT/INTEGER UNSIGNED` → Parquet `INT64`; `BIGINT UNSIGNED` → Parquet `UINT64`. `convertValue` uses `ParseUint` so the full unsigned range round-trips (`BIGINT UNSIGNED` stored by bit pattern: `uint64(v.Int64())` reads it back).
- The **exported `MysqlToParquetNode(string)` signature is unchanged** (signed) — `internal/archive` and `internal/byos` pass bare tokens and keep working; the unsigned-aware variant is internal and threaded through `ParseSchema`.

**#503 item-3 — the silencer:**
- `WriteRow` no longer turns a `convertValue` error into a silent NULL. It **fails loud**, naming the column and raw value, so the run aborts rather than publishing a lossy snapshot — consistent with the project's fail-loud convention (compose empty-dump FATAL, `--allow-gaps` default-abort).

### Behavioral note
Fail-loud means a baseline that previously *silently* NULLed an unconvertible value (e.g. a `0000-00-00` zero-date or a malformed numeric) will now **abort the run** with a located error. That is the intended fix (fail-loud > lossy snapshot), called out here so it isn't later filed as a regression.

## Tests (table-driven, `internal/baseline` unit)
- `TestParseSchemaUnsigned` — pins `Unsigned` per column incl. `int(10) unsigned`, `bigint unsigned zerofill`, an `is_unsigned tinyint` column, and an `unsigned`-in-COMMENT column (both must be `Unsigned=false`).
- `TestConvertValueUnsigned` — `INT UNSIGNED` = 4294967295 and `BIGINT UNSIGNED` = 18446744073709551615 convert without overflow.
- `TestWriteRowUnsignedRoundTrip` — full `WriteRow → Parquet → read-back` for both maxima; neither becomes NULL.
- `TestWriteRowFailsLoudOnUnconvertible` — a bad value through **`WriteRow`** (where the silencer lived) now returns a non-nil error mentioning the column and value.

Verified green: `go build ./...`, `go vet ./internal/baseline/...`, `go test ./internal/baseline/... ./internal/archive/... ./internal/byos/... ./internal/reconstruct/... ./internal/parquetquery/...` (unit), and `go test -tags integration ./internal/baseline/...` (Docker MySQL on :13306).

Closes #506
Addresses #503 item-3 (silent NULL-on-error); items 1 (hex-blob), 2 (GEOMETRY), 4 (column-count guard) remain as follow-ups.

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -17,6 +17,14 @@ import (
 // BinlogEventColumns defines the 15 non-generated binlog_events columns in
 // MySQL table order (pk_hash is a stored generated column and is omitted).
 // Exported for reuse by the buffer package when writing Parquet files.
+//
+// event_id / start_pos / end_pos are BIGINT UNSIGNED in MySQL but are scanned as
+// sql.NullInt64 here (ArchivePartition) and never exceed int64 in practice —
+// AUTO_INCREMENT and binlog byte offsets stay well under 2^63 — so the signed
+// Int64 column they map to via MysqlToParquetNode is lossless. connection_id is
+// the one column that genuinely needs the unsigned widening below: it is INT
+// UNSIGNED (a CONNECTION_ID() can exceed int32's 2147483647) and is written via
+// FormatUint by the buffer path, so a signed Int(32) column would reject it.
 var BinlogEventColumns = []baseline.Column{
 	{Name: "event_id", MySQLType: "bigint", ParquetType: baseline.MysqlToParquetNode("bigint")},
 	{Name: "binlog_file", MySQLType: "varchar", ParquetType: baseline.MysqlToParquetNode("varchar")},
@@ -24,7 +32,7 @@ var BinlogEventColumns = []baseline.Column{
 	{Name: "end_pos", MySQLType: "bigint", ParquetType: baseline.MysqlToParquetNode("bigint")},
 	{Name: "event_timestamp", MySQLType: "datetime", ParquetType: baseline.MysqlToParquetNode("datetime")},
 	{Name: "gtid", MySQLType: "varchar", ParquetType: baseline.MysqlToParquetNode("varchar")},
-	{Name: "connection_id", MySQLType: "int", ParquetType: baseline.MysqlToParquetNode("int")},
+	{Name: "connection_id", MySQLType: "int", Unsigned: true, ParquetType: baseline.MysqlToParquetNode2("int", true)},
 	{Name: "schema_name", MySQLType: "varchar", ParquetType: baseline.MysqlToParquetNode("varchar")},
 	{Name: "table_name", MySQLType: "varchar", ParquetType: baseline.MysqlToParquetNode("varchar")},
 	{Name: "event_type", MySQLType: "tinyint", ParquetType: baseline.MysqlToParquetNode("tinyint")},
--- a/internal/archive/archive_test.go
+++ b/internal/archive/archive_test.go
@@ -1,10 +1,14 @@
 package archive
 
 import (
+	"context"
+	"database/sql"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	_ "github.com/duckdb/duckdb-go/v2" // DuckDB driver — exercises the real read path
 	"github.com/parquet-go/parquet-go"
 
 	"github.com/dbtrail/dbtrail/internal/baseline"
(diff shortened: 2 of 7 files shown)


nethalo:
## Adversarial-review fixes applied (`d9aee97`)

**BLOCKER — zero-date carve-out.** The blanket fail-loud in `WriteRow` aborted the entire baseline run on a *legal* MySQL zero-date (`0000-00-00` / `0000-00-00 00:00:00` / `.000000` variant). mydumper dumps these as a literal string, so they reach `convertValue` with `isNull=false`; the Go time parser rejects them ("month out of range") → `processTable` aborts and the deferred cleanup deletes the partial Parquet. Legacy MySQL tables commonly carry `DEFAULT '0000-00-00 00:00:00'`, so this killed real baselines.

Fix: `convertValue` classifies the all-zero date family (`isZeroDate` prefix check, DATE/DATETIME/TIMESTAMP only — `TIME`'s `00:00:00` is legal midnight and is **excluded**) and returns a typed `errZeroDate` sentinel; `WriteRow` maps that sentinel to a **deliberate NULL + a once-per-column `slog.Warn`** (a `zeroDateWarned` set avoids a per-row log flood on a `DEFAULT '0000-00-00'` table). The zero date *is* MySQL's pseudo-NULL, so NULLing it is faithful — and the warn keeps it visible, not a silent drop. **Genuinely unrepresentable values (UNSIGNED overflow, garbage) still abort** — the carve-out is surgical. *Scope note:* partial-zero dates (`2020-00-00`, `2020-05-00` — real year, zero month/day) remain **fail-loud by design** — only the all-zero forms are MySQL's documented pseudo-NULL sentinel; partial-zero has no clean NULL semantics.

**MEDIUM — test the real read path.** `TestWriteRowUnsignedRoundTrip` only asserted on parquet-go's manual `uint64(row[0].Int64())` reinterpret, but every production consumer reads baselines through DuckDB `parquet_scan` (`reconstruct.ReadBaselineRow`, `fulltable.go`, shim `_snapshot`). Added `TestWriteRowUnsignedDuckDBScan`: reads the written Parquet via `parquet_scan`, asserts **BIGINT UNSIGNED → Go `uint64` = 18446744073709551615** and **INT UNSIGNED → `int64` = 4294967295** (per `schema.go`'s widening — UINT64 vs signed INT64), and round-trips each through `recovery.FormatSQLValue` to the exact literal. Mutation-checked: reverting the BIGINT-UNSIGNED widening makes it scan back as a negative `int64` and the test fails.

**NIT — case-insensitive unsigned.** `colRe`'s `(unsigned)?` group is now `((?i:unsigned))?` so an uppercase `UNSIGNED` populates group 3 (the dead `EqualFold` guard now has something to match). The capture group is preserved on purpose — a bare `(?i:...)` would be non-capturing and panic `m[3]`. Hardening only; mydumper emits lowercase by contract.

**Tests added:** zero-date baseline COMPLETES (→ NULL, warn fired once per column), non-sentinel garbage date still ABORTS (names column+value), uppercase `UNSIGNED` parses, plus the DuckDB-scan round-trip.

**Verified:** `go build ./...`, `go vet ./internal/baseline/...`, unit + integration (Docker MySQL :13306) all green; recovery/reconstruct/archive/byos suites green. Each new fix mutation-checked (its test fails when the fix is reverted). No version/CHANGELOG bump.

nethalo:
## Review fixes applied (CRITICAL + IMPORTANT + the two cheap SUGGESTIONs)

Pushed `7ee6ace`. Build + vet + unit + integration (`:13306` up) all green for `internal/{baseline,archive,buffer,byos}`.

**CRITICAL — fail-loud regression the PR introduced.** The new fail-loud `WriteRow` is shared by four callers. `archive.BinlogEventColumns` declared `connection_id` as a signed `Int(32)` (via the signed-only `MysqlToParquetNode`), but `connection_id` is **INT UNSIGNED** (`CONNECTION_ID()` exceeds 2147483647 on busy long-running servers) and is written via `FormatUint` on the buffer path. A value past int32 hit `convertValue`'s signed `ParseInt(_,10,32)` → post-PR the fail-loud **aborted the whole partition archive / BYOS flush** (pre-PR it silently NULLed and succeeded). The widening added in this PR was unreachable because the unsigned-aware builder was unexported.

Fix:
- Exported `baseline.MysqlToParquetNode2(typeToken, unsigned)` delegating to the existing internal builder; kept signed `MysqlToParquetNode` for back-compat.
- `connection_id` → `{Unsigned:true, MysqlToParquetNode2("int", true)}` → Int64, matching `ParseSchema`.
- Read-path verified safe: `parquetquery` already scans `connection_id` as `sql.NullInt64`, and `union_by_name=true` unifies **old Int32 + new Int64** files (the normal upgrade path) cleanly — proven empirically with a throwaway DuckDB scan and confirmed by the archive integration test against real MySQL.

**IMPORTANT — test gap.** Added `TestParseSchemaWiringUnsignedDuckDBScan`: a `bigint unsigned`/`int unsigned` schema → `ParseSchema` (no hand-built `ParquetType`) → `NewWriter` → `WriteRow` maxima → DuckDB `parquet_scan`, asserting `c_big_u→uint64==18446744073709551615` and `c_int_u→int64==4294967295`. Verified it **fails under MUT-D** (`mysqlToParquetNode(typeToken, false)` at the `ParseSchema` call site) — the existing unsigned tests stay green under that bug because they hand-build the ParquetType.

**FIX 1d round-trips:** `TestWriteRowConnectionIDUnsignedDuckDBScan` (archive) + `TestWriteParquet_connectionIDUnsignedDuckDBScan` (buffer) drive the production `BinlogEventColumns` with `connection_id` = 4294967295 and 3000000000 through `parquet_scan` — round-trip, not abort, not NULL.

**SUGGESTION 3 — partial-zero-date.** `TestWriteRowPartialZeroDateAborts`: `'2020-00-00'` and `'2020-05-00 00:00:00'` (date + datetime + timestamp) must STILL fail loud naming the column — only the all-zero `0000-00-00` prefix is carved out.

**SUGGESTION 4 — comments.** Reworded both stale "UNSIGNED overflow" abort examples in `writer.go` (the `errZeroDate` doc + the `err != nil` branch) to "non-numeric garbage in an int column / out-of-range literal"; noted in-range unsigned no longer reaches that branch (out-of-range unsigned literals still abort).

---

### Other callers the export touched

- **`schema_version` (archive/buffer/byos): considered, deliberately NOT changed.** It is also a `uint32` written via `FormatUint` into a signed `int` column — structurally the same shape — BUT (a) the read path scans it as `sql.NullInt32` (vs `connection_id`'s `sql.NullInt64`), so widening its write would force an uncoordinated `NullInt32→NullInt64` read-path change + column promotion across old files, and (b) it is a schema-snapshot counter (increments once per DDL re-snapshot) that cannot approach 2³¹ — unlike a per-connection id. The codebase already encodes this distinction; respecting it is correct.
- **`event_id`/`start_pos`/`end_pos`:** BIGINT UNSIGNED but scanned as `sql.NullInt64`, bounded under 2⁶³ in practice (AUTO_INCREMENT / binlog byte offsets) — comment added, no change.
- **`internal/byos` needs no code change:** its only numeric in `PayloadColumns` is `schema_version` (above); `event_type` is varchar.

No version/CHANGELOG bump.

PR #523, opened 2026-06-20
feat: fail loud on partial binlog row images

### LINKED ISSUE #493: Harden binlog_row_image=FULL enforcement: detect partial row images and fail loud
## Summary

The `binlog_row_image=FULL` requirement is enforced only by a one-shot check of the **server-global** variable at startup. But the variable is **session-settable**, and the check is **skipped entirely** by `bintrail index` without `--source-dsn`. So a deployment that genuinely complies with the contract (global = `FULL`, validation passes) can still ingest `MINIMAL`/`NOBLOB` events from individual sessions and store them as if complete — silently corrupting before/after images that `recover` later trusts.

This issue is **not** about supporting MINIMAL. It's about making the existing requirement enforceable and honest: **detect partial row images and fail loud** instead of storing NULLs.

## Evidence

- `metadata.ValidateBinlogRowImage` checks `SHOW VARIABLES LIKE 'binlog_row_image'` once — the global, at startup.
- `cmd/bintrail/index.go:100-102`: without `--source-dsn`, validation is skipped with only a `slog.Warn`.
- `RowsEvent.SkippedColumns` is never consulted (appears only in comments); go-mysql pads absent columns to full width with `nil`, so the column-count guard at `internal/parser/parser.go:247` passes.
- Reproduced: under `SET SESSION binlog_row_image='MINIMAL'`, an UPDATE is indexed with unchanged columns and the after-image PK as `nil` (= NULL).

## Goal / suggested fix

Consistent with the project's fail-loud guards (compose `baseline` empty-dump FATAL; `--allow-gaps` defaulting to abort rather than return partial results):

1. Detect partial images per row event using the row-event column bitmaps / `SkippedColumns`, and **abort the stream/index with a clear error** when a non-FULL image is seen (do not store it). A recovery tool returning silently-wrong data is worse than one that refuses.
2. Optionally, make `bintrail index` require `--source-dsn` (or an explicit opt-out flag) so the FULL/format validation always runs.

## References

- 2026-06-14 binlog-fidelity audit (enforcement residual of the FULL requirement). Companion to the docs issue that states the requirement explicitly.
- Readyset "Replication Internals" Part 8 (row-event column bitmaps; absent ≠ NULL).

nethalo:
## Problem

`binlog_row_image=FULL` enforcement was a one-shot, server-global
`SHOW VARIABLES LIKE 'binlog_row_image'` (`metadata.ValidateBinlogRowImage`) —
session-settable, thus **bypassable**. When bypassed (e.g. a client doing
`SET SESSION binlog_row_image=MINIMAL`), the resulting UPDATE/DELETE events carry
**partial row images**. go-mysql pads the absent columns to `nil`, and bintrail
stored those as genuine `NULL`s — silently corrupting the before/after images
that `recover` later trusts. Plus `bintrail index` skipped source validation
**entirely and silently** when `--source-dsn` was omitted.

This is the last open item of the binlog-fidelity audit (#493).

## Fix

**Core — per-row partial-image detector** in the shared `handleRows` chokepoint
in `internal/parser/parser.go`, so it covers **both** the file-index
(`ParseFile`) and `StreamParser` paths in one place. When any row image omitted
columns, `handleRows` returns an error instead of storing NULL-filled rows. Both
call sites `return handleRows(...)` directly, and both outer loops treat a
non-nil result as fatal (file path → `parseErrCh` abort; stream path →
`StreamParser.Run` returns the error), so the abort is real on both.

**Detection API:** `RowsEvent.SkippedColumns` (go-mysql v1.13.0). go-mysql
populates it in lock-step with `Rows` — one entry per decoded image (UPDATE has
two: before + after), each listing the column ordinals whose present-bit was
clear in the event's column bitmap. Under `FULL` every entry is an empty
(non-nil) slice; under MINIMAL/NOBLOB the absent columns appear there while
go-mysql pads them to `nil`. The detector keys on `len(image) > 0`.

**Empirically verified** against go-mysql v1.13.0 / MySQL 8.x with a throwaway
probe before implementing:

| image  | WRITE      | UPDATE              | DELETE       |
|--------|------------|---------------------|--------------|
| FULL   | `[[]]`     | `[[] []]`           | `[[]]`       |
| MINIMAL| `[[]]`     | `[[1 2 3] [0 2]]`   | `[[1 2 3]]`  |

The probe table included a `VIRTUAL` generated column: `columnCount=4` includes
it, yet a FULL image still yields **zero skips** — so the detector does not
false-positive on virtual columns. (Residual boundary, noted for future readers:
the only way it could over-fire on a healthy table is a hypothetical MySQL
version where TABLE_MAP includes a virtual column but the row image clears its
bit; even then the failure mode is a loud, clearly-messaged abort, not silent
corruption.) Note INSERT after-images are full even under MINIMAL, so the
trigger is UPDATE/DELETE.

**CLI — kill the silent skip.** `bintrail index` now requires `--source-dsn`, or
an explicit `--skip-source-validation` opt-out for indexing offline binlogs
against an already-captured snapshot. The per-row guard still applies under the
opt-out, so offline indexing stays protected. `stream` is unchanged (already
requires `--source-dsn`).

## Tests

- **Unit** (`internal/parser/parser_test.go`): table-driven `TestFirstPartialImage`
  over the exact go-mysql `SkippedColumns` shapes; `TestHandleRows_partialImageFailsLoud`
  (fabricated partial UPDATE → error, no event emitted);
  `TestHandleRows_fullImagePasses` (FULL → emits, no false positive).
- **Unit** (`cmd/bintrail/index_test.go`): `TestRunIndex_requiresSourceOrSkip`
  (no `--source-dsn` and no opt-out → actionable error; opt-out → past the guard).
- **Integration** (`internal/parser/partial_image_integration_test.go`,
  `//go:build integration`): sets `binlog_row_image=MINIMAL` at session level on a
  **pinned connection**, generates a partial UPDATE, copies the binlog, and asserts
  `ParseFile` fails loud and emits no NULL-filled event.

### Evidence (ran against Docker MySQL on :13306)

```
go build ./...                                                 ok
go vet ./internal/parser/... ./internal/metadata/... ./cmd/bintrail/...  ok
go test ./internal/parser/... ./internal/metadata/...         ok
go test -tags integration ./internal/parser/                  ok  (full suite — no FULL-image false positives)
go test -tags integration ./internal/parser/ -run TestParseFile_partialRowImageFailsLoud  PASS
```

Closes #493

### DIFF
--- a/cmd/bintrail/index.go
+++ b/cmd/bintrail/index.go
@@ -35,15 +35,16 @@ Files already marked 'completed' in index_state are skipped.`,
 }
 
 var (
-	idxIndexDSN  string
-	idxSourceDSN string
-	idxBinlogDir string
-	idxFiles     string
-	idxAll       bool
-	idxBatchSize int
-	idxSchemas   string
-	idxTables    string
-	idxFormat    string
+	idxIndexDSN        string
+	idxSourceDSN       string
+	idxBinlogDir       string
+	idxFiles           string
+	idxAll             bool
+	idxBatchSize       int
+	idxSchemas         string
+	idxTables          string
+	idxFormat          string
+	idxSkipSourceCheck bool
 )
 
 func init() {
@@ -56,6 +57,8 @@ func init() {
 	indexCmd.Flags().StringVar(&idxSchemas, "schemas", "", "Only index events from these schemas (comma-separated)")
 	indexCmd.Flags().StringVar(&idxTables, "tables", "", "Only index these tables (comma-separated, e.g. mydb.orders,mydb.items)")
 	indexCmd.Flags().StringVar(&idxFormat, "format", "text", "Output format: text or json")
+	indexCmd.Flags().BoolVar(&idxSkipSourceCheck, "skip-source-validation", false,
+		"Index offline binlogs without --source-dsn, skipping the binlog_format/binlog_row_image/FK-cascade pre-flight (the per-row partial-image guard still applies)")
 	_ = indexCmd.MarkFlagRequired("index-dsn")
 	_ = indexCmd.MarkFlagRequired("binlog-dir")
 	bindCommandEnv(indexCmd)
@@ -69,6 +72,15 @@ func runIndex(cmd *cobra.Command, args []string) error {
 	}
 	if !idxAll && idxFiles == "" {
 		return fmt.Errorf("either --files or --all must be specified")
+	}
+	// Without --source-dsn the binlog_format/binlog_row_image/FK-cascade
+	// pre-flight cannot run; silently skipping it (the old behavior) let a
+	// non-FULL source be indexed with NULL-filled partial images. Require an
+	// explicit opt-out so offline binlog indexing stays possible while the skip
+	// is never accidental (#493). The per-row partial-image guard in
+	// internal/parser still fires even under the opt-out.
+	if idxSourceDSN == "" && !idxSkipSourceCheck {
+		return fmt.Errorf("--source-dsn is required for source validation; pass --skip-source-validation to index offline binlogs without it")
 	}
 
 	ctx := cmd.Context()
@@ -98,7 +110,7 @@ func runIndex(cmd *cobra.Command, args []string) error {
 		}
 		fmt.Println("Source: no FK cascades \u2713")
 	} else {
-		slog.Warn("--source-dsn not provided; skipping source server validation")
+		slog.Warn("--skip-source-validation set: skipping binlog_format/binlog_row_image/FK-cascade pre-flight — the per-row partial-image guard still applies")
 	}
 
 	// ── 2. Index database connection ──────────────────────────────────────────
--- a/cmd/bintrail/index_test.go
+++ b/cmd/bintrail/index_test.go
@@ -98,10 +98,14 @@ func TestRunIndex_allSetPassesFirstGuard(t *testing.T) {
 // first guard even without --all.
 func TestRunIndex_filesSetPassesFirstGuard(t *testing.T) {
 	savedFiles, savedAll := idxFiles, idxAll
-	t.Cleanup(func() { idxFiles = savedFiles; idxAll = savedAll })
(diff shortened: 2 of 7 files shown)


nethalo:
Pushed test/doc hardening (b91b234) — code unchanged, this is review-fix only.

**FIX 1 (IMPORTANT — was untested):** added `TestParseFile_fullImageWithVirtualColumnNotPartial`, an integration test that drives a real go-mysql decode of a real binlog with a `total INT AS (qty*2) VIRTUAL` column and asserts a FULL-image UPDATE still emits its event **without** tripping the partial-image guard. This converts the one-off probe into a CI-enforced regression test — the unit tier (`TestHandleRows_fullImagePasses`) could only assert this against a circular hand-built `[][]int` slice. **Ran and passed** against the live test MySQL.

**FIX 2:** one-line comment that NOBLOB yields the identical `SkippedColumns` shape and is covered by the same code path as the existing MINIMAL integration test (no separate parametrization needed).

**FIX 3:** `docs/ddl-tracking.md` "Without `--source-dsn`: Record Only" now notes the decommissioned-server flow requires `bintrail index --skip-source-validation ...` since #493 made the source pre-flight mandatory (the silent no-source skip was removed).

**FIX 4:** `TestRunIndex_filesSetPassesFirstGuard` sets `idxSkipSourceCheck=true` (save/restore in `t.Cleanup`) so it again reaches `config.Connect` — the first guard its comment claims — instead of stopping at the new `--source-dsn` requirement.

**FIX 5:** comment in `firstPartialImage` clarifying the reported indices are go-mysql's 0-based column ordinals, not bintrail's 1-based `ColumnMeta.OrdinalPosition`.

**Verification (CGO_ENABLED=1):** `go build ./...`, `go vet` (plain + `-tags integration`), and unit tests for `internal/parser/...` + `cmd/bintrail/...` all green. Integration tier ran against live MySQL — both partial-image tests (incl. the new VIRTUAL-column case) and the full `internal/parser` integration package pass.

Issue #461, opened 2026-06-11
baseline: converting a dump with zero tables succeeds silently (Stats{}, nil)


nethalo:
## Problem

`baseline.Run` returns `Stats{}, nil` when the input mydumper directory contains no table files (`internal/baseline/baseline.go:72-74`), and `bintrail baseline` exits 0 printing `tables: 0`. Verified live with the published image: a metadata-only dump converts "successfully" without even creating the output dir.

This matters because an empty/partial dump is easy to produce **with mydumper itself exiting 0** (a `--regex` that matches nothing; a dump user lacking `SELECT` on some schemas — both verified). The result is a missing or incomplete baseline that surfaces weeks later as `ErrNoBaseline` — or worse, as `findBaselineLocal` silently falling back to an **older** snapshot for tables absent from the newest one, i.e. Time-travel reconstructing from ever-staler data with no signal.

The compose `baseline` profile (PR #458) now guards this at the dump step, but the Go-level hole remains for everyone calling `bintrail dump` + `bintrail baseline` directly.

## Fix

- `bintrail baseline` should treat zero discovered tables as an **error** (the input dir had a metadata file but no table data → almost certainly a filter/privilege mistake), or at minimum exit non-zero with a loud warning.
- Consider a `slog.Warn` in `findBaselineLocal` when the newest snapshot lacks a requested table and an older one is used (the staleness fallback is currently invisible).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #524, opened 2026-06-20
fix: mark incomplete baselines + surface S3 staleness

### LINKED ISSUE #466: reconstruct: S3 baseline lookups still fall back to older snapshots silently; staleness invisible to console/shim clients
## Problem

PR #463 added the staleness warn to `findBaselineLocal` (#461), but the S3 path **structurally cannot** do the same: `findBaselineS3`'s glob is table-scoped (`prefix/*/<schema>/<table>.parquet`), so snapshots that *lack* the table are invisible — there is no "newest eligible snapshot" to compare against without a broader listing (`listBaselinesS3` already does a `prefix/*/*/*.parquet` glob that could feed it, at a listing-cost tradeoff on large buckets). `--baseline-s3` is the production mode for console reconstruct and the shim's `_snapshot`, so exactly those users keep the silent ever-staler reconstruction #461 set out to kill. A NOTE(#461) comment marking the asymmetry ships in PR #463.

Second half: even where the warn fires (local), it only reaches the **server log**. The console user gets `baseline_time` in the DTO but no explicit stale flag; the shim's MySQL client gets no in-band signal at all.

## Proposed fix

1. `findBaselineS3`: derive the newest eligible snapshot (reuse the broader glob or an SDK listing) and emit the same warn — measure/cap the listing cost.
2. Console: a `stale_baseline` warning entry in the reconstruct response (`warnings` already exists in the DTO) so Time-travel renders it.
3. Shim: at minimum keep the server-side log (shared once 1 lands); in-band signal needs design.

## Acceptance

Reconstructing a table absent from the newest S3 snapshot warns server-side and surfaces in the console reconstruct response.

### LINKED ISSUE #467: baseline: a partially-failed run publishes converted tables as a normal-looking snapshot (no completeness marker)
## Problem

When some tables fail mid-run, `baseline.Run` exits non-zero (loud) — but the tables that converted **before** the failure remain on disk under `<output>/<timestamp>/...`, producing a partial snapshot indistinguishable from a complete one. `reconstruct.FindBaseline` will treat it as the newest snapshot; for the missing tables it silently falls back to an older snapshot (#461's local warn fires — local only, see the S3 sibling issue) or returns `ErrNoBaseline`. Same shape one threshold over from #461: per-table cleanup is excellent (`processTable` removes partial Parquet files), but there is no snapshot-level completeness marker.

(PR #463 already adds a `ctx.Err()` guard so a *cancelled* run can no longer exit 0 with partial output, and aggregates the multi-table failure count in the returned error.)

## Proposed fix

Write the snapshot into a temp dir and rename `<timestamp>` into place only when every table converted (atomic publish), or write a `_SUCCESS`/manifest marker that `FindBaseline`/`DiscoverBaselines`/`ListBaselines` require (needs a compat story for pre-marker snapshots).

## Acceptance

A run that fails on any table leaves nothing the discovery functions treat as the newest snapshot.

nethalo:
Fixes two coupled baseline issues that share `internal/baseline` + `internal/reconstruct`.

## #467 — partial baseline runs publish a complete-looking snapshot

When a run fails on any table, the tables converted **before** the failure stay on disk under `<output>/<timestamp>/…`, byte-indistinguishable from a complete snapshot. Discovery then treats it as the newest baseline and silently reconstructs missing tables from an older snapshot (#461) or hits `ErrNoBaseline`.

**Fix:** snapshot-level completeness markers (`internal/baseline/marker.go`):
- `_SUCCESS` written only when **every** table converted; `_INCOMPLETE` on a partial/cancelled run.
- A **marker, not atomic rename** — the `--retry` resume path reuses already-converted Parquet under `<output>/<tsDir>`, which a rename-away would destroy.
- Discovery (`DiscoverBaselines`, `findBaselineLocal`, `findBaselineS3`, `listBaselinesLocal/S3`) skips an `_INCOMPLETE`-without-`_SUCCESS` snapshot. **Backward compat:** marker-absent (pre-marker) snapshots are complete-by-default, so existing baselines keep working — only an explicit `_INCOMPLETE` is treated as incomplete.
- **S3 marker detection** uses a `prefix/*/_*` glob — DuckDB's `glob()` does **not** brace-expand `{_SUCCESS,_INCOMPLETE}` (verified empirically; a brace glob silently matches nothing and would have defeated #467 on S3).
- **Upload ordering:** `_SUCCESS` is uploaded **last**. `WalkDir` is lexical and `_SUCCESS` (`_`) sorts before sibling schema dirs (`mydb`), so a mid-walk upload death would otherwise publish the complete-marker over partial data — the same #467 bug on the upload path.

## #466 — S3 baseline lookups fall back to older snapshots silently

`findBaselineS3`'s glob was table-scoped, so snapshots lacking the table were invisible — there was no "newest eligible snapshot" to compare against, and `--baseline-s3` is the production mode for console reconstruct and the shim `_snapshot`.

**Fix:**
1. `findBaselineS3` runs **one** broader newest-snapshot glob (`prefix/*/*/*.parquet`, the same `listBaselinesS3` uses — bounded cost) to derive the newest complete snapshot and emits the **same** `slog.Warn` the local path does. The stale `NOTE(#461)` comment is removed.
2. `FindBaseline` now returns a 4th value, `reconstruct.StaleWarning`, carrying the signal. The console reconstruct handler appends a `stale_baseline:` entry to the existing `Warnings` DTO so Time-travel renders it.
3. The shim keeps the server-side log only and **discards** the warning value — the in-band MySQL signal is explicitly deferred by #466 as "needs design".

## FindBaseline signature change + callers

`FindBaseline(ctx, source, schema, table, at) → (path, snapshotTime, StaleWarning, err)`. All **9** call sites updated:
- `cmd/bintrail/reconstruct.go` — discards (already logged server-side)
- `internal/console/reconstruct.go` — **surfaces** via `appendStaleWarning`
- `internal/shim/snapshot.go` ×2 — discards (in-band deferred)
- `internal/reconstruct/fulltable.go` ×2 — discards
- `internal/reconstruct/reconstruct_test.go` ×2 — test call sites

## Tests (unit)

- partial run → `_INCOMPLETE`, skipped by `DiscoverBaselines`; clean run → `_SUCCESS`; legacy markerless → complete-by-default.
- local stale fallback returns a populated `StaleWarning` and reaches the console `Warnings` DTO (and survives JSON encoding).
- `s3IncompleteSnapshots` and `s3NewestSnapshot` proven against **real DuckDB glob()** on a local path (the S3-specific marker/newest logic).

`go build ./...`, `go vet`, and the unit + integration suites are green for the touched packages (Docker MySQL on :13306).

Closes #467
Closes #466


nethalo:
Applied the 2-lens adversarial-review fixes (pushed `2f6b317`):

**#1 (HIGH) — advisory staleness glob no longer fails a found recovery.** `findBaselineS3` had `best := candidates[0]` (a valid located baseline) followed by `s3NewestSnapshot(...)` — a broad `prefix/*/*/*.parquet` glob whose only job is the advisory `StaleWarning`. Its error path did `return "", time.Time{}, StaleWarning{}, err`, throwing away the already-found baseline. On the per-request shim `_snapshot` / console reconstruct hot path (no caching), a transient S3 blip on the broadest glob would fail a recovery that pre-PR succeeded — the inverse of this PR's goal. Extracted the post-`best` step into `staleWarningS3`, which now `slog.Warn`s and returns an empty `StaleWarning` ("not stale") on glob error, returning the located baseline. The correctness filters (`s3IncompleteSnapshots`, table-scoped glob) stay fatal.

**#2 (LOW) — marker scan fails loud.** `s3IncompleteSnapshots`' per-row `Scan` error was `continue` (silent drop); a dropped `_INCOMPLETE` row would demote its partial snapshot to complete-by-default (residual #467). Now `return nil, fmt.Errorf("scan S3 baseline marker path: %w", err)`, mirroring the already-hardened `listBaselinesS3` branch.

**#3 (MEDIUM) — IMPLEMENTED the S3 upload `_INCOMPLETE` symmetry (not deferred).** The local `Run` path writes `_INCOMPLETE`→`_SUCCESS`, but the S3 upload only did files→`_SUCCESS`, so a mid-upload death left a markerless partial that `SnapshotComplete` treats as complete — #467 was NOT closed on the upload path. `uploadBaselineToS3` now: (1) publishes a zero-byte `_INCOMPLETE` per snapshot dir FIRST, (2) uploads data files, (3) `_SUCCESS` LAST, (4) best-effort deletes `_INCOMPLETE`. New `storage.PutEmptyObject`/`DeleteObject` helpers (the local `_INCOMPLETE` file is gone post-`Run`, so there's nothing to walk — markers are keyed off the snapshot dir via the new `snapshotDirsWithSuccess`). Two deliberate scope calls: (a) the trailing `_INCOMPLETE` delete is best-effort because `SnapshotComplete`/`s3IncompleteSnapshots` decide completeness by `_SUCCESS`-present-first, so a leftover `_INCOMPLETE` next to a published `_SUCCESS` is harmless; (b) no `DeleteObject` of partial *data* on failure — resume via `--retry` overwrites in place.

**#4 (NIT) — operator note + error wording.** Added a note near `SnapshotComplete` (pre-release snapshots lack markers and are trusted complete-by-default; re-run `bintrail baseline` if a run may have been partial). `WriteSuccessMarker` now `slog.Warn`s (instead of erroring) a failed stale-`_INCOMPLETE` removal, so a genuinely-completed snapshot no longer surfaces as "could not write `_SUCCESS`".

**Tests added:** `TestStaleWarningS3_errorReturnsFoundBaseline` (closed-`*sql.DB` forces the broad glob to error → asserts the located baseline is returned, empty `StaleWarning`, warn logged); `TestS3IncompleteSnapshots_errorSurfaces` (marker read error surfaces, not silently complete); `TestSnapshotDirsWithSuccess` + `_missingDir` (the `_INCOMPLETE`-first ordering's discovery helper). Note: the upload *orchestration* itself (incomplete-first → data → success-last → delete) is exercised only via `snapshotDirsWithSuccess`; unit-testing the full ordering would mean switching the new helpers to the `s3API` interface, which I declined to match `UploadFile`'s concrete-`*s3.Client` precedent in `s3url.go`. The per-row Scan branch in #2 is defensive — `glob()` returns one VARCHAR, so a Scan can't be made to fail with real DuckDB — but the loud-fail contract is observable via the closed-DB test.

**Verify (Docker MySQL :13306 up):** `go build ./...` + `go vet` (unit + integration tags) green; `go test ./internal/reconstruct/... ./internal/baseline/... ./internal/storage/... ./internal/console/... ./cmd/bintrail/...` green unit AND `-tags integration`. `FindBaseline` stays at 4 return values (shim/console/reconstruct/cmd callers unchanged).

nethalo:
Applied the two IMPORTANT review findings (the original fix-agents hit repeated stream-idle timeouts on this PR; landed inline):

**FIX 1 — crash-safe local marker (`7477ed9`).** `baseline.Run` wrote its completeness marker only after `wg.Wait()`, so an uncatchable kill (OOM/SIGKILL/power loss) mid-conversion left a markerless partial snapshot that `SnapshotComplete` trusts as complete-by-default — the exact #467 silent loss, on the crash path. Now it creates the snapshot dir and writes `_INCOMPLETE` **before** launching the workers (mirroring the already-crash-safe S3 path); success replaces it with `_SUCCESS`. This also fixes a latent bug: a context cancelled before any table converted never created `snapDir`, so the old post-wait `markIncomplete` wrote into a non-existent directory and silently failed. Pinned by `TestRun_cancelledContextMarksIncomplete`.

**FIX 2 — pin the S3 upload ordering (`87f9177`).** The ordering (`_INCOMPLETE` first → data → `_SUCCESS` last → best-effort `_INCOMPLETE` delete) was correct by construction but untested. Extracted the four S3 ops into an injectable `s3ops` seam (closures over the concrete client — no storage API change, no new imports) and a testable `runBaselineUpload`; added a recording-mock test asserting the observable call order + the `--retry` skip-existing path. Also reworded the step-3 doc comment (`_SUCCESS` sorts before sibling dirs only for some DB-name first bytes; the deferral is what guarantees ordering).

Build + vet + scoped unit tests green for `internal/baseline`, `cmd/bintrail`, `internal/reconstruct`, `internal/storage`.

Issue #534, opened 2026-06-20
[pg] bintrail-pg binary + Postgres CI matrix (PG 14/15/16/17)


nethalo:
Part of #527 (PostgreSQL-as-source).

The PG-native artifact + its test matrix. Depends on the `internal/cli` extraction and `internal/pgcapture`.

## A. `cmd/bintrail-pg` binary
- A full PG-native CLI over the **shared** `internal/` packages (via `internal/cli`): `query`/`recover`/`reconstruct`/`shim`/`status`/`snapshot` + a PG `stream`/`watch`.
- PG-native flags (`--slot`, `--publication`, `--replica-identity`, LSN start position); **no** `--source-flavor` enum (it's PG-only).
- The MySQL `bintrail` binary does **not** import `internal/pgcapture`, so pgx/pglogrepl are not linked into its shipped artifact (Go links only the import graph the `main` reaches).
- Add to goreleaser (already multi-binary: `bintrail`, `bintrail-console`, `bintrail-mcp`).
- The supervisor/all-in-one daemon (`watch`) follows capture: a `bintrail-pg watch` embeds the shared read-only console; the standalone `bintrail-console serve` stays the shared multi-source viewer.

## B. Postgres CI matrix
- Compose service + `testutil` helpers (`SkipIfNoPostgres`/`CreateTestPostgres`) paralleling the MySQL/MariaDB harness.
- Versions: **PG 14, 15, 16, 17** (primary 16/17; PG 14 = declared floor; PG 18 smoke once stable; PG 13 documented-but-untested, EOL).
- A `//go:build postgres` tag is optional (the binary split already isolates the shipped-artifact deps); a **CI axis that builds + tests the pg path is mandatory** or it silently rots on the next `Event`/CLI refactor.

## Acceptance
- `bintrail-pg` builds, ships via goreleaser, and runs the full read/recover surface against a real PG source end-to-end.
- CI exercises capture + index + recover/reconstruct against PG 14–17.

Issue #530, opened 2026-06-20
[pg] internal/pgcapture: decode pgoutput logical replication into parser.Event


nethalo:
Part of #527 (PostgreSQL-as-source). **The big rock.**

A new `internal/pgcapture` package that decodes a PostgreSQL logical-replication (`pgoutput`) stream into the neutral `Event` (see the `parser.Event` extraction issue). Shares zero code with the go-mysql binlog parser.

## Scope
- Open a replication connection (`pgx/v5` `pgconn` with `replication=database`).
- Create/adopt a `pgoutput` replication slot + a publication for the in-scope tables.
- Decode the message stream: `Relation` (schema/columns — names arrive **in-band**, no ordinal indirection needed), `Begin`/`Commit` (txn bounds), `Insert`/`Update`/`Delete`, `Truncate`, `Type`, `Origin`.
- Map each row message into the neutral `Event` (`Schema`, `Table`, `EventType`, `PKValues`, `RowBefore`/`RowAfter`, LSN position).
- **Never silently drop a column**: the `'u'` unchanged-TOAST marker → an explicit reserved sentinel in `RowAfter` (the never-drop floor).
- Advance the durable checkpoint (`confirmed_flush_lsn`) only at commit boundaries (the #491 invariant — never checkpoint a half-streamed txn).

## Proven in the spike (drafts/postgres-spike-results-2026-06-20.md)
- `pglogrepl` + `pgx/v5` decode cleanly in Go (pure Go, no new CGO).
- INSERT/UPDATE/DELETE map cleanly to a `parser.Event`-shaped struct; the neutral Event then flows through the real indexer.
- **Footgun**: start replication from `CreateReplicationSlot().ConsistentPoint`, NOT `IdentifySystem().XLogPos` (the latter is past any pre-`StartReplication` DML → empty stream).
- `proto_version '1'` suffices (in-progress-txn streaming / binary mode are unnecessary and counter to commit-boundary checkpointing).

## Acceptance
- A live PG change stream is decoded into neutral `Event`s and indexed into `binlog_events` via the existing indexer.
- TOAST unchanged-value handling produces the explicit sentinel, never a missing column.
- Pin the `pglogrepl`/`pgx` versions.

Issue #529, opened 2026-06-20
[pg] Extract the command layer to a shared internal/cli (parameterized by source)


nethalo:
Part of #527 (PostgreSQL-as-source).

**Precondition for the separate `bintrail-pg` binary.**

The cobra command builders live in `cmd/bintrail` as `package main`, so they are **not importable** by a second binary. Without extracting them, `bintrail-pg` would have to duplicate the entire CLI surface (`query`/`recover`/`reconstruct`/`shim`/`status`/`snapshot`/…) — the exact duplication we are avoiding.

## Scope
- Extract the command builders into a shared package (e.g. `internal/cli`) parameterized by source type, so each `main` is ~20 lines: `cli.Register(root, sourceType)`.
- `cmd/bintrail` becomes a thin entrypoint over `internal/cli` (MySQL/MariaDB source).
- This sets up `cmd/bintrail-pg` (separate issue) to register the same shared commands with PG-native flags.

## Notes
- A PG DBA running `bintrail recover` (MySQL-named) defeats the PG-native point — the full read/recover surface must be reachable under `bintrail-pg`, hence sharing the command layer rather than the binary.
- Mirrors the existing `bintrail-console`/`bintrail-mcp` multi-binary pattern, but those have narrow surfaces and don't reuse `cmd/bintrail`'s commands; this extraction is what makes full-surface reuse cheap.

## Acceptance
- `cmd/bintrail` behavior byte-identical (full suite green).
- The command set is importable and registrable from a second `main` with a source parameter.

Issue #527, opened 2026-06-20
Epic: PostgreSQL-as-source (alpha → beta)


nethalo:
Tracking issue for adding **PostgreSQL as a source** engine (capture the WAL via logical decoding / `pgoutput`; the index stays MySQL). Sibling to the MariaDB epic #517.

bintrail is a data-*recovery* tool, so the bar is: capture is faithful, the silent-data-loss classes are closed, and the advertised surfaces are covered. Staged **alpha-first** (like MariaDB): the capture spike is done and de-risked the unknowns; the remaining work is engineering.

---

## ✅ CLOSED 2026-06-23 — PostgreSQL-as-source reached BETA

All 6 beta-BLOCKING data-safety gates are closed:

- #534 (binary + CI), #531 (RI-FULL validator + PK recovery), #532 (slot/WAL + durable badge + reset), #533 (type fidelity) — closed, genuinely delivered.
- **TOAST resolve-on-read** — met *upstream* (resolve-at-decode under the required RI FULL; the sentinel is dead code for supported sources). Read-side fail-loud hardening tracked in the GA epic (#592).
- **DDL drift detection** — met *by-construction* (synchronous `RelationMessage` shape-swap + fail-loud backstops). Test coverage + the last silent-loss guards landed in **#596** (closes #591 DDL-drift test, #555 UNLOGGED, #556 cascade-child, #559 TimescaleDB) — the 3-agent close-out review caught a CRITICAL (the #555 guard was dead code: it queried `pg_publication_tables`, where an UNLOGGED table never appears) and a HIGH (silent probe-failure), both fixed before merge.

**bintrail now supports PostgreSQL as a source at BETA.** `recover` is the supported recovery surface; capture is type-faithful, RI-FULL-enforced, slot/WAL-monitored, DDL-drift-safe, and the silent-loss classes (UNLOGGED, cascade-child, TimescaleDB, TOAST) are surfaced loud.

**beta → GA continues in #597**: baseline@`consistent_point` LSN → reconstruct/time-travel (#593), managed-PG smoke matrix (#594), console source-aware presentation incl. forensics-degraded note (#595), sequence synthesis (#558), ICP survey (#535), + hardening (#592 and others).

---

## Decisions (made 2026-06-20)

[...]

PR #561, opened 2026-06-21
feat(bintrail-pg): standalone PostgreSQL-source binary (#534)

nethalo:
## What

Adds **`cmd/bintrail-pg`** — the PostgreSQL sibling of the core `bintrail` binary, the first slice of #534. It captures a live PostgreSQL logical-replication stream into the **same MySQL index** (the "one index schema for all sources" red line, #527) and serves the **identical read plane** — `status`/`query`/`recover`/`reconstruct`/`shim` — via `internal/cli.AddReadCommands`.

This is thin, well-tested glue over the already-merged producer/consumer (`internal/pgcapture` #530, `internal/pgstreamrun`); the capture engine itself is unchanged.

## Why a separate binary

The PostgreSQL capture path links `jackc/pgx` + `pglogrepl`, which the core MySQL binary must stay free of. Splitting the **capture plane** per source keeps each binary's dependency surface honest, while the **read plane** is shared code in `internal/cli` (the payoff of the #529 extraction). The user wants a recognizable `bintrail-pg` artifact for the Postgres-recovery niche.

## Changes

- **`cmd/bintrail-pg/main.go`** — `rootCmd "bintrail-pg"`, reusing the core's `main.Version/CommitSHA/BuildDate` ldflags vars so the Makefile `BINTRAIL_LDFLAGS` applies verbatim (exactly as `bintrail-console` does) + `cli.AddReadCommands`.
- **`cmd/bintrail-pg/stream.go`** — `bintrail-pg stream` wires flags/env → `pgstreamrun.One`. PostgreSQL needs **two distinct connections**: `--repl-dsn` (walsender mode, cannot run SQL) and `--query-dsn` (catalog/PK lookups). `index-dsn`/`server-id` reuse `cli.EnvBindings`; the `BINTRAIL_PG_*` connection vars get a direct env fallback. Config assembly is a pure, unit-tested seam (`pgStreamConfigFromFlags`), mirroring the MySQL stream's `streamConfigFromFlags`.
- **`cmd/bintrail/pgfree_test.go`** — decouple guard: the core MySQL binary must NOT link `internal/pgcapture`, `internal/pgstreamrun`, `jackc/pgx`, or `jackc/pglogrepl` (mirrors `uifree_test.go`).
- **`THIRD-PARTY-NOTICES` + `scripts/`** — `bintrail-pg` links `jackc/pgx` + `pglogrepl` into a published binary (previously unreachable from any main), so `gen-notices.sh` scans the fourth main and the regenerated notices include their licenses (jackc `pgio`/`pglogrepl`/`pgpassfile`/`pgservicefile`/`pgx`). The dependency-graph hash is unchanged (pgx entered `go.mod` in #530), so `check-notices` stays green.
- **Makefile** — `make build-pg` / `install` / `clean` / `all` wire the new binary.

## Tests

- `go test ./... -count=1` — green (33 pkgs).
- `cmd/bintrail-pg` unit tests cover the config seam: missing-required aggregation, happy path, `BINTRAIL_PG_*` env fallback, flag-wins-over-env, and `--start-lsn` parse (valid + invalid).
- Guard test passes (`cmd/bintrail` does not link the pgx stack).
- Smoked the three end-user error paths (required-flag, missing-PG-settings, bad start-lsn) — all clear, actionable messages.

## Scope / follow-ups (rest of #534)

- **Postgres CI matrix** (next slice) — compose service + `testutil.CreateTestPostgres` + CI axis (PG 14–17). This is what finally gives `TestOne_EndToEnd_PostgresToIndexToRecovery` (currently `SkipIfNoPostgres`) a runner; the capture path has no CI coverage until it lands.
- **goreleaser** (later slice) — Docker/deb/rpm publishing for `bintrail-pg`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ BINARY_NAME=bintrail
 MCP_BINARY=bintrail-mcp
 GATEWAY_BINARY=mcp-gateway
 CONSOLE_BINARY=bintrail-console
+PG_BINARY=bintrail-pg
 VERSION=$(shell git describe --tags --always --dirty 2>/dev/null || echo "dev")
 COMMIT=$(shell git rev-parse --short HEAD 2>/dev/null || echo "none")
 BUILD_DATE=$(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
@@ -9,12 +10,12 @@ BUILD_DATE=$(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
 BINTRAIL_LDFLAGS=-ldflags "-X main.Version=$(VERSION) -X main.CommitSHA=$(COMMIT) -X main.BuildDate=$(BUILD_DATE)"
 MCP_LDFLAGS=-ldflags "-X main.mcpVersion=$(VERSION)"
 GATEWAY_LDFLAGS=-ldflags "-X main.gatewayVersion=$(VERSION)"
-# bintrail-console reuses BINTRAIL_LDFLAGS: it injects the same
-# main.Version/CommitSHA/BuildDate vars as the core binary.
+# bintrail-console and bintrail-pg both reuse BINTRAIL_LDFLAGS: they inject the
+# same main.Version/CommitSHA/BuildDate vars as the core binary.
 
-.PHONY: all build build-mcp build-gateway build-console clean test console-e2e lint install build-all tidy deps notices check-notices
+.PHONY: all build build-mcp build-gateway build-console build-pg clean test console-e2e lint install build-all tidy deps notices check-notices
 
-all: build build-mcp build-gateway build-console
+all: build build-mcp build-gateway build-console build-pg
 
 build:
 	go build $(BINTRAIL_LDFLAGS) -o $(BINARY_NAME) ./cmd/bintrail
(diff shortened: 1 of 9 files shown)


nethalo:
## Three-agent review applied (code-reviewer + silent-failure-hunter + pr-test-analyzer)

All three returned **sound design, ship-ready, zero Critical/Important blocking issues**. Non-blocking findings applied in `a6be8c8`:

| # | Finding | Resolution |
|---|---------|------------|
| pr-test 4.1 | `resetPGFlags` hardcodes int defaults independent of the `IntVar` registrations → a default drift fails no test | Added `TestPGStreamCmd_defaults` pinning `--batch-size`/`--checkpoint`/`--partitions` `DefValue` (mirrors the MySQL `TestStreamCmd_defaults`) |
| pr-test 4.2 | `BINTRAIL_PG_START_LSN` env binding never exercised | Added a `valid-from-env` subcase driving it via `t.Setenv` |
| silent-failure 2 | `json.Encode` error on the error-reporting path discarded | Fall back to text if the JSON encode fails — applied to **both** mains to preserve the deliberate parity |
| code-review 4 | header em-dash line-wrap | Tidied; `THIRD-PARTY-NOTICES` regenerated to match |

**Deliberately deferred (with rationale), not dropped:**
- **code-review 2 (`--reset`/escape hatch):** a MySQL-style `--reset` is *semantically misleading* for PostgreSQL — the replication slot governs the resume position (`ensureSlot` clamps forward; older WAL may be reclaimed), so clearing the checkpoint alone does **not** rewind to `--start-lsn`. Instead of shipping a misleading flag, the stream help now documents the real re-seed mechanic (drop the slot + clear the checkpoint row). A first-class reset/re-baseline UX belongs in the capture-completeness slice, which will have a live-PG CI runner to test it.
- **code-review 1 (Prometheus metrics on the PG daemon):** needs `pgstreamrun` to grow metrics fields like `streamrun` has — out of reach for a binary-glue PR. Tracked for the same later #534 slice.

**Confirmed-correct highlights from the reviews:** dropped `FatalCloseError` mapping is *provably* safe (`*agent.FatalCloseError` is unreachable in bintrail-pg — no `internal/agent` in its dep graph); `--partitions` on `stream` is necessary (no `init` command, so `pgstreamrun.One`→`CreateIndexTables` bootstraps); the decouple guard's prefix-match correctly catches `jackc/pgx/*`; the 12-field `Config` mapping (incl. seconds→`Duration`) is exact; both red lines upheld (shared MySQL index; pgoutput client, slot/publication validated-not-created).

PR #562, opened 2026-06-21
ci(postgres): integration-postgres-source matrix (PG 14-17) (#534)

nethalo:
## What

Adds the **`integration-postgres-source`** CI job — slice 2 of #534 — giving the PostgreSQL capture path a **live CI runner**. Until now `TestCapturer_*` (`internal/pgcapture`) and `TestOne_*` (`internal/pgstreamrun`) skip in every job (no `BINTRAIL_TEST_PG_DSN` is set anywhere), so the entire PG→capture→index→recovery pipeline merged in #530/#531 had **zero CI coverage**. This closes that gap.

## The job

Mirrors `integration-mariadb-source`: a new PostgreSQL **source** container paired with the existing MySQL **index** container. Both must be present because the e2e test (`TestOne_EndToEnd_PostgresToIndexToRecovery`) streams PG changes *into* the MySQL index and generates recovery SQL — a single `integration` matrix cell can't host two servers, and the PG tests need `wal_level=logical` + replication slots the MySQL container can't provide.

- **matrix:** `postgres: ["14","15","16","17"]` (14 = declared minimum, PG13 is EOL; 17 = newest GA), `fail-fast: false` to match the MySQL matrix.
- **PG container:** `wal_level=logical`, `max_replication_slots/wal_senders=10`, host port **15432** (avoids any PostgreSQL the runner image ships), **TCP** `pg_isready` readiness (dodges the first-init unix-socket-only false-ready race).
- **scoped by path** (`./internal/pgcapture/ ./internal/pgstreamrun/`), not `-run`: streamrun has its own `TestOne_*` *MySQL* tests, so a name filter would be ambiguous. `-p 1` serializes (the PG tests share one server, creating/dropping their own slots/publications).

## testutil

`SkipIfNoPostgres` now honors `BINTRAIL_REQUIRE_POSTGRES=1` (set by this job): a missing DSN **fails loud** instead of green-via-skip, mirroring the existing `MariaDBRequired` guard. The MySQL matrix and console-e2e leave it unset, so PG tests keep skipping cleanly there.

## Verification

- The **exact job command** passes against **live `postgres:14` AND `postgres:16`** + MySQL — all 7 PG integration tests (5 `pgcapture` + 2 `pgstreamrun`, incl. the full e2e + recovery). PG14 is the cell that exercises the `indkey::int2[]` cast added in #547.
- Both require-flag modes confirmed: `REQUIRE=1` + no DSN → **FAIL**; neither set → **SKIP**.
- `ci.yml` parses (PyYAML); `go vet -tags=integration` clean on both PG packages.

## Scope

This is the **de-risk slice** — it makes the previously-skipped PG e2e actually run. The remaining #534 slice is **goreleaser** publishing (Docker/deb/rpm) for `bintrail-pg`. Deferred capture-completeness items (`--reset`, Prometheus metrics on the PG daemon) now have a live-PG runner to be tested against when implemented.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,90 @@ jobs:
       - name: MariaDB-source integration tests
         run: go test -tags=integration -race -p 1 -count=1 -run 'MariaDB|Mariadb|mariadb|Flavor|FourColumns' ./...
 
+  integration-postgres-source:
+    # PostgreSQL-as-source (#534). Like integration-mariadb-source, this pairs a
+    # NEW PostgreSQL SOURCE container with the existing MySQL INDEX container
+    # (13306): two containers a single `integration` matrix cell can't host, and
+    # the PG logical-replication tests need wal_level=logical + replication slots
+    # the MySQL index container can't provide. The end-to-end test
+    # (TestOne_EndToEnd_PostgresToIndexToRecovery) streams PG changes INTO the
+    # MySQL index and generates recovery SQL, so both servers must be present.
+    runs-on: ubuntu-22.04
+
+    # Spans the supported PostgreSQL range: 14 (declared minimum — PG13 is EOL)
+    # through 17 (newest GA). fail-fast: false so a regression on one version
+    # still reports the others, matching the MySQL matrix.
+    strategy:
+      fail-fast: false
+      matrix:
+        postgres: ["14", "15", "16", "17"]
+
+    env:
+      BINTRAIL_TEST_DSN: "root:testroot@tcp(127.0.0.1:13306)"
+      # Host port 15432 (not 5432) avoids colliding with any PostgreSQL the
+      # runner image ships; the container-internal port stays 5432. sslmode is
+      # disabled — the stock postgres image serves plaintext. replDSN appends
+      # replication=database onto this for the WAL connection.
+      BINTRAIL_TEST_PG_DSN: "postgres://postgres:testpg@127.0.0.1:15432/pgtest?sslmode=disable"
+      # A PostgreSQL source is guaranteed present here, so the PG tests must FAIL
+      # (not silently skip) if the DSN is missing — guards against green-via-skip.
+      # The 8.0/8.4 matrix and console-e2e leave it unset, so PG tests skip there.
+      BINTRAIL_REQUIRE_POSTGRES: "1"
+
+    steps:
+      - uses: actions/checkout@v5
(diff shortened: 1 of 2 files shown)


nethalo:
## Two-agent review applied (code-reviewer + silent-failure-hunter)

Both returned **sound / shippable, no Critical or Important blockers**. Fixes applied in `5112b59`:

| Agent | Finding | Resolution |
|-------|---------|------------|
| silent-failure (MEDIUM) | Both scoped packages also hold **non-tagged unit tests**, so a build-tag/path drift that silently dropped the integration tests would exit 0 green with **zero PG coverage** — the `BINTRAIL_REQUIRE_POSTGRES` guard only fires when an integration test *runs*, so it's blind to "didn't run at all." | Run with `-v` + `tee` and **assert `TestOne_EndToEnd_PostgresToIndexToRecovery` genuinely PASSed** — a positive tripwire, not just `go test` exit 0. Verified both directions against live PG16+MySQL (passes; a nonexistent-name grep fires it → exit 1). |
| code-reviewer (Suggestion) | The path-scoping comment claimed streamrun has its own MySQL `TestOne_*` tests — **false** (the only `TestOne_*` are the two PG tests). The code is correct & more robust; only the stated reason was wrong. | Reworded to the real rationale (naming-independent selection). Also dropped the now-stale "17 (newest GA)" claim (PG18 has GA'd; its stream changes tracked with beta work). |

**Confirmed sound by both:** the require-flag fail-vs-skip logic mirrors `MariaDBRequired` exactly; the env-var name matches (`BINTRAIL_REQUIRE_POSTGRES`=="1"); all 7 PG integration tests route through `SkipIfNoPostgres`; the TCP `pg_isready` probe genuinely defeats the first-init false-ready race; matrix cells are distinct; readiness loops fail red on timeout; in-test assertions (`pg_column_size>2000`, TOAST non-empty, `n!=0` reversals) prevent vacuous green; `${{ matrix.postgres }}` is author-controlled (no injection).

### Surfaced, NOT changed here (out of diff, pre-existing, repo-wide)
silent-failure (HIGH): `main` branch protection has **empty `required_status_checks`** — a RED job (this one *or* the existing unit/integration/mariadb jobs) does not block a merge at the gate. This is a repo branch-protection setting affecting all jobs equally, not something this PR can fix; flagging for a maintainer decision. (The repo's `--admin` squash-merge flow gates on green CI manually today.)

Issue #577, opened 2026-06-22
Surface recover-cascade in the web console (free query_explorer tier)


nethalo:
## Goal

Bring the `recover-cascade` capability (FK `ON DELETE CASCADE` / `ON DELETE SET NULL` recovery, shipped free-core in **v0.18.0**, epic #548) to the **web console** so a DBA doing incident response can reconstruct cascade-deleted children from the UI — not just the CLI.

## Why it fits (architecture)

- The console already has `POST /api/recover` → `recovery.GenerateSQLFromRows` → returns **SQL text, never executes** (`internal/console/api.go`). `recover-cascade` is dry-run-only by design → same contract, no conflict with the "binary never touches a server process" red line.
- **Proven precedent**: the Time-travel / reconstruct tab (endpoint + `/api/capabilities` + `cap-on` frontend gate) is the exact template to follow.
- The console already imports `internal/recovery`, `metadata`, `reconstruct`; `internal/cascade` is a lightweight new import (depends only on `query`, already in scope). The per-server `bundle` already carries the engine, resolver, and baseline source.

## Conceptual shape (not a flag on the recover tab)

The existing `recover` is *blast-radius-by-filter* (schema/table/window → reverse what matches). `recover-cascade` is *blast-radius-by-discovery* (give the **parent** PK → find the invisible children InnoDB deleted below the binlog). Distinct input (parent table + PK), distinct output (synthesized victims + INCOMPLETE coverage caveats). It deserves its own flow, modeled on reconstruct.

## Open-core

Free, under the `query_explorer` surface — consistent with the CLI making `recover-cascade` free-core. (The stale "FK awareness = SaaS-only" note predates that decision.) The DTO must stay `connection_id`-free (the open-core boundary lives in `eventDTO`).

## Gate (cleaner than reconstruct)

reconstruct *requires* a baseline. Cascade does **not** — Phase-1 works without one; Phase-2 auto-enables when the server has a baseline configured (reuse `baselineConfigured` on the bundle). So the tab is available whenever `recover` is.

## Slices (dependency-ordered)

- [ ] **Slice A (#578)** — extract the shared cascade-recovery SQL emission (precondition: factor `emitCascadeSQL` + `cascadeExit` out of `internal/cli/recover_cascade.go` into a binary-neutral helper both the CLI and the console call, so the script format can't drift). No behavior change.
[...]

PR #582, opened 2026-06-22
refactor(cascade): extract cascade-recovery SQL emission to internal/cascaderecover (#578)

### LINKED ISSUE #578: Console cascade — Slice A: extract shared cascade-recovery SQL emission
Precondition for #577 (recover-cascade in the console).

Factor the cascade-recovery SQL assembly out of the CLI command layer so the CLI and the console produce **byte-identical** scripts and can't drift.

## Scope
- `internal/cli/recover_cascade.go` currently owns `emitCascadeSQL` (preamble + `SET FOREIGN_KEY_CHECKS=0/1` wrapper + `GenerateSQLFromRows` for CASCADE INSERTs + `FormatSetNullRestore` for SET NULL UPDATEs, all-or-nothing pre-validation) and `cascadeExit` (the complete/incomplete exit contract).
- Move the **emission** logic into a binary-neutral home (e.g. a new `internal/cascade` emit helper or a small `internal/cascaderecover` package) that takes the synthesized `cascade.Result` + a `*recovery.Generator` + resolver and writes the script to an `io.Writer`. No cobra, no command globals.
- The CLI keeps its flag parsing / exit-code mapping; it calls the extracted emitter.

## Constraints
- **No behavior change.** The generated SQL must be identical before/after (pin with a golden/string test).
- Keep the all-or-nothing pre-validation (build every SET NULL statement before writing a byte) and the `… AND fk IS NULL` guard — these were #571-review fixes; don't regress them.
- Keep the structured coverage (`Result.Incomplete`/`Complete()`) accessible to callers so the console can surface it.

## Done when
- Both the CLI and a hypothetical second caller can emit the cascade script via the shared helper; CLI tests still green; a test pins the script format.

nethalo:
closes #578

## Summary

Slice A of console cascade (#577) — a precondition: factor the cascade-recovery SQL **emission** out of the CLI command layer so the CLI and the future console caller produce **byte-identical** scripts and can't drift.

- **New `internal/cascaderecover`** (`EmitSQL` + `Header`): binary-neutral, no cobra, no command globals. A leaf composition layer over `cascade`+`recovery`+`metadata`+`query`; **none of its deps import it back**, so the `cascade` synthesis engine stays free of emission/recovery coupling. (Placement is decoupling-driven, not cycle-avoidance — neither candidate package creates a cycle; `go list` confirms `recovery`/`query`/`metadata` never import `cascade`.)
- **`internal/cli/recover_cascade.go`**: `emitCascadeSQL`/`cascadeHeader` deleted; the two call sites (JSON + text) now use `cascaderecover.EmitSQL`/`Header`. Flag parsing and the exit-code mapping (`cascadeExit`) **stay in the CLI** per the issue's split.

## Faithful extraction + one deliberate API tightening

This is **not** a pure byte-for-byte move. It is a faithful relocation **plus** one small, behavior-preserving API change driven by review:

- The relocated emit body is otherwise identical to the original (verified by a normalized source diff before the tightening).
- **Tightening for the second caller (#577):** the SET NULL count in the preamble is now derived from `len(setNullRows)` inside `EmitSQL` rather than carried as a `Header.SetNulls` field. At the CLI call site the two were always equal, so **the generated SQL is unchanged** — but a future caller can no longer desync the header count from the statements actually emitted. (The field is unreleased, so folding this in is safe.)

The **output** byte-identity is what matters and is proven below.

## Invariants preserved (constraint: no behavior change)

- **Byte-identical SQL** — locked by a new DB-free golden test (full-string compare, not `Contains`) across both preamble branches (Phase-1 / Phase-2-active), the caveats block, and the SET NULL section; the existing CLI integration suite pins the same output through the full command.
- **All-or-nothing SET NULL pre-validation** (#571 fix) — every SET NULL statement is built before a single byte is written; a nil resolver / unresolvable table / absent PK column aborts with **zero bytes** written (so a script can never open `FK_CHECKS=0` and never close it). Locked by two negative-path tests asserting `buf.Len()==0`.
- **`… AND fk IS NULL` idempotency guard** — untouched; `EmitSQL` keeps delegating to `recovery.FormatSetNullRestore`.
- **Statement-count contract** (`n`) — unchanged (CASCADE INSERTs + 1 per SET NULL UPDATE), still feeds the JSON `statements` field.

## Forward-compat note for #577 (not in scope here)

The **"complete" determination** (`len(caveats)==0 && synthErr==nil`) and the **caveats assembly** (`res.Incomplete` + the archive/limit/synth-error caveats) still live in the CLI command (`recover_cascade.go`), **not** in the shared package. That's correct for slice A (the issue scopes this to *emission* and keeps `cascade.Result.Incomplete/Complete()` accessible to callers). But if the #577 console re-implements that logic slightly differently, the two surfaces could drift on coverage — the exact failure mode this extraction exists to prevent. A later slice should share that determination consciously. (Filed as a note on #577.)

## Review

Three specialized agents reviewed the diff (code-reviewer, silent-failure-hunter, type-design-analyzer): faithful extraction, all error semantics preserved, all-or-nothing invariant intact. Their comment-accuracy nits (a self-contradictory package-doc line, a stale `emitCascadeSQL` reference) and the `SetNulls` tightening above are applied.

## Test plan

- [x] `go test ./... -count=1` (full unit suite; new golden runs DB-free)
- [x] `go vet ./internal/cli/ ./internal/cascaderecover/`
- [x] `go test -tags integration ./internal/cli/ -run TestRecoverCascade` (full-command SQL pinning, all green)

### DIFF
--- a/internal/cascaderecover/emit.go
+++ b/internal/cascaderecover/emit.go
@@ -1,0 +1,125 @@
+// Package cascaderecover assembles the cascade-recovery SQL script (the
+// documented preamble, the SET FOREIGN_KEY_CHECKS=0/1 wrapper, the CASCADE
+// re-INSERTs, and the idempotent SET NULL restorations) from a synthesized
+// cascade result.
+//
+// It is the binary-neutral home for the emission logic that used to live in the
+// `recover-cascade` CLI command: factoring it out lets the CLI and the console
+// (#577) produce BYTE-IDENTICAL scripts that cannot drift. The package depends
+// on cascade (for the SetNullRestore rows), recovery (the reversal-SQL
+// generator), metadata (the child PK columns), and query (the row type); nothing
+// it depends on imports it back, so it stays a leaf composition layer (no import
+// cycle) and keeps the cascade synthesis engine free of any emission/recovery
+// coupling.
+//
+// It owns no cobra flags and no command globals. Callers keep their own flag
+// parsing and exit-code mapping; the structured coverage (cascade.Result's
+// Incomplete/Complete()) stays with the caller so each surface can present it.
+package cascaderecover
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/dbtrail/dbtrail/internal/cascade"
+	"github.com/dbtrail/dbtrail/internal/metadata"
+	"github.com/dbtrail/dbtrail/internal/query"
+	"github.com/dbtrail/dbtrail/internal/recovery"
+)
+
+// Header carries the values rendered into the SQL preamble. Parents and Children
+// cannot be derived from EmitSQL's flattened rows (it cannot split the
+// parents++victims concat back apart), so the caller supplies them; the SET NULL
+// count is NOT carried here — EmitSQL derives it from len(setNullRows) so a
+// caller can never desync the header count from the statements emitted.
+type Header struct {
+	Schema, Table     string
+	Parents, Children int
+	Caveats           []string
+	BaselineActive    bool
+}
+
+// EmitSQL writes the documented preamble, the FK-checks-off wrapper, the CASCADE
+// reversal statements (DELETE→INSERT via the generator), and the SET NULL FK
+// restorations (idempotent guarded UPDATEs). Returns the total statement count.
+// resolver supplies child PK columns for the SET NULL WHERE clauses; gen must be
+// built from the same (or an equivalent) resolver (recovery.New(db, resolver)).
+func EmitSQL(w io.Writer, gen *recovery.Generator, rows []query.ResultRow, setNullRows []cascade.SetNullRestore, resolver *metadata.Resolver, hdr Header) (int, error) {
+	// Build every SET NULL restoration BEFORE writing a byte (all-or-nothing): a
+	// missing resolver, an unresolvable table, or an absent PK column must abort
+	// the whole emit cleanly — returning mid-script would leave the parent/child
+	// INSERTs written but drop the closing `SET FOREIGN_KEY_CHECKS=1`, handing the
+	// operator a script that re-enables nothing.
+	var setNullStmts []string
+	if len(setNullRows) > 0 {
+		if resolver == nil {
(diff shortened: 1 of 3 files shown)


nethalo:
## Comprehensive review pass (`/pr-review-toolkit:review-pr`)

Five specialized agents reviewed the final diff. **No critical or blocking issues.**

| Aspect | Agent | Verdict |
|---|---|---|
| General code | code-reviewer | **Clean** — no issues ≥80% confidence; byte-identical CLI output confirmed via `git show`, no dangling refs, builds/vets/tests green |
| Error handling | silent-failure-hunter | **Preserved** — all-or-nothing pre-build intact (the count change writes to the in-memory builder *before* the first `w` write); `n` correct on every late writer-error path |
| Type design | type-design-analyzer | **Clean pass** — prior `SetNulls`-desync finding resolved; invariant expression & enforcement rose 3/5 → 4/5 (illegal "count ≠ statements" state is now unrepresentable) |
| Comments | comment-analyzer | **Accurate** — both prior comment-rot fixes confirmed; one optional wording tweak applied (`SAME`→`same (or an equivalent)` resolver) |
| Tests | pr-test-analyzer | 2 worthwhile additions (below) — neither blocking |

### Applied from the review

- **Added `TestEmitSQL_setNullAbsentPKColumnIsAllOrNothing`** — covers the third all-or-nothing abort point the package doc advertises (PK column absent from the child row → `FormatSetNullRestore` fails → zero bytes written). Previously uncovered.
- **Added `TestEmitSQL_goldenSetNullMultiRow`** — a two-row SET NULL golden that locks the one non-verbatim change in this PR (`Header.SetNulls` → `len(setNullRows)`): preamble reads "restores 2", the section header emits exactly once, and `n` accumulates to 2 — pinning *header count == statements emitted == n* for N>1.
- Doc wording tweak on the `gen ⇔ resolver` precondition.

Coverage 87.2% → **89.4%**; all 8 `cascaderecover` tests + the CLI integration suite green.

PR #588, opened 2026-06-23
feat(console): cascade recovery frontend tab (#580)

### LINKED ISSUE #580: Console cascade — Slice C: Cascade recovery frontend tab
Frontend for #577. Depends on Slice B.

## Scope (model on the Time-travel/reconstruct tab in `internal/console/assets/app.js`)
- New route `"cascade"` + `renderCascade(params)`: a form with **parent** schema/table/PK (+ optional `since`/`until`, `lookback`, `max-depth`, `allow-incomplete`), a "Preview victims" (read-only) and a "Generate cascade recovery SQL" action, and a result code panel.
- **Coverage UX (important)**: surface the `incomplete[]` caveats and the `complete=false` state prominently (a visible banner, like reconstruct's gap warnings) — never let a partial recovery read as whole. Echo the INCOMPLETE framing from the CLI.
- Nav item with `data-route="cascade" data-capability="recover_cascade"`; `cap-on` gating + redirect-to-overview guard when the capability is off; command-palette entry gated the same way.
- Optional: an "Undo cascade" affordance from an event/DELETE row that deep-links into the tab pre-filled with the parent table+PK.

## Verification
- Browser-render check with real headless Chrome (per the project's console-frontend rule — a class/`[hidden]` gating bug is invisible to Go tests). Confirm the tab hides when the capability is off and the warnings render.

## Done when
- The tab generates cascade recovery SQL end-to-end against a live console, gating works on capability, coverage warnings are visible, headless-Chrome verified.

nethalo:
closes #580

## Summary

Slice C of console cascade (#577) — **completes the epic**. A new **Cascade recovery** tab in the read-only web console (modeled on Time-travel), driving the slice-B `POST /api/recover-cascade`. Never executes SQL.

- **`internal/console/assets/app.js`**: `renderCascade`/`runCascade` + the route wired in `ROUTES`, the `navigate()` capability guard, the `renderRoute` switch, and a ⌘K command — all gated on `recover_cascade`. Form takes the **parent** schema/table/PK + since/until/lookback/max-depth + an allow-incomplete checkbox; result is a code panel (Copy/Download) + a prominent coverage banner.
- **`internal/console/assets/index.html`**: nav item in the Resolve group with `data-route="cascade" data-capability="recover_cascade"` — the CSS `.cap-on` gating and the generic nav click binding are automatic (no JS wiring).

## Coverage UX (the issue's hard requirement)

A partial recovery must **never** read as whole. The banner is driven off the response `complete` flag (NOT the allow-incomplete checkbox, which is a server-side no-op) and surfaces `incomplete[]` — the cascade response has **no `warnings` key**, so a verbatim copy of the reconstruct tab would silently drop every caveat. When `complete=false`, a bold "INCOMPLETE — provably partial" lead caveat is prepended, echoing the CLI framing.

## Clone-bug guards

`max_depth` is sent as a **JSON int** (a string fails Go's unmarshal); `allow_incomplete` as a **real boolean**; `lastSQL` is set **before** `codePanel` so Copy/Download serve the cascade SQL, not stale.

## Scope (Occam)

**One "Generate" action** — the endpoint is dry-run and returns SQL+counts together, so `victim_count`/`set_null_count` ARE the victim preview (no victim-rows endpoint exists, and showing rows would breach the connection_id-free boundary). The optional event-row **"Undo cascade" deep-link is deferred**.

## Verification — real headless Chrome

`test/console-e2e/console_e2e.mjs` extended with **+12 cascade assertions (26/26 pass)**, driven against a live console with system Chrome: nav item + ⌘K entry appear with the capability on, the form renders, and — **the gating-bug class Go tests can't see** — the nav item **hides** and the route **redirects to overview** when `recover_cascade` is off.

## Test plan
- [x] `go build ./...`; `go test ./internal/console/` green (assets embedded)
- [x] `node --check` on app.js
- [x] Headless-Chrome E2E: 26/26 (12 new cascade assertions) via `PW_CHANNEL=chrome test/console-e2e/run.sh`

### DIFF
--- a/internal/console/assets/app.js
+++ b/internal/console/assets/app.js
@@ -39,7 +39,7 @@ const EVENT_CSV_COLUMNS = [
 const BADGE_CLASS = { UPDATE: "b-update", INSERT: "b-insert", DELETE: "b-delete" };
 function badgeClass(t) { return BADGE_CLASS[t] || "b-baseline"; }
 
-const ROUTES = ["overview", "events", "timetravel", "recover", "status", "storage"];
+const ROUTES = ["overview", "events", "timetravel", "recover", "cascade", "status", "storage"];
 
 const MON_STATE_TITLES = {
   failed: "stream failing — retrying with backoff; press Start for details",
@@ -558,6 +558,8 @@ function navigate(route, params, push = true) {
   if (!ROUTES.includes(route)) route = "overview";
   // Reconstruct surface is gated; never navigate to a disabled Time-travel.
   if (route === "timetravel" && !capsCache.reconstruct) route = "overview";
+  // Cascade recovery is the free tier but refused under an RBAC profile.
+  if (route === "cascade" && !capsCache.recover_cascade) route = "overview";
   // Storage is a watch-daemon surface (rotation + archiving live there).
   if (route === "storage" && !capsCache.monitor) route = "overview";
   const qs = params && Object.keys(params).length
@@ -576,6 +578,7 @@ function renderRoute() {
     case "events": return renderEvents(params);
     case "timetravel": return renderTimetravel(params);
     case "recover": return renderRecover(params);
+    case "cascade": return renderCascade(params);
     case "status": return renderStatus();
     case "storage": return renderStorage();
     default: return renderOverview();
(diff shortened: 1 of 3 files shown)


nethalo:
## Review pass (2 agents, frontend-focused) — clean

| Lens | Verdict |
|---|---|
| code-reviewer | **Clean** — route wired in all 5 spots; self-guard mirrors renderTimetravel (replaceState, no recursion); the 3 clone-bug guards correct (lastSQL-before-codePanel, max_depth int, allow_incomplete bool); serverGen staleness guard; no innerHTML. **Cross-checked every JSON key against the Go struct tags** — request/response/capability/route all match; `complete` has no `omitempty` so the banner can't false-fire from `undefined`. |
| silent-failure / coverage UX | **Sound** — banner driven off `data.complete` (NOT the checkbox no-op); uses `data.incomplete` (NOT the absent `data.warnings` key — the verbatim-copy trap is avoided); error path clears warnings + replaces stale SQL; zero-victim true-negative honest. Fail-safe direction (over-warns, never under-warns). Defense-in-depth: the caveats are also embedded in the SQL preamble, so the INCOMPLETE warning rides along in Copy/Download. |

No fixes required. Headless-Chrome E2E (26/26) already pins the gating/redirect that Go tests can't see.

nethalo:
## Comprehensive review (`/pr-review-toolkit:review-pr`, 5 agents) — applied

| Aspect | Verdict |
|---|---|
| code-reviewer | **Clean** — wiring in all 5 spots; cross-checked every JSON key vs the Go struct tags |
| silent-failure / coverage UX | **Sound** — partial never reads as whole; fail-safe; caveats also embedded in the SQL preamble |
| code-simplifier | **Clean pass** — faithful clone of renderTimetravel/generateUndo; any "simplification" would be a regression |
| comment-analyzer | **6/6 comments accurate**, exemplary "why" comments anchored in the live contract |
| pr-test-analyzer | Sound + **1 real gap (7/10)** → **fixed** |

### Applied (pr-test-analyzer)
- **Untested redirect self-guard closed**: the E2E only exercised the `navigate()` guard; `renderCascade`'s own `replaceState` self-guard (the sole defense for a bookmarked/typed `/cascade` URL or Back/popstate, which dispatch `renderRoute` directly, bypassing `navigate()`) could have been **deleted with the suite still green**. Added an assertion that calls `renderCascade()` directly on a forced `/cascade` URL and confirms it redirects to overview AND renders no form. Also swapped a `waitForTimeout(150)` for `waitForSelector("#cascade-form")` (robustness).
- Headless-Chrome E2E now **27/27** (14 cascade assertions).

### Accepted/named gap
- `runCascade`'s success path (real POST → SQL panel + INCOMPLETE banner) isn't driven by the E2E (the e2e console has no seeded cascade data). The banner logic was verified by the silent-failure read + the backend-contract grep; seeding a cascade fixture into `run.sh` is disproportionate for this harness. Highest-value future follow-up if a fixture becomes cheap.

No other changes. type-design N/A (no new types — frontend JS).

Issue #593, opened 2026-06-23
[pg] Baseline at slot consistent_point LSN → enable reconstruct/time-travel for PostgreSQL (GA)


nethalo:
## Summary

Wire a PostgreSQL **baseline producer** stamped at the slot's `consistent_point` LSN, so `reconstruct` / full-table time-travel works for PG sources. Today this is a documented "not yet"; `recover` (delta-only) is the supported PG recovery surface for beta. **GA gate** (epic #527 Beta-DEFERRABLE list).

## Current state (verified)

- **Producer absent.** `bintrail-pg` registers only `doctor` / `reset` / `stream` plus the shared read plane (`cmd/bintrail-pg/`); there is **no** baseline/dump/snapshot command. `internal/baseline/baseline.go` is mydumper-only — it embeds MySQL coords (`MetaKeyBinlogFile/Pos/GTIDSet`, `baseline.go:164-170`) with zero `pg_dump`/`COPY`/LSN awareness.
- **The LSN is captured but never used as a baseline anchor.** `slot.go:243` parses `res.ConsistentPoint` only as the streaming **start** LSN; `pgstreamrun` checkpoints the LSN to `stream_state` via the gtid-mode columns (`pgstreamrun.go:31-32,261`) — never stamped into a baseline.
- **Consumer present but fail-clean.** `reconstruct` IS registered for `bintrail-pg` (shared read plane, `internal/cli/commands.go:22`), but with no PG baseline it hits `FindBaseline → ErrNoBaseline` and fails **cleanly**, not silently wrong. Documented at `docs/postgres.md:319-322,405-409` ("Not yet for PostgreSQL: full-table reconstruct and the time-travel shim …").

## Work

- [ ] PG baseline producer: `pg_dump`/`COPY` → Parquet (reuse `internal/baseline.Writer`), taken consistently with the slot's `consistent_point` so deltas pick up exactly where the snapshot ends.
- [ ] Generalize the baseline metadata to carry an LSN anchor (today it is MySQL binlog-file/pos/GTID only).
- [ ] Align `reconstruct`/planner gap-checking to LSN coordinates for PG sources.
- [ ] Then un-gate full-table reconstruct + the `_snapshot` time-travel shim for PG and update `docs/postgres.md`.

Part of epic #527 (GA).

nethalo:
[...]

Issue #592, opened 2026-06-23
[pg] Harden recover/reconstruct to fail loud on a residual unchanged-TOAST marker


nethalo:
## Summary

Defense-in-depth: make the recover/reconstruct value path **fail loud** if it ever encounters a residual unchanged-TOAST sentinel, instead of silently JSON-marshalling it into output. This is the read-side belt-and-suspenders for the epic #527 TOAST gate.

## Context — the gate is already met upstream

Under the required `REPLICA IDENTITY FULL`, the decoder resolves an unchanged out-of-line TOASTed value (`'u'` after-image datum) from the before-image at **decode time** (Option B, `internal/pgcapture/decoder.go:444-455`), and RI-FULL is enforced on every live `RelationMessage` before any row event (`decoder.go:196-199`). So the real value lands in both images and the sentinel `UnchangedToastKey = "__bintrail_unchanged_toast__"` (`decoder.go:35`) is **never persisted** for a supported source — proven by `internal/pgstreamrun/pgstreamrun_integration_test.go:643-654` (zero sentinels under RI FULL) and `internal/recovery/recovery_test.go:228-258` (PG row round-trips with no sentinel in the predicate).

The only branch that could persist the marker (`decoder.go:455`) requires `role==roleAfter && before==nil && datum=='u'`, and `before==nil` happens only for INSERT, where `pgoutput` never emits `'u'`. **The marker is dead code for any RI-FULL source.**

## The gap

If a marker ever *did* slip through (a future bug, an unsupported capture path), `formatValuePG` (`internal/recovery/recovery.go:582-589`) **silently** `json.Marshal`s it into `{"__bintrail_unchanged_toast__":true}` — the opposite of fail-loud. `grep UnchangedToastKey` across `internal/recovery`, `internal/reconstruct`, `internal/query`, `internal/shim` returns nothing: nothing on the read side knows the sentinel exists.

## Acceptance

- [ ] `formatValuePG` (and any shared recover/reconstruct value formatter) detects a value carrying `UnchangedToastKey` and returns an **error** ("unresolved unchanged-TOAST marker — capture invariant violated") rather than marshalling it.
- [ ] Unit test feeding a synthetic row with the sentinel asserts a loud error, not silent output.

Low priority / not a beta blocker (the path is unreachable for supported sources), but cheap and honors the project's no-silent-wrong red line. Part of epic #527.

nethalo:
[...]

PR #596, opened 2026-06-23
feat(pg): pre-beta no-silent-loss guards + DDL-drift test coverage (#555 #556 #559 #591)

### LINKED ISSUE #555: [pg] Preflight: detect UNLOGGED tables in scope (silent capture loss — no WAL)
**Finding from the 2026-06-21 source-semantics discussion (verified vs PG primary sources).**

Logical decoding reads WAL. `UNLOGGED` tables **bypass WAL entirely** → `pgoutput` captures **nothing** for them — not even INSERTs. This is a **silent** capture-loss (invisible in a demo), the same class as the TOAST / binlog-fidelity silent-loss findings: the index simply has no events for an in-scope table, and a recovery "succeeds" while missing that table's data entirely.

## What to do
Fold into the #531 PG preflight validator: for in-scope (published) relations, detect `pg_class.relpersistence = 'u'` and **warn/refuse — never silently skip**. Surface it the same way the RI-FULL check does.

- Temp tables (`relpersistence = 't'`) share the property — lower priority (ephemeral, rarely published).
- This is preflight detection only; we can't make an UNLOGGED table recoverable (the data never hits WAL). The point is **visible, not silent**.

## Refs
- Extends #531 (per-table RI-FULL validator + PK-aware recovery)
- Epic #527
- Design record: `drafts/postgres-sequences-globals-extensions-2026-06-21.md`

### LINKED ISSUE #556: [pg] FK ON DELETE CASCADE/SET NULL is visible on PG (no synthesis) — validate cascade-child coverage
**Finding from the 2026-06-21 source-semantics discussion (verified TRUE vs PG primary sources).**

On PostgreSQL, FK `ON DELETE CASCADE` / `SET NULL` / `SET DEFAULT` is **visible in the logical stream** — the **opposite** of MySQL.

- PG docs ("Overview of Trigger Behavior"), verbatim: *"If a foreign key constraint specifies referential actions (that is, cascading updates or deletes), those actions are performed via ordinary SQL `UPDATE` or `DELETE` commands on the referencing table."* → real heap ops → real WAL → decoded as ordinary child-relation Delete/Update events.
- Corroboration (PG "Logical Replication Architecture"): the subscriber apply worker runs with `session_replication_role='replica'`, which **disables RI triggers** — yet children are still deleted on the subscriber, because the child deletes **arrive via the stream** as their own row changes.
- Contrast: MySQL ≤8.4 / MariaDB do **not** log cascade child rows (InnoDB-internal; Bug #32506; MySQL 8.0 manual: *"none of the changes are logged"*). That is why `internal/cascade` exists (synthesize victims from the FK graph + last `row_after`; spike PASSED byte-exact). **MySQL 9.6 (Jan 2026)** moved enforcement to the SQL layer → cascades logged (reversible via `innodb_native_foreign_keys`).

## Consequence
**Do NOT port `internal/cascade` (synthesize-from-FK-graph) to the PG path** — it is unnecessary in-window. Plain `recover` reverses cascade children directly from their before-images. Same for `SET NULL` (the child UPDATE with the old FK value is captured directly).

## What to do (preflight, extends #531)
For schemas with FK `ON DELETE CASCADE`/`SET NULL`, validate two hard preconditions:
1. **Cascade-child tables are in the publication** — else the parent DELETE is decoded but the cascaded child deletes are **silently absent** (partial visibility).
2. **Child has `REPLICA IDENTITY FULL`** — else only the PK is in the before-image (incomplete reversal).

If either is missing → warn (partial cascade visibility); the residual is the usual "row never in our history" gap, covered by baseline. Document the PG-vs-MySQL contrast in SUPPORT.md / the cascade draft.

## Refs
- Extends #531; epic #527
- `drafts/cascade-recovery-port-2026-06-21.md`, `drafts/postgres-sequences-globals-extensions-2026-06-21.md`

### LINKED ISSUE #559: [pg] Out-of-scope guard: detect TimescaleDB hypertable chunks (decode-granularity mismatch)
**Finding from the 2026-06-21 source-semantics discussion (verified vs TimescaleDB/PG sources).**

TimescaleDB hypertables = a parent abstraction physically split into dynamically-named child **chunk** tables (`_timescaledb_internal._hyper_1_42_chunk`). Logical decoding / publications surface INSERT/UPDATE/DELETE at the **chunk** relation level, **not** the hypertable. Consequences:

- bintrail would index rows under internal chunk names (which change over time; compression complicates further) → `recover`/`reconstruct` would target chunk relations → fragile / wrong.
- `publish_via_partition_root` does **not** reliably abstract Timescale chunks (they are inheritance-based, not native declarative partitions).
- Columnar / compressed chunks may not decode as ordinary row changes at all.

This is a known-hard CDC case (Debezium ships a dedicated SMT to reroute chunk events back to the hypertable).

## What to do (alpha)
**Out of scope.** Detect-and-**warn-once** if an in-scope/published relation is a `_timescaledb_internal` chunk → "TimescaleDB hypertables are out of scope." **Never silently index under chunk names** (silent-wrong-target). A faithful hypertable mapping is a deferred investigation, not an alpha dependency.

## Refs
- Epic #527
- `drafts/postgres-sequences-globals-extensions-2026-06-21.md`

### LINKED ISSUE #591: [pg] Test coverage: mid-stream ALTER (DDL drift) re-snapshot — correct-by-construction but untested
## Summary

DDL-drift handling on the PostgreSQL path (mid-stream `ALTER TABLE`) is **correct-by-construction but has zero test coverage**. This is the verification half of the epic #527 beta-BLOCKING gate *"DDL drift detection (Relation-message-driven re-snapshot)"* — the behavior is in place; what's missing is a test that pins it before beta sign-off.

## Why the behavior is already correct

`pgoutput` re-emits a `RelationMessage` before the first changed row of a relation **and again after its shape changes**. The decoder consumes that:

- `cacheRelation` overwrites `d.relations[m.RelationID]` with the new column set — `internal/pgcapture/decoder.go:264-269`. This is a **synchronous** shape swap (structurally stronger than the MySQL `#396` re-snapshot, which needed an out-of-band `information_schema` re-read).
- The consumer re-persists on every `EventRelation`: `internal/pgstreamrun/pgstreamrun.go:303-317` calls `metadata.WritePGSnapshot(...)` and re-stamps `snapshotByTable[schema.table]`.
- `metadata.WritePGSnapshot` (`internal/metadata/metadata.go:371-422`) allocates a **fresh** `snapshot_id` (`MAX(snapshot_id)+1`) with the current `RelationMessage`'s columns, so a changed shape yields a new shape-correct `snapshot_id` and subsequent rows are stamped with it.
- Fail-loud backstops cover the rest: PK-column drift → error (`decoder.go:238-243`); tuple column-count mismatch → error (`decoder.go:432-435`); non-FULL replica identity on a mid-stream new relation → error (`decoder.go:196-199`).

Net: a mid-stream `ALTER` does **not** silently shift row shape.

## The gap

Every `ALTER` in `internal/pgcapture` and `internal/pgstreamrun` integration tests is **setup-time** (`SET STORAGE EXTERNAL` / `SET REPLICA IDENTITY FULL`). No decoder unit test re-emits the same OID with a different column set, and no integration test alters a table mid-stream. The alpha PRs (#545/#546/#547/#554) delivered decoder + capturer + consumer but no drift test.

## Acceptance

- [ ] Decoder unit test: feed `relMsg(oid,"id","a")` then `relMsg(oid,"id","a","b")` then a row, assert the row decodes against the new (2-data-column) shape.
- [ ] Integration test: `ALTER TABLE ... ADD COLUMN` mid-stream, assert the post-ALTER row gets a **different** `snapshot_id` than the pre-ALTER row **and** recovers correctly (PK-scoped WHERE on the new shape).
- [ ] (Highest-value scenario is the additive non-PK `ADD COLUMN`; DROP/RENAME-into-different-PK and column-count-shrink are already covered by the fail-loud backstops above.)

Part of epic #527.

nethalo:
## Why

The epic #527 close-out audit found the **6 beta-BLOCKING gates all met**, with three small **silent-loss / silent-target** residuals and **one untested** beta gate left before declaring PostgreSQL-as-source beta. This is the pre-close lot. None of these change existing public signatures.

| # | Gap | Class | Guard added |
|---|-----|-------|-------------|
| #555 | UNLOGGED tables write no WAL → captured as nothing | silent **capture** loss | preflight WARN (startup + doctor) |
| #556 | published parent + unpublished cascade child → cascade deletes uncaptured | silent **cascade** loss | preflight WARN (startup + doctor) |
| #559 | TimescaleDB chunks streamed by physical name | silent **wrong target** | decoder warn-once |
| #591 | Relation-message shape-swap correct-by-construction but untested | **verification** gap | unit + integration test |

## What

**#555 — UNLOGGED preflight.** `listUnloggedPublishedTables` (slot.go) finds `relpersistence='u'` tables in the publication. A non-fatal `WARN` at capture startup (`warnCaptureCoverage`) + a doctor check. WARN, not FAIL — an operator may keep ephemeral data UNLOGGED on purpose; the point is it's *visible*, not discovered during a failed recovery.

**#556 — FK cascade-child coverage.** `listUncoveredCascadeChildren` joins `pg_constraint` (`confdeltype IN ('c','n','d')`) to find children whose parent is published but the child isn't. On PG, `ON DELETE CASCADE/SET NULL` is real per-row WAL on the child, so `recover` works *if* the child is captured — this surfaces the gap. `FOR ALL TABLES` covers children by construction; a published child is already forced to RI FULL by the existing validator, so only the not-published case is flagged. Residual (the "row predates capture" baseline class) is documented and tracked at #593 (GA).

**#559 — TimescaleDB out-of-scope guard.** `decoder.cacheRelation` warns **once per stream** on a `_timescaledb_internal._hyper_*` chunk relation, so an operator knows capture targets the raw chunk, not the logical hypertable. Warn-once (a busy hypertable has many chunks); never silently skipped.

**#591 — DDL-drift test coverage.** The mechanism (pgoutput re-emits a `RelationMessage` after a shape change → `cacheRelation` swaps; the consumer re-snapshots with a fresh `schema_version`) was correct but untested.
- **Unit (local):** `TestDecode_RelationShapeChange_MidStream` — discriminating: a 3-column tuple decodes only if the shape swapped; against the stale 2-column relation `decodeTuple` would error (column-count mismatch), or drop the column (caught by the assertions).
- **Integration (CI-only):** `TestOne_DDLDrift_MidStreamAlter` — drives a real `ALTER TABLE ... ADD COLUMN` mid-stream, asserts the added column is captured, the post-ALTER row gets a **different** `schema_version`, and recovery still produces valid SQL.

The doctor gains two WARN checks (`No UNLOGGED tables`, `FK cascade-child coverage`) with pure, unit-tested mappers (`unloggedResult`/`cascadeCoverageResult`), added to `dependentCheckNames` so they SKIP cleanly when `wal_level != logical`.

## Deferred (not in this PR)

**#592** (TOAST read-side fail-loud) is intentionally left as a follow-up: the marker path is dead code under the required RI FULL, and the current behavior is *visible-not-silent* (renders a distinct literal, not a silent corruption). Making it *error* would require threading `formatValue → (string, error)` across `recovery.go` for unreachable code — wrong trade. See the issue.

## Testing

- `go build ./...`, `go vet`, and the full unit suites for `internal/pgcapture`, `internal/pgstreamrun`, `internal/recovery` pass locally.
- New unit tests (`TestDecode_RelationShapeChange_MidStream`, `TestDecode_TimescaleChunk_WarnsOnce`, `TestUnloggedResult`, `TestCascadeCoverageResult`) pass locally.
- **`TestOne_DDLDrift_MidStreamAlter` is CI-only** (needs a live PostgreSQL via `BINTRAIL_TEST_PG_DSN`); a green local `go test` does **not** exercise it — the PG CI matrix (PG 14–17) does.

closes #555
closes #556
closes #559
closes #591

### DIFF
--- a/internal/pgcapture/capturer.go
+++ b/internal/pgcapture/capturer.go
@@ -195,15 +195,20 @@ func (c *Capturer) Run(ctx context.Context, out chan<- event.Event) error {
 // logged at debug and never blocks the stream — these are advisory, and the same
 // catalog state is also reported by `bintrail-pg doctor`.
 func (c *Capturer) warnCaptureCoverage(ctx context.Context, conn *pgx.Conn) {
-	if unlogged, err := listUnloggedPublishedTables(ctx, conn, c.cfg.Publication); err != nil {
-		c.logger.Debug("pgcapture: could not check for UNLOGGED tables", "error", err)
+	// A probe FAILURE is itself surfaced as a WARN, not swallowed at Debug: at the
+	// default --log-level=info a Debug line is invisible, so downgrading the failure
+	// would silently disable the very guard this function exists to provide — the exact
+	// silent-failure class #555/#556 close. Visibility matches the doctor side, which
+	// also maps a probe failure to a WARN.
+	if unlogged, err := listUnloggedCaptureTables(ctx, conn, c.cfg.Publication, c.cfg.Filters); err != nil {
+		c.logger.Warn("pgcapture: UNLOGGED-table coverage check failed to run — could NOT determine whether a captured table writes no WAL (a silent recovery hole); run `bintrail-pg doctor` to re-check", "error", err)
 	} else if len(unlogged) > 0 {
-		c.logger.Warn("pgcapture: UNLOGGED table(s) in the publication produce no WAL — their changes are NOT captured and cannot be recovered (ALTER TABLE <t> SET LOGGED if you need them)",
+		c.logger.Warn("pgcapture: UNLOGGED table(s) in capture scope produce no WAL — their changes are NOT captured and cannot be recovered (ALTER TABLE <t> SET LOGGED if you need them)",
 			"tables", strings.Join(unlogged, ", "))
 	}
 
 	if children, err := listUncoveredCascadeChildren(ctx, conn, c.cfg.Publication); err != nil {
-		c.logger.Debug("pgcapture: could not check FK cascade-child coverage", "error", err)
+		c.logger.Warn("pgcapture: FK cascade-child coverage check failed to run — could NOT determine whether a cascade child is unpublished (cascade deletes would be silently lost); run `bintrail-pg doctor` to re-check", "error", err)
 	} else {
 		for _, ch := range children {
 			c.logger.Warn("pgcapture: FK cascade child is NOT in the publication — cascade deletes from its parent are not captured and cannot be recovered (add it to the publication)",
--- a/internal/pgcapture/capturer.go
+++ b/internal/pgcapture/capturer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -137,6 +138,13 @@ func (c *Capturer) Run(ctx context.Context, out chan<- event.Event) error {
 		return err
 	}
 
+	// Non-fatal capture-coverage warnings (#555 UNLOGGED, #556 FK cascade child): tables
+	// whose changes will be silently ABSENT from the index. Unlike the validate* gates
+	// above, these do NOT abort capture (an operator may keep UNLOGGED data on purpose,
+	// and a missing cascade child may be intentionally out of recovery scope) — but they
+	// must be visible now, not discovered during a failed recovery.
+	c.warnCaptureCoverage(startupCtx, queryConn)
+
 	startLSN, err := ensureSlot(startupCtx, replConn, queryConn, c.cfg.SlotName, c.cfg.StartLSN, c.cfg.ExpectExistingSlot)
 	if err != nil {
 		return err
@@ -180,6 +188,28 @@ func (c *Capturer) Run(ctx context.Context, out chan<- event.Event) error {
 		return nil // graceful shutdown (ctx-cancel surfaced as a receive/emit error)
 	}
 	return err
+}
+
+// warnCaptureCoverage logs the non-fatal coverage gaps (#555 UNLOGGED tables, #556 FK
+// cascade children whose parent is published but the child is not). A probe FAILURE is
+// logged at debug and never blocks the stream — these are advisory, and the same
+// catalog state is also reported by `bintrail-pg doctor`.
+func (c *Capturer) warnCaptureCoverage(ctx context.Context, conn *pgx.Conn) {
+	if unlogged, err := listUnloggedPublishedTables(ctx, conn, c.cfg.Publication); err != nil {
+		c.logger.Debug("pgcapture: could not check for UNLOGGED tables", "error", err)
+	} else if len(unlogged) > 0 {
+		c.logger.Warn("pgcapture: UNLOGGED table(s) in the publication produce no WAL — their changes are NOT captured and cannot be recovered (ALTER TABLE <t> SET LOGGED if you need them)",
+			"tables", strings.Join(unlogged, ", "))
+	}
+
+	if children, err := listUncoveredCascadeChildren(ctx, conn, c.cfg.Publication); err != nil {
+		c.logger.Debug("pgcapture: could not check FK cascade-child coverage", "error", err)
+	} else {
+		for _, ch := range children {
+			c.logger.Warn("pgcapture: FK cascade child is NOT in the publication — cascade deletes from its parent are not captured and cannot be recovered (add it to the publication)",
+				"child", ch.Child, "parent", ch.Parent, "action", ch.Action)
+		}
+	}
 }
 
 // receiveLoop is the single-goroutine replication loop. pgconn is NOT concurrency-
--- a/internal/pgcapture/decoder.go
+++ b/internal/pgcapture/decoder.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log/slog"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/jackc/pglogrepl"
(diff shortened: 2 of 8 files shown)


nethalo:
## Multi-agent review applied (3 agents)

Ran code-reviewer + silent-failure-hunter + pr-test-analyzer. Findings and fixes (commit `d2664b2`):

**CRITICAL — #555 guard was dead code** (code-reviewer, verified empirically vs live PG 16.14). It queried `pg_publication_tables`, but an UNLOGGED table is **never** there — PG rejects `FOR TABLE u` at DDL time and excludes UNLOGGED from `FOR ALL TABLES` via `pg_relation_is_publishable`. The query could never match a row. **Fixed:** `listUnloggedCaptureTables` now scans `pg_class` directly for the `FOR ALL TABLES` case (the only case where an UNLOGGED table is silently uncaptured), narrowed by the client filter; `FOR TABLE` is already covered by `validatePublication`.

**HIGH — silent probe failure** (silent-failure-hunter). `warnCaptureCoverage` logged catalog-probe failures at `Debug`, invisible at the default `--log-level=info` — silently disabling the very guard this PR adds. **Fixed:** both promoted to `Warn`, matching the doctor side.

**Test gaps** (pr-test-analyzer). The catalog queries themselves were covered only by pure-mapper unit tests. **Fixed:** added `TestBuildPGReport_CoverageWarnings_Integration` (live-PG, exercises both #555 and #556 SQL + `cascadeAction`) and `TestDecode_TimescaleChunk_WarnsBeforeReplicaIdentityAbort` (locks the #559 warn-before-RI-abort ordering).

**Reviewed clean (no change):** the #556 cascade query (FK direction + `confdeltype` verified against live PG), the warn-once ordering, the doctor WARN-on-probe-failure semantics. The two flagship unit tests (`TestDecode_RelationShapeChange_MidStream`, `TestDecode_TimescaleChunk_WarnsOnce`) were **mutation-proven discriminating** — they fail under the exact regressions #591/#559 guard against.

PR #615, opened 2026-06-25
feat(console): opt-in "Create baseline" button — in-process dump→convert→upload (#613)

### LINKED ISSUE #613: Trigger a baseline snapshot from the console UI (control-plane)
## Goal

Let an operator trigger a **baseline snapshot** for a monitored server from the console UI (`watch`/control-plane), instead of only via the `docker compose --profile baseline` one-shot. A baseline enables Time-travel reconstruct from the console; today the UI can only *list* existing baselines (`GET /api/baselines`), not create one.

## Current state
- Console is read-mostly; the only writes are the server registry YAML + control-plane endpoints (`POST /api/servers/{id}/monitor/start|stop`, `POST /api/recover`, `POST /api/recover-cascade`).
- The baseline pipeline is **mydumper** (consistent dump + binlog coords in its metadata) → `bintrail baseline` (Parquet convert with the embedded coords). Neither mydumper nor the `dump`/`baseline` subcommands live in the `bintrail-console` image (debian-slim + the console binary only). The daemon shells out to nothing today.

## Decision (2026-06-24)
- **Not** blocking a release; tracked as a dedicated follow-up.
- Preferred approach: **reuse the pinned mydumper docker image** (`mydumper/mydumper:v1.0.3-1`) — `bintrail dump` already supports `--mydumper-image` (it `docker run`s mydumper). The daemon triggers the dump→convert pipeline via the **docker socket** (`/var/run/docker.sock` mounted into the console container). Rejected alternatives: bundling mydumper into the console image (bloat + console becomes an active source operator); SQL-in-daemon snapshot (cleanest long-term but reimplements consistent multi-table snapshot + binlog-coord capture — delicate fidelity).

## Scope sketch
- `POST /api/servers/{id}/baseline` (control-plane, `watch` only; gated like monitor start/stop), runs the dump→convert against that server's source DSN, writes to the per-server baseline dir/state volume, returns progress/result.
- Reuse the system-schema exclusion default just shipped (PR #612) for the dump.
- Frontend: a "Create baseline" action on the Storage/Baselines panel; surface running/last-result status.
- **Security**: mounting the docker socket = host-root-equivalent for the console container. Acceptable for a loopback console; must be opt-in / documented, and the socket mount added to compose only for this capability. Consider gating behind an explicit flag.

## Open questions
- Docker-socket security gate / opt-in design.
- Where the dump runs (host docker vs the daemon's own container) and how it reaches the source network + state volume.
- Concurrency / one-baseline-at-a-time per server; surfacing the GUARD (zero-table/partial-dump) failures in the UI.

nethalo:
closes #613

## Summary
Adds an **opt-in "Create baseline" button** to the console's Storage → Baselines panel, so an operator can produce a baseline snapshot for a monitored server from the UI instead of only via `docker compose --profile baseline` or the CLI.

The key design choice (see issue thread): the console **never mounts the docker socket**. It runs the whole pipeline **in-process** — the `bintrail-console` image bundles `mydumper`, the watch daemon runs it as a **local subprocess**, then converts to Parquet and uploads to S3 by calling `internal/baseline` directly. A console compromise can therefore never escalate to host-root, and the source DSN never leaves the process (never written to disk or any HTTP response).

## What changed
- **`internal/baseline.Upload`** — extracted the crash-safe S3 upload (the `_INCOMPLETE`-first → data → `_SUCCESS`-last ordering, #467/#524) from `cmd/bintrail` so the CLI and the console share one implementation. Its unit tests moved with it.
- **`internal/console`** — `BaselineController` interface (wired by `watch`, mirroring `MonitorController`); `POST`/`GET /api/servers/{id}/baseline`; `baseline_trigger` capability. Gating, in order: control-plane only (403 on read-only `serve`), opt-in (403 unless enabled), source DSN configured (400), baseline destination configured (400), one at a time per server (409).
- **`cmd/bintrail-console`** — `baselineSupervisor` runs `mydumper → baseline.Run → baseline.Upload` in a goroutine with in-memory status. For an `s3://` destination the dump is staged in a fresh temp dir, uploaded, then discarded (so a re-run never re-uploads the local backlog). Opt in with `BINTRAIL_CONSOLE_BASELINE_TRIGGER=1`.
- **Frontend** — "Create baseline" button on the Baselines panel (gated by the capability + a configured destination), polls status, toasts the result. The manual CLI/compose recipe stays as the always-available path.
- **Packaging/docs** — both console Dockerfiles bundle `mydumper`; `docker-compose.yml` passes `BASELINE_TRIGGER` through (+ a staging dir on the state volume); `docs/console.md` documents the opt-in and the security rationale.

## Security notes
- `mydumper` runs as a **local subprocess** — no docker socket, no host-root surface on the network-facing console.
- Source DSN (with password) stays in-process; the trigger response carries only non-secret status. A unit test asserts the password never appears in the HTTP response.
- Off by default; the button is hidden on `serve` and whenever the opt-in is off.

## Test plan
- [x] `go test ./... -count=1` (unit) — new tests: endpoint gating (403/400/404/409/202), capability gate, no-secret-leak, status read-back, `mydumper` arg-building, and the relocated upload-ordering tests.
- [ ] **Build-verify the image** (`docker build -f Dockerfile.bintrail-console .`) — confirms `apt-get install mydumper` resolves on `bookworm-slim` (CI builds the image; flagged because it couldn't be built in the dev sandbox).
- [ ] Manual: on the EC2 demo, `BASELINE_TRIGGER=1`, click Create baseline for `wp`, confirm a new snapshot lands in S3 and appears in the panel.

### DIFF
--- a/Dockerfile.bintrail-console
+++ b/Dockerfile.bintrail-console
@@ -29,15 +29,22 @@ RUN CGO_ENABLED=1 go build \
 # ── Stage 2: runtime ───────────────────────────────────────
 # Debian slim for glibc compatibility with DuckDB.
 FROM debian:bookworm-slim
+ARG TARGETARCH
 
+# mydumper powers the OPT-IN in-process baseline trigger (#613): with
+# BINTRAIL_CONSOLE_BASELINE_TRIGGER=1 the watch daemon runs mydumper here as a
+# LOCAL subprocess (no docker socket), then converts+uploads in-process. We pin
+# the SAME version the compose baseline-dump pipeline uses (mydumper v1.0.3-1) so
+# console-created baselines are identical to CLI/compose ones — bookworm's apt
+# mydumper is 0.10, too old for the CONSISTENT lock-free --trx-tables snapshot a
+# least-privilege replication user (no RELOAD/FLUSH_TABLES) needs (verified against
+# a real Percona 8.0 source). Idle bytes when the trigger is not enabled.
 RUN apt-get update && \
-    # mydumper powers the OPT-IN in-process baseline trigger (#613): with
-    # BINTRAIL_CONSOLE_BASELINE_TRIGGER=1 the watch daemon runs mydumper here as
-    # a LOCAL subprocess (no docker socket), then converts+uploads in-process.
-    # bookworm ships mydumper < 0.18, so the console's runner omits the
-    # lighter-locking flags (a baseline holds a brief read lock — acceptable for
-    # an occasional snapshot). Idle bytes when the trigger is not enabled.
-    apt-get install -y --no-install-recommends ca-certificates mydumper && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates wget libatomic1 libglib2.0-0 libpcre2-8-0 zstd && \
+    wget -q "https://github.com/mydumper/mydumper/releases/download/v1.0.3-1/mydumper_1.0.3-1.bookworm_${TARGETARCH}.deb" -O /tmp/mydumper.deb && \
+    dpkg -i /tmp/mydumper.deb && \
+    rm -f /tmp/mydumper.deb && \
     rm -rf /var/lib/apt/lists/* && \
     # Pin uid 999: the bundled compose chowns the index password secret to
     # uid 999 (mode 0600) for this process to read — see docker-compose.yml.
--- a/Dockerfile.bintrail-console
+++ b/Dockerfile.bintrail-console
@@ -31,7 +31,13 @@ RUN CGO_ENABLED=1 go build \
 FROM debian:bookworm-slim
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends ca-certificates && \
+    # mydumper powers the OPT-IN in-process baseline trigger (#613): with
+    # BINTRAIL_CONSOLE_BASELINE_TRIGGER=1 the watch daemon runs mydumper here as
+    # a LOCAL subprocess (no docker socket), then converts+uploads in-process.
+    # bookworm ships mydumper < 0.18, so the console's runner omits the
+    # lighter-locking flags (a baseline holds a brief read lock — acceptable for
+    # an occasional snapshot). Idle bytes when the trigger is not enabled.
+    apt-get install -y --no-install-recommends ca-certificates mydumper && \
     rm -rf /var/lib/apt/lists/* && \
     # Pin uid 999: the bundled compose chowns the index password secret to
     # uid 999 (mode 0600) for this process to read — see docker-compose.yml.
--- a/Dockerfile.bintrail-console.goreleaser
+++ b/Dockerfile.bintrail-console.goreleaser
@@ -11,7 +11,13 @@
 FROM debian:bookworm-slim
 
 RUN apt-get update && \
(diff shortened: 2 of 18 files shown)


nethalo:
## Build verification (local Docker) ✅

Both deferred test-plan items are now confirmed:

- **`apt-get install mydumper` resolves on `bookworm-slim`** — verified in a throwaway container. Version is **mydumper 0.10.0** (built against the MariaDB 10.11 client that bookworm packages), which is `< 0.18` — confirming the console runner correctly **omits** the `--sync-thread-lock-mode`/`--trx-tables` flags (passing them to a pre-0.18 build fails with "unknown option").
- **Full image build** (`docker build -f Dockerfile.bintrail-console .`) exits 0. Smoke test on the result: `mydumper --version` runs as the bundled uid-999 binary, `bintrail-console watch --help` works. Image size **252 MB**.

One note for first real-source run: the bookworm `mydumper` is linked against `libmariadb3` (not the MySQL client). bookworm's `libmariadb3` (3.3.x) supports `caching_sha2_password`, so it should authenticate against MySQL/Percona 8.0 sources fine — worth a sanity check on the first trigger against a real source, since the previous compose flow used the MySQL-linked `mydumper/mydumper` image.

nethalo:
## End-to-end validation on a real EC2 (arm64 / Percona 8.0) ✅

Deployed this branch's console image to a live EC2 demo (Graviton/arm64, source = Percona Server 8.0, least-privilege replication user) and clicked **Create baseline** for a real monitored server. Final result:

```
baseline: snapshot complete  server=wp  tables=47  rows=33277  uploaded=48
```

A new snapshot landed in S3 (`s3://…/baselines/2026-06-25T19-09-57Z/`) with 47 Parquet files + `_SUCCESS` (no `_INCOMPLETE` — clean upload) and appeared in the console's Baselines panel. The whole pipeline ran in-process — `mydumper` as a local subprocess, **no docker socket**.

**The real-path test caught three bugs that unit tests and two inspection-based review rounds could not** (all now fixed in this PR):

1. **Schema-only dump for least-privilege users** — the image installed bookworm's apt `mydumper` (0.10), too old for the consistent lock-free `--trx-tables` snapshot. Without it mydumper attempts a global FTWRL, which a replication user lacking `RELOAD`/`FLUSH_TABLES` can't acquire → only the schema dumped, no data. Fixed by pinning `mydumper v1.0.3-1` via the upstream `.deb` (the same version the compose pipeline uses) and passing `--sync-thread-lock-mode NO_LOCK --trx-tables`.
2. **Unfiltered dump dies on system schemas** — a least-priv user can't read the `sys` views; with no schema filter the dump hit them. Fixed by excluding the system schemas with the same negative-lookahead PCRE as the compose pipeline (#612).
3. **S3 upload failed with no AWS_REGION** — `NewS3Client(ctx, "")` relied on the SDK resolving region from the chain, but AWS SDK Go v2 does NOT fall back to EC2/ECS IMDS for region. In an IAM-role-only deployment (no `AWS_REGION`), region was empty → `region was not a valid DNS name`. Fixed with an IMDS region fallback in `NewS3Client`.

The EC2 was reverted to the published `latest` image afterward; the demo is untouched.

PR #656, opened 2026-06-29
config+compose: raise index max_allowed_packet ceiling (stopgap for #652)

### LINKED ISSUE #652: capture: oversized BLOB/JSON row events (>~48 MB raw) are silently dropped — data loss
## Summary

Row events whose before/after image exceeds ~48 MB of raw binary BLOB (or ~64 MB of JSON text) **fail to index and are silently lost**. The row stays in the source but never lands in `binlog_events`. This is silent data loss with no non-zero exit and no surfaced "gap".

Reproduced on the EC2 demo host with a freshly built **v0.23.0** arm64 binary against **Percona Server 8.0.45** (`binlog_row_image=FULL`, `max_allowed_packet=67108864` = the 64 MB default, `STRICT_TRANS_TABLES`).

## Root cause

Two factors combine, both in the capture path:

1. **base64 inflation.** Binary BLOBs are not valid JSON, so `marshalRow` (`internal/indexer/indexer.go`) falls through to `json.Marshal`, which base64-encodes them (~1.333× inflation) before they go into the JSON `row_before`/`row_after` columns of `binlog_events`.
2. **`max_allowed_packet` is never tuned.** `config.Connect` (`internal/config/config.go`) sets only `parseTime` + a 10 s timeout — it never raises `maxAllowedPacket` on the index connection, so the go-sql-driver / server 64 MB default governs the INSERT packet.

Net effective ceiling per blob ≈ 64 MB ÷ 1.333 ≈ **48 MB raw**, far below the 64 MB an operator would expect.

## Reproduced evidence

Pinned threshold (offline `index`, `--batch-size 1`, fresh binlog so events are isolated):

| row | raw blob | stored as `row_after` | result |
|---|---|---|---|
| id=3 | 40 MB (41,943,040 B) | **55,924,074 B** (= ×1.333, base64 confirmed) | indexed OK (< 64 MB) |
| id=4 | 50 MB (52,428,800 B) | base64 ≈ 66.7 MB > 64 MB | **INSERT fails, event lost** |

The failing INSERT:

```
Error 1105 (HY000): Parameter of prepared statement which is set through
mysql_send_long_data() is longer than 'max_allowed_packet' bytes
```

Reconciliation after the run: **source = 2 rows (id 3,4), index = 1 event.** id=4 is gone.

## Failure modes (both observed / one inferred)

- **`stream` (live capture):** the failing flush blocks checkpoint persistence, so `stream_state` is never written. On restart the stream **auto-discovers the current binlog position** (past the failed event) and **silently skips it forever** — observed: `source=2, index=1`. When a checkpoint already exists before the oversized event, the symmetric outcome is a **wedge/crashloop** (the same event re-fails every restart, no progress) — this variant is *inferred from the checkpoint-after-successful-flush ordering* in `internal/streamrun/streamrun.go`, not directly observed here.
- **offline `index`:** logs `level=ERROR msg="indexing failed" ...max_allowed_packet...` but then logs `level=INFO msg="indexing complete" events_indexed=1` and **exits 0**. A CI/cron wrapper reads this as success while a row was silently dropped. This exit-0-on-data-loss is the nastiest property.

## Impact

Silent, unrecoverable data loss for any table with large LOB/JSON values (media blobs, serialized documents, large JSON aggregates). The operator gets no error on the happy path and no "no data lost" signal that this happened.

## Proposed direction (not prescriptive)

- Raise/negotiate `maxAllowedPacket` on the index DSN, and/or split batches and retry oversized single events at a higher packet limit.
- Detect an oversized event before the INSERT and **fail loud / surface a gap** instead of dropping it — never let the offline `index` path report `complete` + exit 0 when an event was dropped.
- Consider hex/raw storage instead of base64 to remove the 1.333× penalty (see the sibling recover-corruption issue).

## Environment
v0.23.0 (commit 4cd010d) · Percona Server 8.0.45 aarch64 · max_allowed_packet 64 MB default · reproduced via SSM on the demo EC2 host 2026-06-28. Full audit notes in `drafts/large-data-robustness-audit-2026-06-28.md`.

nethalo:
## What

Quick-win **step 0** from the large-data robustness grooming. Raises the index `max_allowed_packet` ceiling so large BLOB/JSON row images can be written, and makes the client honor it.

**Two legs, both required together:**
1. `docker-compose.yml` index-mysql → `--max-allowed-packet=1G` (the MySQL maximum), alongside the existing `--sort-buffer-size=4M` (SHIP carve-out: tuning the bundled index MySQL).
2. `internal/config/config.go` `Connect` → the go-sql-driver caps the **client** at 64 MiB by default even when the server allows more. Setting `maxAllowedPacket=0` makes the driver fetch `@@max_allowed_packet` from the server and honor it. Extracted a unit-testable `buildDSN` seam; an explicit `maxAllowedPacket` in the user's DSN is preserved (precedence).

## Why

`binlog_events` stores full before/after row images as JSON. A binary BLOB is base64-inflated ~1.33×, so a ~48 MB blob already crosses the 64 MiB packet limit and the index INSERT is rejected — today that event is **silently dropped** (reproduced on EC2: source=2 rows / index=1).

## Scope — this is a ceiling raise, NOT the fix

Partial mitigation for #652. It does **not** close #652:
- Events larger than the server's `max_allowed_packet` are still rejected — the silent-drop must be made **fail-loud** (#652 ROOT-2).
- BYO indexes still at 64 M are unaffected until they raise their own limit (validate-warn is #652 ROOT-1).

Deliberately **not** changing the storage encoding (the grooming rejected re-encode: worse ceiling with hex, crosses the one-index-schema red line, O(every-row) irreversible migration).

## Tests

`internal/config` unit tests: `maxAllowedPacket=0` is emitted (honor server), an explicit DSN value wins, parseTime/Loc invariants survive, invalid DSN errors. All green.

Refs #652

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -120,9 +120,18 @@ services:
     # by sorting narrow keys only (internal/query late materialization); 4M is
     # defense-in-depth for any other sort, bounded and cheap at this index's low
     # connection concurrency.
+    #
+    # --max-allowed-packet=1G raises the 64M stock default. A single INSERT into
+    # binlog_events carries the full before/after row images as JSON; a large
+    # BLOB/JSON value is base64-inflated ~1.33× and can exceed 64M, which the
+    # server would reject (#652). 1G is the MySQL maximum. bintrail's client
+    # honors this via config.Connect (maxAllowedPacket=0 → fetched from server).
+    # This is a ceiling raise, not a fix: events above 1G are still rejected and
+    # must fail loud rather than drop silently (see #652).
     command:
       - "--innodb-flush-log-at-trx-commit=2"
       - "--sort-buffer-size=4M"
+      - "--max-allowed-packet=1G"
     environment:
       # 8.4 defaults to caching_sha2_password (native_password disabled);
       # bintrail's driver negotiates it over the plaintext Docker network.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -121,13 +121,14 @@ services:
     # defense-in-depth for any other sort, bounded and cheap at this index's low
     # connection concurrency.
     #
-    # --max-allowed-packet=1G raises the 64M stock default. A single INSERT into
-    # binlog_events carries the full before/after row images as JSON; a large
+    # --max-allowed-packet=1G raises the 64M stock default. Each INSERT into
+    # binlog_events carries full before/after row images as JSON; a single large
     # BLOB/JSON value is base64-inflated ~1.33× and can exceed 64M, which the
-    # server would reject (#652). 1G is the MySQL maximum. bintrail's client
-    # honors this via config.Connect (maxAllowedPacket=0 → fetched from server).
-    # This is a ceiling raise, not a fix: events above 1G are still rejected and
-    # must fail loud rather than drop silently (see #652).
+    # server rejects (the reproduced Error 1105 in #652). 1G is the MySQL
+    # maximum. bintrail's client tracks this via config.Connect
+    # (maxAllowedPacket=0 → fetched from the server). This is a ceiling raise,
+    # not a fix: events above 1G are still rejected and must — but do not yet
+    # (#652) — fail loud rather than drop silently.
     command:
       - "--innodb-flush-log-at-trx-commit=2"
       - "--sort-buffer-size=4M"
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -114,11 +114,17 @@ func Connect(dsn string) (*sql.DB, error) {
 	return db, nil
 }
 
+// driverDefaultMaxAllowedPacket is the client-side max_allowed_packet the
+// go-sql-driver assigns when a DSN omits the parameter. Read from the driver
+// itself (via NewConfig) so it tracks any change to that default instead of
+// hardcoding 64 MiB.
+var driverDefaultMaxAllowedPacket = mysql.NewConfig().MaxAllowedPacket
+
 // buildDSN applies bintrail's connection invariants to a user DSN and returns
 // the normalized DSN string. Split out from Connect so the invariants are unit
 // testable without a live server. Invariants: parseTime=true (DATETIME scans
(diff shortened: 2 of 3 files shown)


nethalo:
## Self-review (4 agents) + fixes — `0525b92`

Ran code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer. No critical/important blockers; the refactor is behavior-preserving and `maxAllowedPacket=0` is the correct driver mechanism. Three things addressed:

1. **Framing correction (important).** The comment-analyzer verified that for bintrail's actual write path (prepared statements → `COM_STMT_SEND_LONG_DATA`, chunked at `maxAllowedPacket`), the client cap doesn't reject large params locally — the reproduced rejection in #652 is **server-side** (`Error 1105 … mysql_send_long_data() … 'max_allowed_packet'`). So the **load-bearing change is the compose `--max-allowed-packet=1G`**; the client `maxAllowedPacket=0` is alignment (keep it — harmless and correct, and it honors a *lower* BYO server limit too). Corrected the inaccurate "client rejects locally" comment accordingly. The original "two legs, both required together" wording in this PR description is the imprecise bit — the server leg is what stops the loss.

2. **Precedence hardening (medium silent-failure).** The original substring guard (`strings.Contains(ToLower(dsn), "maxallowedpacket")`) could false-positive on a credential/dbname, and was case-insensitive while the driver is case-sensitive — both silently re-capping the client at 64 MiB. Replaced with a comparison against the driver's own default (`mysql.NewConfig().MaxAllowedPacket`): no string-matching, so a mis-cased or credential-embedded occurrence falls through to the safe server-honoring `0`.

3. **Test quality.** Dropped a vacuous `Loc=UTC` assertion, pinned the default `Timeout`, and added `forcesLocUTC` (real loc override), `miscasedParamHonorsServer`, and `credentialSubstringHonorsServer`. All config tests green.

Still a deliberate **ceiling raise, not a fix** — the fail-loud handling of over-limit events remains #652.

Issue #654, opened 2026-06-29
reconstruct/recover/query: unbounded Go-heap memory at scale (bounded by row-count, not payload size)


nethalo:
## Summary

The offline read/recover/reconstruct paths load the **entire** matching event set into Go memory and are bounded only by **row count** (`--limit`), never by **payload size**. At scale — a large or heavily-churned table, or a wide-window reconstruct — this OOMs the process. `--ultrafast` / DuckDB `memory_limit` / `temp_directory` spill do **not** help, because the unbounded growth is on the Go heap, outside DuckDB's accounting.

This is an analysis-led issue (the reconstruct OOM was **not** reproduced — the demo host has only 3.7 GB RAM and building a TB-scale churned table there isn't feasible). The recover buffering below **was** reproduced.

## Three related unbounded paths

1. **`reconstruct` full-table / PITR — OOM (not reproduced, code-evidence).** `ReconstructTable` (`internal/reconstruct/fulltable.go`) sets no `Limit` on the fetch, so every stage materializes the full window:
   - `engine.Fetch` → `scanRows` appends all live rows into `[]ResultRow`;
   - `parquetquery.Fetch` drains all archive rows into a Go slice (the DuckDB scan itself streams + spills, but the result is fully collected);
   - `query.MergeResults` sorts+dedups the whole union in RAM with `limit=0`;
   - the change map `changes := make(map[string]*ResultRow, len(events))` holds a pointer to **every** event keyed by PK.
   Two unbounded sizes: O(events in window) and O(distinct PKs touched). The project's own bench drafts already note "a huge-range full-table `_snapshot` can OOM; keep probes PK-filtered or cap-bounded" and only ever pre-flight ~1M events — a churned TB table is 1–2 orders of magnitude past that.

2. **`recover` whole-script buffer — reproduced.** `GenerateSQLFromRows` (`internal/recovery/recovery.go`) does `slices.Reverse(rows)` (all rows resident) and buffers the **entire** reversal script in one `bytes.Buffer` before writing (required by the up-front schema-drift refusal, so non-streamable by design). Reproduced: a **single-row** recover of one 40 MB blob produced a **55,924,543-byte** (~56 MB) script held entirely in memory. Bounded by `--limit` (default 1000) × per-row payload — so 1000 rows of fat images is gigabyte-scale, with no size guard.

[...]

Issue #653, opened 2026-06-29
recover: binary BLOB columns restored as base64 text — corrupt round-trip


nethalo:
## Summary

`bintrail recover` regenerates an INSERT/UPDATE that writes the **base64 text** of a binary BLOB column instead of the **original bytes**. Restoring the generated SQL would corrupt the column. Reproduced on EC2 with v0.23.0.

## Reproduced evidence

A `LONGBLOB` column `b` was set to 40 MB of byte `0x6D` (`REPEAT('m', 40MB)`), then the row was DELETEd and indexed. `recover --dry-run` for that PK emits:

```sql
INSERT INTO `blobtest`.`big` (`b`, `id`) VALUES ('bW1tbW1tbW1tbW1t...', 3);
```

`'bW1t'` is **base64 of `'mmm'`** (`printf 'mmmmmm' | base64` = `bW1tbW1t`). The reference encodings:
- base64 of `mmmmmm` → `bW1tbW1t`  ← **this is what recover emitted**
- hex of `mmmmmm` → `6d6d6d6d6d6d` (a correct `X'6d6d...'` literal would be this)

So executing the reversal would write the ASCII string `bW1tbW1t...` into the LONGBLOB — **not** the original 40 MB of `0x6D`. The blob round-trip is broken.

## Root cause

The capture path base64-encodes binary BLOBs into the JSON `row_before`/`row_after` columns (because they aren't valid JSON — see the sibling oversized-event issue). On the way **out**, `recover` reads the JSON value back as a Go **string** (the base64 text) and `formatValue` emits it as a quoted string literal. It never base64-**decodes** it back to bytes, so the documented `X'<hex>'` binary path (`internal/recovery/recovery.go`, ~:633) is never reached for these values — that hex path only fires for actual `[]byte`, which no longer exists after the JSON round-trip.

Note: this **refutes** the earlier inspection-only assumption that blobs are emitted as `X'<hex>'` at 2× size. Empirically the output is base64-sized (~1.333×) **and semantically wrong**.

## Likely blast radius

The same base64-in-JSON storage feeds `reconstruct` (single-row PITR and full-table mydumper output) and the shim's `_flashback`/`_snapshot`. Any path that re-materializes a binary BLOB from the index is suspect. Not yet verified for reconstruct — flagging for the fix to cover all consumers, not just `recover`.

## Proposed direction

[...]

PR #659, opened 2026-06-29
stream/index: fail loud on a flush failure instead of silently skipping

### LINKED ISSUE #652: capture: oversized BLOB/JSON row events (>~48 MB raw) are silently dropped — data loss

nethalo:
closes #652

## Summary

An event the index cannot accept — e.g. a row image whose serialized size exceeds the server's `max_allowed_packet` — was **silently dropped**:
- `stream`: the ticker/shutdown `checkpoint()` swallowed the flush error (`slog.Warn` + return) and kept going.
- offline `index`: logged "indexing failed" but returned **exit 0** — a cron/CI wrapper read it as success.

(Reproduced on EC2: source had 2 rows, index had 1.)

The **batch-full** and **DDL** flush paths in `streamLoop` *already* propagate flush errors (`if err := flush(); err != nil { return err }`). The bug was a single inconsistent path. Occam: align it, no new machinery.

## Changes
- **`internal/streamrun/streamrun.go`** — `checkpoint()` returns the flush error instead of swallowing it; the 3 call sites (ticker, ctx-done, channel-closed) propagate it. The stream aborts loudly and replays from the last durable checkpoint on restart → **transient errors lose nothing** (re-streamed from the binlog); a genuinely un-indexable event aborts loudly rather than vanishing. A `saveCheckpoint` failure stays a warning (it only causes a re-stream on restart, not data loss).
- **`cmd/bintrail/index.go`** — `runIndex` returns non-zero when any file failed, plus a `failed_files` count in `--format json`. `--all` still processes every file; the per-file `status='failed'` record in `index_state` is unchanged.

## Trade-off (approved)
A truly un-indexable event (now > 1G after #656/#657, very rare) makes the stream **abort and retry** (potentially crashloop) rather than be skipped — this is the same behavior the batch-full path already had, made consistent. It is **fail-loud, not silent-skip**; the operator raises `max_allowed_packet` or excludes the table. The fancier record-and-advance alternative was deliberately not built (Occam).

## Scope guard
Not in scope: storage encoding (#653), unbounded read-path memory (#654). `#656`/`#657` raised the ceiling (stopgap); this is the fail-loud fix.

## Test plan
- [x] `internal/streamrun` + `cmd/bintrail` integration (`-tags integration`) — pass, **no regression** in existing streamLoop / live-replication / gtid tests
- [x] New `TestStreamLoop_flushFailurePropagates` — a checkpoint flush failure now propagates (was a silent nil)
- [x] New `TestRunIndex_failedFileReturnsNonZero` — a failed file now exits non-zero (was exit 0)

### DIFF
--- a/cmd/bintrail/index.go
+++ b/cmd/bintrail/index.go
@@ -219,28 +219,49 @@ func runIndex(cmd *cobra.Command, args []string) error {
 	})
 
 	var totalEvents int64
+	var failedFiles int
 	for _, filename := range files {
 		n, err := indexFile(ctx, p, idx, indexDB, idxBinlogDir, filename, bintrailID)
 		totalEvents += n
 		if err != nil {
-			// Log and continue so --all processes remaining files.
+			// Log and continue so --all still processes the remaining files;
+			// the failure is surfaced as a non-zero exit after the loop (#652).
 			slog.Error("indexing failed", "file", filename, "error", err)
+			failedFiles++
 		}
 	}
 
-	slog.Info("indexing complete", "files_processed", len(files), "events_indexed", totalEvents)
+	slog.Info("indexing complete",
+		"files_processed", len(files),
+		"events_indexed", totalEvents,
+		"failed_files", failedFiles)
 
 	if idxFormat == "json" {
-		return cliutil.OutputJSON(struct {
+		if err := cliutil.OutputJSON(struct {
 			FilesProcessed int   `json:"files_processed"`
 			EventsIndexed  int64 `json:"events_indexed"`
+			FailedFiles    int   `json:"failed_files"`
 		}{
 			FilesProcessed: len(files),
 			EventsIndexed:  totalEvents,
-		})
+			FailedFiles:    failedFiles,
+		}); err != nil {
+			return err
+		}
+	} else {
+		fmt.Printf("\nTotal events indexed: %d\n", totalEvents)
+		if failedFiles > 0 {
+			fmt.Printf("WARNING: %d of %d file(s) failed to index — see logs above\n", failedFiles, len(files))
+		}
 	}
 
-	fmt.Printf("\nTotal events indexed: %d\n", totalEvents)
+	// Fail loud: a file that could not be fully indexed (e.g. an event the
+	// server rejects as over max_allowed_packet) must not exit 0 and read as
+	// success to a cron/CI wrapper (#652). The per-file failure is already
+	// recorded as status='failed' in index_state.
+	if failedFiles > 0 {
+		return fmt.Errorf("indexing finished with %d of %d file(s) failed", failedFiles, len(files))
+	}
 	return nil
 }
 
--- a/cmd/bintrail/index_integration_test.go
+++ b/cmd/bintrail/index_integration_test.go
@@ -1,0 +1,68 @@
+//go:build integration
+
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/dbtrail/dbtrail/internal/testutil"
+)
+
+// TestRunIndex_failedFileReturnsNonZero verifies that when a binlog file cannot
+// be indexed, `bintrail index` exits non-zero instead of logging the failure and
+// returning success (exit 0) — the silent failure a cron/CI wrapper read as a
+// clean run (#652). The per-file failure is still recorded as status='failed' in
+// index_state and --all still processes the remaining files; only the final exit
+// code changes.
+//
+// The failure is forced with a garbage binlog file (invalid magic), which the
+// parser rejects — a deterministic stand-in for the oversized-event rejection
+// that motivated the issue, reaching the same per-file error aggregation.
+func TestRunIndex_failedFileReturnsNonZero(t *testing.T) {
+	indexDB, name := testutil.CreateTestDB(t)
+	testutil.InitIndexTables(t, indexDB)
+	// A snapshot must exist so EnsureResolver (sourceDB=nil) loads from the index
+	// and execution reaches the file loop.
+	testutil.InsertSnapshot(t, indexDB, 1, "2026-01-01 00:00:00",
+		"testdb", "orders", "id", 1, "PRI", "int", "NO")
+
+	dir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(dir, "binlog.000001"), []byte("not a valid binlog file"), 0o644); err != nil {
+		t.Fatalf("write garbage binlog: %v", err)
+	}
+
+	saved := struct {
+		index, source, binlogDir, files, format, schemas, tables string
+		all, skip                                                bool
+	}{idxIndexDSN, idxSourceDSN, idxBinlogDir, idxFiles, idxFormat, idxSchemas, idxTables, idxAll, idxSkipSourceCheck}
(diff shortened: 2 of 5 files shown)


nethalo:
## Self-review (3 agents) + fixes — `0ecbb6b`

Ran silent-failure-hunter, code-reviewer, pr-test-analyzer.

### 🔴 CRITICAL (silent-failure-hunter) — fixed
My ticker-path propagation exposed a **latent deadlock in `streamrun.One`**: when `streamLoop` returns a flush error mid-stream (ctx still live), `One`'s `<-parseErrCh` drain hangs forever — the parser goroutine returns only on ctx cancel, and `One`'s *deferred* cancel can't run while `One` is blocked on the drain. The fail-loud abort would become a **silent wedge** (green-but-stuck to the supervisor) — worse than the bug it fixes.

**This deadlock already existed** on the pre-existing batch-full and DDL `return err` paths; my change just made it reachable via the common ticker path. Fixed by extracting the drain into `drainParser()`, which **cancels before the receive**. `cancel()` shadows `One`'s *derived* ctx (not the parent), so `monitor.go`'s `ctx.Err()` restart gate is unaffected → loud crash → restart → replay from the durable checkpoint.

Proven a **real anchor**: `TestDrainParser_cancelsBeforeReceive` hangs to its 5s watchdog when the `cancel()` is reverted, passes (0.00s) with it.

### code-reviewer — clean
No high-confidence issues. Verified all 3 call sites, the `--all` aggregation ordering, the `--format json` error contract, metrics/continuity (no regression), and the `ctx.Err()` supervisor interaction.

### pr-test-analyzer — addressed
- **Gap 1**: `TestRunIndex_allContinuesPastFailure` (2 files; both must be attempted — catches a fail-fast regression of `--all`).
- **Gap 2**: `TestStreamLoop_saveCheckpointFailureDoesNotAbort` (a checkpoint-write failure must stay a warning, not abort).
- **Q4**: tightened the index assertion to the distinctive `"file(s) failed"` phrase so an early-return setup regression can't pass it for the wrong reason.

All green: unit suite + `internal/streamrun` & `cmd/bintrail` integration, no regression in existing streamLoop / live-replication / gtid tests.

nethalo:
## Round-2 self-review — clean, cosmetic fixes only (`0c74b0b`)

Re-reviewed the deadlock fix + new tests with 4 agents (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer). **Consensus: the fix and tests are correct and mergeable.** Verified against the real code:

- `drainParser`'s `cancel()` unblocks the parser on **every** error path (GetEvent *and* a full-buffer `out <- ev` — every send is a `select` with `<-ctx.Done()`).
- It does **not** mask a genuine parser error: `sp.Run` converts cancellation to a `nil` return, and on the error paths One returns `loopErr` before consulting `parseErr` anyway.
- `cancel()` cancels One's **derived** ctx (shadowed), not the supervisor's parent — so `monitor.go`'s `ctx.Err()` restart gate still fires → loud "failed" badge → restart → replay. (Had it touched the parent, the supervisor would mis-classify the crash as a clean "stopped" — the fix gets this right.)
- All 4 new tests are real anchors, none flaky, none pass for the wrong reason.

Cosmetic items applied:
- `func One` had lost its doc comment (drainParser was inserted between One's doc and the function — Go re-attached it). Moved drainParser above; both documented.
- Corrected the drainParser comment (`sp.Run` returns `nil` on cancel, not `context.Canceled`; softened "returns only when cancelled" to the blocked-state case).
- Clarified the `TestDrainParser` comment scope (helper contract in isolation).

Noted as a follow-up (pre-existing, not introduced here): the synchronous DDL hook runs ctx-unaware index-DB queries inside the parser goroutine — bounding those with a context would close the last theoretical hang. Out of scope for #652.

Issue #661, opened 2026-06-29
shim: _flashback/_snapshot return BLOB/TEXT as base64 text (same root as #653)


nethalo:
Sibling of #653 (fixed for \`recover\`). The shim's `_flashback`/`_snapshot` virtual schemas return reconstructed rows over the **MySQL wire protocol** (`internal/shim/`), so a BLOB or TEXT column comes back to the client as the base64 TEXT it is stored as, not the original bytes/string.

Different emission mechanism than recover (`X'hex'` SQL) and reconstruct (mydumper) — needs its own decode pass when building the resultset row values. Same root: go-mysql delivers BLOB/TEXT as `[]byte` → `marshalRow` base64-encodes them (confirmed in #653).

The fix mirrors #653's `base64Cols`/`decodeStoredBase64`: gate on the resolver's column DataType (blob-family → raw bytes, text-family → string), decode the stored base64 before returning the value to the client.

Issue #660, opened 2026-06-29
reconstruct: BLOB/TEXT columns emitted as base64 text (same root as #653)


nethalo:
Sibling of #653 (fixed for \`recover\`). The same storage encoding — go-mysql delivers BLOB and TEXT (both \`MYSQL_TYPE_BLOB\`) as \`[]byte\`, which \`marshalRow\` base64-encodes into the JSON columns — corrupts `reconstruct` output too.

`reconstruct` emits **mydumper-format** full-table / single-row state (`internal/reconstruct/fulltable.go`), a different emission mechanism than recover's `X'hex'` SQL, so it needs its own fix: for BLOB columns base64-decode → mydumper binary form, for TEXT → the original string.

The fix mirrors #653's `base64Cols` / `decodeStoredBase64` (recovery.go) but for the mydumper writer. Worth hoisting the helper into a shared spot once the second consumer (shim, sibling issue) lands.

Verified root in #653: a row with `bl BLOB`, `tx TEXT` stores `{"bl":"<base64>","tx":"<base64>"}`; only VARCHAR/CHAR come through as plain strings.

PR #662, opened 2026-06-29
recover: decode BLOB/TEXT columns instead of emitting their base64 text

### LINKED ISSUE #653: recover: binary BLOB columns restored as base64 text — corrupt round-trip

nethalo:
closes #653

## Summary

`recover` restored a binary BLOB (and a TEXT) column as the **base64 text** it is stored as, not the original value — a corrupt round-trip (reproduced on EC2: a LONGBLOB recovered as `'bW1t...'`).

**Root** (verified empirically with a storage-encoding probe): go-mysql delivers BLOB *and* TEXT columns (both binlog type `MYSQL_TYPE_BLOB`) as `[]byte`, which `marshalRow` base64-encodes into the `binlog_events` JSON. On read the value comes back as a base64 **string**, and `FormatSQLValue`'s string branch emitted it verbatim. VARCHAR/CHAR arrive as plain strings and were always fine.

## Fix (recover, MySQL dialect only)

- `base64Cols` types each column from the resolver: BLOB family (binary) vs TEXT family (text).
- `decodeStoredBase64` reverses the storage encoding before formatting: **BLOB → `[]byte` → `X'hex'`** (reusing the existing `[]byte` branch — no new emission path), **TEXT → the original string**.
- Untouched: VARCHAR/CHAR (plain strings), numbers, and a BLOB/TEXT whose bytes were valid JSON (stored raw as a map/array, not a string). A non-decodable string is left as-is (defensive).
- Gated to `MySQLDialect` (PG bytea is a separate concern).

Mirrors the existing `coerceUnsigned` per-column-type coercion pattern.

## Scope

Recover only. The same root corrupts **reconstruct** (mydumper output) and the **shim** (wire-protocol resultsets) — different emission mechanisms, filed as #660 and #661. Out of scope: VARBINARY/BINARY (delivered as strings, a separate U+FFFD capture-loss), PG bytea.

## Test plan
- [x] `internal/recovery` integration — pass, no regression
- [x] New `TestGenerateSQL_blobAndTextRoundTrip`: asserts the exact `X'42494e00424c4f42'` (BLOB) + `'hello text'` (TEXT), and that the base64 strings do NOT survive. **Proven a real anchor** — neutralizing the fix makes it fail with "emitted as base64 text".


nethalo:
## Self-review (4 agents) + fixes — `bac37f8`

Ran code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer. The fix was verified correct (go-mysql BLOB/TEXC→[]byte→base64 confirmed against v1.13.0 source; literals/hex recomputed). Items addressed:

- 🟠 **code-reviewer (~65) — completeness fix:** a BLOB/TEXT column can be a **PK with a prefix length** (`PRIMARY KEY (c(20))`), and `pkWhereClause` emitted its base64 value **undecoded** → the WHERE matched zero rows (silent no-op recovery). My "PKs can't be BLOB/TEXT" premise was wrong. Now decodes there too (same `base64Cols`, MySQL-gated) so VALUES/SET/WHERE are consistent.
- 🔵 **pr-test-analyzer (rated 8):** the `buildUpdate` SET path was untested (a regression on it would have passed). Added `TestGenerateSQL_updateSetBlobTextDecoded` (SET decode + NULL→NULL + empty-blob→`X''`) and `TestGenerateSQL_textPKWhereDecoded` (the WHERE fix).
- 🔵 **comment-analyzer:** the helpers were inserted between `generateInsert`'s doc and `func generateInsert`, orphaning the doc. Relocated; both documented.
- 🔵 **silent-failure-hunter:** documented the **pre-existing** JSON-scalar-string blob edge (a valid-JSON string blob is indistinguishable from base64 after the storage round-trip — already corrupt at capture, out of scope) and the resolver-unavailable degradation. Confirmed the decode-failure fallback and the PG gate are sound (no new silent failure).
- Noted GEOMETRY/VECTOR are also `[]byte` but out of scope (a bare `X'hex'` can't load them — need `ST_GeomFromWKB`).

All green: unit suite + `internal/recovery` integration, no regression.

PR #663, opened 2026-06-29
reconstruct: decode BLOB/TEXT delta-event values instead of base64 text

### LINKED ISSUE #660: reconstruct: BLOB/TEXT columns emitted as base64 text (same root as #653)

nethalo:
closes #660

## Summary

Sibling of #653 (recover). The full-table mydumper `reconstruct` merges baseline rows (DuckDB scan → `[]byte`, already correct) with **delta-event** `row_after` images. go-mysql delivers BLOB/TEXT as `[]byte` → `marshalRow` base64-encodes them into `binlog_events` JSON → a delta event's blob/text value is a base64 **string**, which the writer's `recovery.FormatSQLValue` emitted verbatim. So **any row touched by a binlog event** since the baseline reconstructed with corrupt blob/text.

## Fix

`rowAfterOrdered` decodes BLOB/TEXT columns (typed from the resolver via `binaryColsFromTableMeta`, threaded through `mergeInput.BinaryCols`) before the tuple reaches the writer: **BLOB → `[]byte`** (FormatSQLValue emits `X'hex'`), **TEXT → the original string**. Baseline-sourced values arrive as `[]byte` (not a string) and pass through untouched — so only delta-event base64 strings are decoded; NULL passes through.

The `base64StoredKind`/`decodeStoredBase64` helpers are duplicated from `internal/recovery` (those are unexported); a follow-up should hoist a single shared copy once #661 (shim, the third consumer) lands.

## Scope
Out of scope, same as #653: VARBINARY/BINARY (delivered as strings — separate U+FFFD capture loss) and PG bytea. Shim (#661) is the remaining sibling.

## Test plan
- [x] `internal/reconstruct` unit + integration (full merge path) — pass, no regression
- [x] New `TestRowAfterOrdered_decodesBlobText`: BLOB→`[]byte`, TEXT→string, NULL→nil, baseline `[]byte` passes through. **Proven a real anchor** — neutralizing the decode makes it fail (value stays the base64 string).

### DIFF
--- a/internal/reconstruct/fulltable.go
+++ b/internal/reconstruct/fulltable.go
@@ -3,6 +3,7 @@ package reconstruct
 import (
 	"context"
 	"database/sql"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -407,6 +408,7 @@ func ReconstructTable(
 		Changes:           changes,
 		OutputDir:         cfg.OutputDir,
 		ChunkSize:         cfg.ChunkSize,
+		BinaryCols:        binaryColsFromTableMeta(tm),
 	}, rep); err != nil {
 		return nil, err
 	}
@@ -434,6 +436,10 @@ type mergeInput struct {
 	Changes           map[string]*query.ResultRow
 	OutputDir         string
 	ChunkSize         int64
+	// BinaryCols maps each BLOB/TEXT column to whether it is binary, so
+	// delta-event row_after values (stored base64) are decoded before emission
+	// (#660). nil = nothing to decode.
+	BinaryCols map[string]bool
 }
 
 // mergeBaselineIntoWriter streams the local baseline Parquet via DuckDB,
@@ -491,6 +497,15 @@ func mergeBaselineIntoWriter(ctx context.Context, in mergeInput, rep *TableRepor
 	if err := mw.WriteSchema(in.CreateTableSQL); err != nil {
 		return err
 	}
+
+	// Decode the BLOB/TEXT base64 of the delta-event after-images up front, on
+	// the Changes map — the one place where every value is unambiguously
+	// event-sourced (loaded from binlog_events JSON, so blob/text are uniformly
+	// base64 strings). This covers BOTH event emit paths in mergeBaselineImages
+	// (the matched-PK emit and the leftover-INSERT drain, both reading
+	// in.Changes) without touching baseline pass-through rows, whose TEXT values
+	// arrive from the DuckDB scan as Go strings and must NOT be decoded (#660).
+	decodeChangeBinaries(in.Changes, in.BinaryCols)
 
 	// emit re-orders each emitted row map into the baseline Parquet column
 	// order the writer was constructed with. For baseline pass-through rows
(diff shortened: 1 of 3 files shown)


nethalo:
## Self-review (4 agents) caught a CRITICAL bug I introduced — fixed (force-pushed)

All four reviewers independently reproduced a **silent data-corruption** bug in the first version: I decoded BLOB/TEXT at emit time in `rowAfterOrdered`, which serves BOTH baseline pass-through rows and event after-images. My safety assumption ("baseline values arrive as []byte") was **false for TEXT** — the baseline writer stores TEXT as `parquet.String()` (`baseline/schema.go:146`), which DuckDB scans back as a Go **string** (the repo documents this in `pk_canonicalize.go:90`). So an **untouched baseline TEXT value whose content is valid base64** (a hex digest, a token, `"test"`) was silently base64-decoded into garbage in the dump.

**Fix — decode by provenance, not by Go type.** `decodeChangeBinaries` decodes the BLOB/TEXT base64 of every delta event's `RowAfter` up front, on the **Changes map** — the one boundary where every value is unambiguously event-sourced (loaded from `binlog_events` JSON). It covers both event emit paths (matched-PK + leftover-INSERT, both reading Changes) and `rowAfterOrdered` no longer decodes anything, so baseline pass-through rows are never touched.

**The anchor test that would have caught it:** `TestMergeBaseline_blobTextProvenance` emits an untouched baseline TEXT (`'YWJjZA=='`, valid base64) and asserts it survives **verbatim**, plus an event TEXT that IS decoded. Proven a real anchor — re-introducing the decode in `rowAfterOrdered` makes it fail with "baseline TEXT was wrongly base64-decoded". Added `TestDecodeChangeBinaries` and `TestBinaryColsFromTableMeta` (the builder was unguarded).

Also: this branch was accidentally stacked on #653 — **rebased onto main**, so the diff is now reconstruct-only (no #653 dependency). All green: unit suite + `internal/reconstruct` integration.

Lesson logged: "decode by where the value came from, not by inspecting the value" — the same class as #653's JSON-scalar-string edge.

nethalo:
## Round-2 re-review — fix confirmed correct & complete; refinements applied

Re-reviewed the provenance fix with 4 agents. **Consensus: the round-1 corruption is genuinely gone, the fix is correct and complete.** Verified: `rowAfterOrdered` no longer decodes (single caller of `decodeStoredBase64` = `decodeChangeBinaries`); `decodeChangeBinaries` only touches the event-sourced `Changes` map and covers both emit paths (matched-PK + leftover-INSERT, same pointers, no clone); in-place mutation is safe (PK matching uses stored `PKValues`, not RowAfter; `MapEventEnumLabels` precedent; nothing reads pre-decode after).

Refinements applied:
- **Test seam closed (the round-1 failure shape):** `TestMergeBaseline_blobTextProvenance` now **derives `BinaryCols` through the real `binaryColsFromTableMeta(tm)`** instead of hand-injecting it, and adds a **BLOB column** so the `X'hex'` output branch is wired-tested (was unit-only). Baseline BLOB → `X'524157'`, event BLOB → decoded `X'42494e'`, baseline TEXT verbatim, event TEXT decoded. Still anchor-proven (re-introducing the decode in `rowAfterOrdered` fails it).
- **Comment accuracy:** reworded the `internal/recovery` attribution (those helpers live in the sibling #662, not in this base) and narrowed the pre-existing `mergeBaselineImages` "can never drift" comment — the writer now decodes but the shim does not, so the surfaces DO diverge on blob/text until **#661** (the shim fix) lands. That confirms #661 is real and needed.

All green: unit suite + `internal/reconstruct` integration.

Issue #672, opened 2026-06-30
verify: false MISMATCH on BLOB/TEXT — SnapshotFullTableImages never decodes base64, isDeferredType doesn't cover TEXT


nethalo:
Found during the #668 review (full-table reconstruct epoch-blind BLOB/TEXT decode). This is a distinct, more severe bug: it isn't epoch-blind, it's **undecoded entirely**, and it can produce a false MISMATCH verdict in `verify`'s core correctness check.

### Mechanism

go-mysql delivers BLOB/TEXT as `[]byte`, which `marshalRow` base64-encodes into the stored event JSON (VARCHAR/CHAR arrive as a plain Go `string`, stored unencoded). Every other full-table-image consumer in `internal/reconstruct` decodes this before use — the writer path (`mergeBaselineIntoWriter` / `decodeChangeBinaries`, the subject of #668) and the single-row path (`DecodeEventBinaries`, #666/#667).

`SnapshotFullTableImages` (`internal/reconstruct/fulltable.go:765-780`) does **not**. It calls `mergeBaselineImages(ctx, mergeCore{..., Changes: in.Changes}, emit)` directly with zero decode step — no `DecodeEventBinaries`, no `decodeChangeBinaries`, nothing. Confirmed via grep across the three production callers — none of them decode before or around the call either:

- `internal/shim/snapshot.go:184` — the `_snapshot` virtual-schema full-table path
- `internal/verify/verify.go:228` — baseline-anchored full-table reconstruction for drift comparison
- `internal/verify/explain_baseline.go:211` — the `--explain` row-level drill-down

### Why this is worse than #668: a false MISMATCH, not just corruption-on-edge-case

`internal/verify/verify.go:258-278` has a `hasDeferredRepr`/`isDeferredType` guard that downgrades comparisons on types whose event-image representation can legitimately differ from the baseline/source render (ENUM/SET ordinal-vs-label, JSON canonicalization, binary families) to `Inconclusive` instead of a false `Mismatch`:

```go
func isDeferredType(dataType string) bool {
	switch strings.ToLower(dataType) {
	case "enum", "set", "json",
		"binary", "varbinary", "blob", "tinyblob", "mediumblob", "longblob", "bit":
		return true
	}
	return false
}
```

[...]

Issue #666, opened 2026-06-30
reconstruct: single-row path returns BLOB/TEXT as base64 text (sibling of recover #662 / shim #661)


nethalo:
The single-row reconstruction path (`bintrail reconstruct` CLI and the console Time-travel reconstruct) returns BLOB/TEXT column values as the **base64 text** they are stored as, not the real bytes/string — the same root as recover #653/#662, reconstruct full-table #663, and shim #661.

### Mechanism (confirmed by inspection)
- go-mysql delivers BLOB/TEXT as `[]byte`, which `marshalRow` base64-encodes into the stored event JSON.
- The single-row path folds events onto the baseline via `reconstruct.ApplyAt` / `BuildHistory` (`internal/reconstruct/reconstruct.go`). `applyEvent` just `copyMap(ev.RowAfter)` — **no base64 decode**.
- `MapEventEnumLabels` (called first in both `internal/cli/reconstruct.go:300` and `internal/console/reconstruct.go:280`) decodes ENUM/SET epoch-aware but **not** BLOB/TEXT base64.
- The full-table merge path (#663) decodes via `decodeChangeBinaries` at `internal/reconstruct/fulltable.go:508`, but that is **not** on the single-row path.

Result: a single-row reconstruct whose answer is event-sourced (an UPDATE/INSERT touched the row in the window) emits the base64 string for any BLOB/TEXT column, instead of the real value. A never-touched baseline row is unaffected (DuckDB scans the raw value).

### Scope
- `bintrail reconstruct` (single-row, incl. `--history`, `--sql`-via-baseline, `--format json/table/csv`).
- Console Time-travel reconstruct (`internal/console/reconstruct.go`).
- The shim (#661, fixed) decodes its deltas before `ApplyAt`, so it is not affected.

### Fix
Decode the fetched events' BLOB/TEXT base64 **before** the `ApplyAt`/`BuildHistory` fold, **epoch-aware** (typing each column at the snapshot in effect at the event's timestamp — a VARCHAR→TEXT widening must not wrongly decode an old plain-string value), and only on event-sourced rows (never the baseline). Mirrors the shim #661 fix.

### Related (separate follow-up, not this issue)
The full-table merge's `decodeChangeBinaries` types columns via `binaryColsFromTableMeta(tm)` from a single snapshot — it is **epoch-blind**, so it carries the same VARCHAR→TEXT corruption edge #661 fixed in the shim. Worth a separate look.

PR #673, opened 2026-06-30
reconstruct: full-table BLOB/TEXT decode is epoch-aware (#668)

### LINKED ISSUE #668: reconstruct: full-table BLOB/TEXT decode is epoch-blind (VARCHAR→TEXT corruption edge)
The full-table reconstruction merge path (`bintrail reconstruct --output-format mydumper`) decodes BLOB/TEXT base64 using column types from the **latest** snapshot only, so a column widened `VARCHAR→TEXT` across the reconstruction window can corrupt old plain-string values.

Surfaced during the #666 review (single-row reconstruct fix) and noted there; filing as its own issue so it isn't lost.

### Mechanism
- go-mysql delivers BLOB/TEXT as `[]byte`, which `marshalRow` base64-encodes into the stored event JSON. A `VARCHAR`/`CHAR` column is delivered as a Go `string` → stored **plain** (never base64).
- The full-table path (#663) decodes via `decodeChangeBinaries` at `internal/reconstruct/fulltable.go:508`, typed by `binaryColsFromTableMeta(tm)` (line 411).
- `tm` comes from `resolver.Resolve(schema, table)` (line 306) where `resolver = metadata.NewResolver(db, 0)` (line 152) — **snapshot id 0 = the latest snapshot**, used for every event regardless of its epoch.

Result: if a column was `VARCHAR` when an old event was captured (value stored plain) and is `TEXT` in the latest snapshot, `binaryColsFromTableMeta` flags it, and `decodeStoredBase64` runs on the old plain value. A plain value that happens to be valid base64 (e.g. `"test"`, `"user"`, any 4/8-char run in the base64 alphabet) is silently decoded to garbage bytes in the mydumper output.

### Same class as the now-fixed siblings
- recover #662, shim #661 (PR #665), single-row reconstruct #666 (PR #667) all resolve BLOB/TEXT column typing **per event epoch** (the snapshot in effect at the event's timestamp) precisely to avoid this. The full-table path is the remaining epoch-blind consumer.

### Fix sketch
Type the decode columns per event epoch instead of from the single latest `tm` — mirror `reconstruct.DecodeEventBinaries` (#666) / the shim's `mapEventImages` (#661): build the BLOB/TEXT column map at the snapshot in effect at each event's timestamp (`EpochAt` → `NewResolver(db, id)`), memoized per epoch. The events already run through `MapEventEnumLabels` (epoch-aware) at `fulltable.go:383` right before the change map is built, so the same per-epoch resolution is available.

Caution: the full-table path applies `decodeChangeBinaries` to the `in.Changes` map (the same event objects `MapEventEnumLabels` mutates). A per-epoch decode must replace `decodeChangeBinaries`, not stack on top of it, to avoid double-decoding (`decodeStoredBase64` is not idempotent).

### Severity
Lower than the single-row/shim bugs: full-table at least decodes (just by the wrong epoch on a column-type-drift edge), where the single-row path did not decode at all. BLOB columns are unaffected at the JSON layer but the mydumper output is a real `X'hex'`/string emission, so a misclassified value does materialize wrong.

nethalo:
closes #668

## Summary
- `ReconstructTable`'s full-table writer path typed BLOB/TEXT decode columns once from the *latest* schema snapshot and applied that to every delta event regardless of when it was captured — a column captured as VARCHAR (plain string) at an old epoch and later widened to TEXT (base64-encoded) would have its old plain value wrongly decoded under the new typing, corrupting any value that happens to be valid base64 (the issue's own example: `"test"`).
- Fix: call `DecodeEventBinaries` — the epoch-aware decode already built for the single-row reconstruct path (#666/#667) — on the full `events` slice before the change map is built, mirroring the established sibling pattern (recover #662, shim #661/PR #665) instead of writing new bespoke logic.
- Removed the now-redundant `decodeChangeBinaries`/`mergeInput.BinaryCols` from the writer path (would have double-decoded otherwise — `decodeStoredBase64` isn't idempotent) and updated the stale doc comments that referenced it.

## Test plan
- [x] Integration tests pass (`go test -tags integration ./... -count=1`) — except a pre-existing flake in `internal/cascade` unrelated to this change (binlog FK-cascade blind-spot test, green in isolation; not touched by this PR)
- [x] New integration test (`internal/cli/reconstruct_fulltable_epoch_blobtext_integration_test.go`) drives the real CLI path: two `schema_snapshots` epochs (`body` VARCHAR → TEXT), one delta event captured under each. Verified the test **fails against the pre-fix code** (corrupts the old VARCHAR value `"test"` to garbage bytes, exactly the issue's repro) and passes after the fix.

## Scope note
While researching this, found a separate, more severe bug in `SnapshotFullTableImages` (used by the shim's `_snapshot`, `verify`, and `verify --explain`) — it doesn't decode BLOB/TEXT at all, and `verify`'s `isDeferredType` doesn't cover TEXT types, so it can produce a false `MISMATCH` verdict. Filed separately as #672 since it's a different function/different consumers/different severity class.

### DIFF
--- a/internal/cli/reconstruct_fulltable_epoch_blobtext_integration_test.go
+++ b/internal/cli/reconstruct_fulltable_epoch_blobtext_integration_test.go
@@ -1,0 +1,193 @@
+//go:build integration
+
+package cli
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dbtrail/dbtrail/internal/baseline"
+	"github.com/dbtrail/dbtrail/internal/indexer"
+	"github.com/dbtrail/dbtrail/internal/testutil"
+)
+
+// TestRunReconstruct_fullTable_blobTextEpochDrift is the end-to-end regression
+// for #668: the full-table reconstruct writer path must type each delta
+// event's BLOB/TEXT decode columns from the schema snapshot in effect AT THE
+// EVENT'S TIMESTAMP, not from the latest snapshot. Before the fix,
+// ReconstructTable resolved the decode-column set once from the latest
+// snapshot and applied it to every event regardless of epoch.
+//
+// Repro: `body` starts as VARCHAR (epoch 1, delivered as a plain Go string by
+// go-mysql — never base64-encoded at storage time) and is later widened to
+// TEXT (epoch 2, delivered as []byte — base64-encoded at storage time, #668's
+// own "VARCHAR→TEXT" example). An old event captured under epoch 1 stores its
+// plain string "test" — which happens to be valid base64 — verbatim. A new
+// event captured under epoch 2 stores its TEXT value base64-encoded. Typing
+// both from the latest (epoch 2 / TEXT) snapshot makes the old plain "test"
+// look like a BLOB/TEXT column needing decode, corrupting it to garbage bytes;
+// typing each event by its own epoch (the fix) leaves it untouched and still
+// correctly decodes the new event.
+func TestRunReconstruct_fullTable_blobTextEpochDrift(t *testing.T) {
+	db, dbName := testutil.CreateTestDB(t)
+	testutil.InitIndexTables(t, db)
+	if err := indexer.EnsureSchema(db); err != nil {
+		t.Fatalf("EnsureSchema: %v", err)
+	}
+
+	h1 := time.Now().UTC().Add(-48 * time.Hour).Truncate(time.Hour)
+	h2 := h1.Add(time.Hour)
+
+	epoch1TS := h1.Format("2006-01-02 15:04:05")
+	epoch2TS := h2.Format("2006-01-02 15:04:05")
+
+	// Epoch 1 (snapshot_id=1, at h1): body is VARCHAR.
+	testutil.InsertSnapshot(t, db, 1, epoch1TS, "testdb", "orders", "id", 1, "PRI", "int", "NO")
+	testutil.InsertSnapshot(t, db, 1, epoch1TS, "testdb", "orders", "body", 2, "", "varchar", "NO")
+	// Epoch 2 (snapshot_id=2, at h2): body widened to TEXT.
+	testutil.InsertSnapshot(t, db, 2, epoch2TS, "testdb", "orders", "id", 1, "PRI", "int", "NO")
+	testutil.InsertSnapshot(t, db, 2, epoch2TS, "testdb", "orders", "body", 2, "", "text", "NO")
+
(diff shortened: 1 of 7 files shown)


nethalo:
Ran a 4-agent review pass (correctness, silent-failures, test-coverage, comments) with adversarial verification on the diff, mirroring the review depth used on the sibling PRs (#665/#667). 3 real issues confirmed and fixed in the latest commit:

1. **Test-coverage (important)**: `TestMergeBaseline_blobTextProvenance`'s delta-event fixture used `"hello"` as the "already decoded" value — not valid base64, so a reintroduced double-decode on the Changes/event path would silently no-op instead of failing the test. Switched to `"test"` (valid base64) and empirically confirmed: reintroducing the deleted double-decode now makes the unit test fail, where it previously passed silently.
2. **Comments**: `internal/shim/handler.go` and `internal/reconstruct/enumlabels_test.go` had stale comments naming `decodeChangeBinaries`/`TestDecodeChangeBinaries`, both deleted by this PR — updated to the current names.
3. **Comments**: `mergeBaselineImages`'s doc-comment described the old writer-side decode mechanism and an outdated `#661` reference (already stale before this PR) — updated to describe the current upstream `DecodeEventBinaries` mechanism and point at #672 (the still-open gap on the shim's `SnapshotFullTableImages` path).

All findings were independently re-verified (not just taken on the reviewer's word) before fixing. Full unit + integration suites still green.

nethalo:
Second review pass (5 agents: code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier) against the full PR diff. One real, corroborated finding fixed in the latest commit:

**Fixed**: `mergeBaselineImages`'s doc-comment (added in the previous review-fix commit) incorrectly claimed the shim's `SnapshotFullTableImages` caller doesn't decode BLOB/TEXT yet. Two independent reviewers (code-reviewer + comment-analyzer) caught this and I verified it directly: `runSnapshotFullTable` already decodes via `mapEventImages` (#661) before building the Changes map. Corrected the comment, and corrected #672's body on GitHub too, since it inherited the same shim misattribution from my original research — #672's real scope is just `verify`'s two `SnapshotFullTableImages` callers.

**Reviewed and intentionally not acted on** (all rated non-blocking/low-severity by their respective reviewers):
- A pre-existing, PR-unrelated stale comment at `fulltable.go:948-952` ("not yet in main" for already-merged #661/#662) — out of this PR's scope.
- A CHANGELOG.md line that'll go stale post-merge — this project documents changelog entries in a separate follow-up PR after merge (established pattern), not in the feature PR itself.
- A minor comment-redundancy nit in the new test (same fact stated 3x) — cosmetic, no functional issue.
- A philosophical note on the #666/#667 Debug-vs-Warn fallback-logging design — explicitly out of scope for this PR per the reviewer's own assessment.

Test coverage and silent-failure passes both came back clean (no critical/important gaps). Full unit + integration suites still green.

PR #674, opened 2026-06-30
verify: decode BLOB/TEXT before comparing, defer TEXT in isDeferredType (#672)

### LINKED ISSUE #672: verify: false MISMATCH on BLOB/TEXT — SnapshotFullTableImages never decodes base64, isDeferredType doesn't cover TEXT

nethalo:
closes #672

## Summary
- `SnapshotFullTableImages` never decodes BLOB/TEXT base64 itself — every caller must decode its own `Changes` map first. The shim already does this (#661); the offline reconstruct writer does it via #668. `verify`'s three call sites (`VerifyTable` in verify.go, `VerifyBaselinePair` in verify_baseline.go, `ExplainBaselinePairMismatch` in explain_baseline.go) didn't — so an event-sourced TEXT value reached comparison as raw base64, a guaranteed mismatch surfaced as a false `MISMATCH` verdict, verify's core correctness signal.
- Fix: call the existing `reconstruct.DecodeEventBinaries` at all three sites, right after `query.FetchMerged` and before the events fold into a `Changes` map — zero new decode logic, mirrors the established #668/#661 pattern exactly.
- `isDeferredType` (the safety net that downgrades known-divergent types like ENUM/JSON/binary to Inconclusive instead of a false MISMATCH) deliberately does **not** get TEXT added, despite the original issue's fix sketch suggesting it. Review surfaced that `VerifyTable`'s gate is coarse (table merely *contains* a deferred column + anything changed in the window), and TEXT columns are ubiquitous — adding TEXT there would make live-source verify report Inconclusive instead of a genuine MISMATCH on most real tables, defeating the point of decoding TEXT in the first place (once decoded, it's directly comparable text, unlike ENUM/JSON/binary's genuine representation gap).

## Test plan
- [x] Integration tests pass (`go test -tags integration ./... -count=1`) — except the pre-existing `internal/cascade` flake (binlog FK-cascade blind-spot test, unrelated package, confirmed green in isolation)
- [x] `TestVerifyTable_TextEventDecoded` (live-source path) and `TestVerifyBaselinePair_TextOnlyChange_Match` (baseline-anchored path, isolated from the explain path's own independent decode) each directly assert `StatusMatch` on a TEXT-only event change. Both confirmed to **fail against the pre-fix code** (via `git checkout main -- <file>`, one file at a time) and pass after.
- [x] `TestVerifyBaselinePair_TextEventDecoded` and `TestExplainBaselinePairMismatch_DeferredDrift` cover the explain drill-down path and the existing ENUM/JSON deferral behavior, respectively.
- [x] `TestIsDeferredType` pins the case list, including that TEXT is intentionally excluded.

### DIFF
--- a/internal/reconstruct/fulltable.go
+++ b/internal/reconstruct/fulltable.go
@@ -623,9 +623,10 @@ type mergeStats struct {
 // its own Changes map before it ever reaches this function: the writer caller
 // upstream in ReconstructTable, on the full events slice before the Changes
 // map is even built (DecodeEventBinaries, #668); the shim caller upstream in
-// runSnapshotFullTable, via mapEventImages (#661). `verify`'s two
-// SnapshotFullTableImages callers (reconstructDigest, the --explain drill-down)
-// are the remaining undecoded consumers (#672).
+// runSnapshotFullTable, via mapEventImages (#661); `verify`'s callers
+// (reconstructDigest and ExplainBaselinePairMismatch) the same way, upstream
+// of their own Changes map builds (#672). Every SnapshotFullTableImages caller
+// now decodes before this function ever sees a value.
 func mergeBaselineImages(ctx context.Context, in mergeCore, emit func(map[string]any) error) (mergeStats, error) {
 	var stats mergeStats
 
--- a/internal/verify/deferred_test.go
+++ b/internal/verify/deferred_test.go
@@ -8,23 +8,26 @@ import (
 	"github.com/dbtrail/dbtrail/internal/query"
 )
 
-// TestIsDeferredType pins the #672 addition: TEXT family types must be
-// deferred alongside BLOB/binary, since both are decoded by the same
-// epoch-aware DecodeEventBinaries call and degrade the same way (an
-// unresolvable epoch leaves the value as stored base64).
+// TestIsDeferredType pins the case list, including the #672 decision: TEXT
+// family types are decoded by DecodeEventBinaries (same as BLOB) but are NOT
+// deferred — once decoded, TEXT is directly comparable to the baseline/source
+// text, unlike ENUM/JSON/binary, so deferring it would mask genuine
+// divergences instead of guarding against an unresolved representation gap.
 func TestIsDeferredType(t *testing.T) {
 	deferred := []string{
 		"enum", "set", "json",
 		"binary", "varbinary", "blob", "tinyblob", "mediumblob", "longblob", "bit",
-		"text", "tinytext", "mediumtext", "longtext",
-		"TEXT", "Blob", // case-insensitive
+		"BLOB", "Enum", // case-insensitive
 	}
 	for _, dt := range deferred {
 		if !isDeferredType(dt) {
 			t.Errorf("isDeferredType(%q) = false, want true", dt)
 		}
 	}
-	notDeferred := []string{"int", "varchar", "char", "datetime", "double", "decimal"}
+	notDeferred := []string{
+		"int", "varchar", "char", "datetime", "double", "decimal",
+		"text", "tinytext", "mediumtext", "longtext", "TEXT",
+	}
 	for _, dt := range notDeferred {
 		if isDeferredType(dt) {
 			t.Errorf("isDeferredType(%q) = true, want false", dt)
--- a/internal/verify/deferred_test.go
+++ b/internal/verify/deferred_test.go
@@ -8,6 +8,30 @@ import (
 	"github.com/dbtrail/dbtrail/internal/query"
 )
 
+// TestIsDeferredType pins the #672 addition: TEXT family types must be
+// deferred alongside BLOB/binary, since both are decoded by the same
+// epoch-aware DecodeEventBinaries call and degrade the same way (an
+// unresolvable epoch leaves the value as stored base64).
+func TestIsDeferredType(t *testing.T) {
+	deferred := []string{
+		"enum", "set", "json",
+		"binary", "varbinary", "blob", "tinyblob", "mediumblob", "longblob", "bit",
+		"text", "tinytext", "mediumtext", "longtext",
+		"TEXT", "Blob", // case-insensitive
+	}
+	for _, dt := range deferred {
+		if !isDeferredType(dt) {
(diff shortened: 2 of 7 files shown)


nethalo:
Ran the same 5-agent review pass used on #673 (code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier) against the full diff. The decode wiring itself was cleared by all reviewers — consistent across all three call sites, no missed caller, correct DB handle, correct empty-map handling for `VerifyBaselinePair`'s truth-side read. Three real findings, all fixed in the latest commit:

1. **Design issue (code-reviewer, cross-checked with advisor)**: the original fix sketch added TEXT to `isDeferredType` as defense-in-depth alongside BLOB. But `VerifyTable`'s gate (`hasDeferredRepr`) is coarse — it doesn't check whether the TEXT column itself changed, only whether the table *contains* one and *anything* changed in the window. TEXT columns are ubiquitous (every `wp_options.option_value` is `longtext`), so this made live-source verify report Inconclusive instead of a genuine MISMATCH on most real-world tables — defeating the point of decoding TEXT in the first place (once decoded, it's directly comparable text, unlike ENUM/JSON/binary which have a genuine representation gap to defer). **Dropped TEXT from `isDeferredType`** — the decode wiring is the real fix; the deferred-type set stays scoped to types that still can't be compared byte-for-byte after decode.
2. **Test-coverage gap (silent-failure-hunter + pr-test-analyzer, independently corroborated via the same empirical method — reverting one source line at a time and re-running)**: `TestVerifyBaselinePair_TextEventDecoded` didn't actually isolate `VerifyBaselinePair`'s own decode call — it was silently backstopped by `ExplainBaselinePairMismatch`'s independent decode, and the one assertion that does fire (`StatusMismatch`) was satisfied regardless by an unrelated divergence in the fixture. Added `TestVerifyBaselinePair_TextOnlyChange_Match` — a TEXT-only change with no other divergence, asserting `StatusMatch` directly. Confirmed it fails when `verify_baseline.go`'s decode line alone is reverted.
3. **Test-coverage gap (same two reviewers)**: `VerifyTable` (live-source mode) had zero test coverage for this fix. Added `TestVerifyTable_TextEventDecoded`, mirroring the existing live-source test's structure. Confirmed it fails when `verify.go`'s fix is reverted.

Also fixed a stale comment in `internal/reconstruct/fulltable.go` (outside this PR's original diff, but invalidated by it) that named `verify`'s two callers as the remaining undecoded `SnapshotFullTableImages` consumers.

Full unit + integration suites still green (same pre-existing, unrelated `internal/cascade` flake as #673, confirmed green in isolation).

PR #688, opened 2026-07-01
Console: full lifecycle for bintrail verify (trigger, poll, explain)

### LINKED ISSUE #677: Console: full lifecycle for bintrail verify (trigger, poll, explain) — not just read results
## Problem

`bintrail verify` (docs/verify.md) proves the recovery chain — baseline + indexed binlog deltas — would faithfully reproduce a table, but it is **CLI/cron/CI-only**. The console has no surface for it at all today; I confirmed this by grepping `internal/console/` — the only "verify"-adjacent thing in the UI is the unrelated stream-**continuity** badge (`ok`/`gap_lost`/`unknown`, PR #649), which answers a different question ("did capture drop events?" vs. "would reconstruction reproduce this table?").

This issue is for the **whole lifecycle**, not a read-only results viewer: trigger a run from the console, watch it progress, see per-table results, and drill into a mismatch — mirroring what the Baseline-trigger UI (#613/PR #615) already did for `bintrail baseline`.

## Proposed scope

Mirror the `BaselineController` pattern (`internal/console/baseline_trigger.go`) with a `VerifyController`:

```go
type VerifyController interface {
    Trigger(req VerifyRequest) error         // starts a run in the background, one at a time per server (409 if already running)
    Status(serverID string) VerifyStatus     // pollable: state, per-table results as they land, summary counts
    Explain(serverID, schema, table string) (*verify.MismatchExplanation, error)  // on-demand row-level drill-down for one mismatched table
}
```

Backed directly by the existing engine — no new verify logic needed, just orchestration:
- `internal/verify.FindBaselinePair` / `VerifyBaselinePair` (baseline-anchored, default) and `VerifyTable` (live-source) — same calls `internal/cli/verify.go` already makes per table.
- `internal/verify.ExplainBaselinePairMismatch` for the drill-down, wired the same way `verify --explain` does today.

Endpoints (mirroring `POST/GET /api/servers/{id}/baseline`):
- `POST /api/servers/{id}/verify` — body: `mode` (`baseline-anchored` default | `live-source`), optional `tables` filter. Returns 202 + initial status.
- `GET /api/servers/{id}/verify` — poll status: `state` (idle/running/succeeded/failed), per-table results accumulated so far (status/detail/digests), summary (match/mismatch/inconclusive/error counts), started/finished timestamps.
- `GET /api/servers/{id}/verify/explain?schema=&table=` — on-demand drill-down for one mismatched table (NOT precomputed for every mismatch — `ExplainBaselinePairMismatch` re-reconstructs, so it should only run when the operator asks for a specific table).

Gating:
- Requires a baseline destination configured for the server (same precondition `handleBaselineTrigger` already checks: `e.BaselineDir`/`e.BaselineS3`) — baseline-anchored mode needs at least two baselines to compare.
- `live-source` mode additionally requires the server's source DSN and should carry an explicit "this reads the whole table off your live source — run off-peak" confirmation in the UI, matching the warning already in docs/verify.md.
- Extend `GET /api/capabilities` with a `verify_trigger` bool, same shape as `baseline_trigger`.

## Open question: default-on or opt-in?

Unlike the baseline trigger (which shells `mydumper`, a new subprocess — #676 argues that one should flip to opt-out), `verify` is **already documented as read-only and reuses only in-process DuckDB/MySQL reads** the console already does for `reconstruct`. That's a meaningfully lower risk bar, so this could reasonably default ON from day one rather than repeating the opt-in-then-flip cycle — worth deciding explicitly during implementation rather than defaulting to opt-in by habit. The `live-source` sub-mode is the one piece that isn't "free" (full live-table scan) and may warrant its own confirmation step regardless of the overall feature's default.

## Frontend

A "Verification" panel (Storage tab, next to Baselines) with:
- A "Run verification" trigger (mode selector, defaulting to baseline-anchored) + status while running.
- A per-table results table (match/mismatch/inconclusive/error, color-coded, with `Detail`).
- An "Explain" action on mismatched rows that fetches and renders the PK/column-level drill-down.

Open question: verify results have no natural persisted artifact (unlike baselines, which are Parquet files with a durable listing via `/api/baselines`) — decide whether run history is purely in-memory/ephemeral (like `BaselineStatus` today) or worth persisting the last N runs' summaries to the state volume so a console restart doesn't lose the last result.

## Context

Surfaced 2026-06-30 while reviewing docs/verify.md — the doc's "See also" line pointing at the continuity badge was being mistaken for verify itself having a UI. Related: #640 (data-consistency guarantee epic, `verify` command + `--explain`), #613/PR #615 (baseline-trigger UI, the pattern this should mirror), #676 (baseline-trigger opt-in→opt-out default, same design question raised again here for a lower-risk feature).

nethalo:
closes #677

## Summary

Mirrors the Baseline-trigger pattern (#613/PR #615) for `bintrail verify`, giving the console a full lifecycle instead of only reading results a CLI/cron/CI run produced elsewhere:

- **`VerifyController`** (`internal/console/verify_trigger.go`): `Trigger`/`Status`/`Explain` interface, wired watch-only (`Config.VerifyCtrl`, nil on the standalone `serve` console) — mirrors `BaselineController`'s gating exactly.
- **`verifySupervisor`** (`cmd/bintrail-console/verify.go`): the concrete implementation. Fully self-sufficient like `baselineSupervisor` — opens its own index/source connections per run rather than reaching into the console's per-server bundle. One job at a time per server; per-table results are appended as each table completes (`internal/verify` has no progress callback of its own, so this loop *is* the "as they land" streaming). Both modes are supported:
  - **baseline-anchored** (default): `verify.FindBaselinePair` → `verify.VerifyBaselinePair` per table, sorted deterministically (the engine's own pair-building ranges a Go map). The "only one baseline" and "zero baselines" cases are told apart exactly like the CLI does.
  - **live-source**: `verify.VerifyTable` per table, needs the server's source DSN.
- **Explain**: on-demand only, never precomputed for every mismatch (`verify.ExplainBaselinePairMismatch` re-reconstructs). The supervisor caches the exact `BaselinePair` behind each mismatched table from the run that produced the verdict — re-deriving it later via a fresh `FindBaselinePair` could silently explain a *different* pair if a new baseline landed in between.
- **Endpoints**: `POST/GET /api/servers/{id}/verify`, `GET /api/servers/{id}/verify/explain?schema=&table=` (registry-servers-only, mirroring baseline's `requireMonitorEntry` gate). `GET /api/capabilities` gains `verify_trigger` (process-global) plus per-server `verify`/`verify_live_source`.
- **RBAC**: verify's engine (`verify.Config`/`verify.BaselineConfig`) carries no `DenyTables`/`RedactColumns` — like Time-travel's baseline reads, it bypasses redaction structurally. Both endpoints and the capability response explicitly refuse while an RBAC profile is active, since (unlike Reconstruct) verify doesn't route through the bundle's `baselineConfigured` gate that already collapses under a profile.
- **Frontend**: a new Verification panel on the Storage page — mode selector (live-source only offered when the server has a source DSN), a live-updating results list reusing the doctor-preflight card styling (match→pass, mismatch/error→fail, inconclusive→warn), and an Explain modal for mismatched rows.
- **Default**: on in the bundled compose stack (`VERIFY_TRIGGER`, opt-out — `docker-compose.yml`), same shape as `BASELINE_TRIGGER` after #683's flip. Unlike baseline creation, verify starts no subprocess and its default mode reads no live source, so it skips the opt-in-then-flip cycle. A bare `watch` invocation stays opt-in (`BINTRAIL_CONSOLE_VERIFY_TRIGGER=1`).

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean
- [x] `go test ./...` — full repo suite green
- [x] New unit tests: capability gating (including the RBAC-collapse case), 403/400/404/409/202 gating on trigger/status/explain, secret non-leak, pure-helper tests (`tableFilter`, `toWireResult`, `indexDBName`)
- [x] Live end-to-end smoke test against a real MySQL index (Docker): built both binaries, seeded a registry entry with a baseline dir, drove the Storage page with a real browser. Confirmed: the panel renders and gates correctly (live-source option hidden with no source DSN configured), clicking "Run verification" round-trips through the real HTTP API into a real background goroutine hitting `internal/verify` against a real index DB (exercised both the "no schema snapshot" and, after `bintrail snapshot`, the "no baselines found" failure paths — status polling and toast both reflected the real error), zero JS console errors throughout.
- [ ] **Not exercised against live data**: the match/mismatch card styling and the Explain modal, since that needs a real two-baseline mismatch fixture (out of scope to hand-roll for this pass — noting explicitly rather than claiming full coverage).

### DIFF
--- a/cmd/bintrail-console/verify.go
+++ b/cmd/bintrail-console/verify.go
@@ -1,0 +1,456 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"fmt"
+	"log/slog"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/go-sql-driver/mysql"
+
+	"github.com/dbtrail/dbtrail/internal/config"
+	"github.com/dbtrail/dbtrail/internal/console"
+	"github.com/dbtrail/dbtrail/internal/metadata"
+	"github.com/dbtrail/dbtrail/internal/parquetquery"
+	"github.com/dbtrail/dbtrail/internal/query"
+	"github.com/dbtrail/dbtrail/internal/verify"
+)
+
+// verifySupervisor implements console.VerifyController by running
+// internal/verify's engine IN-PROCESS, one table at a time in a background
+// goroutine — the same functions internal/cli/verify.go calls, looped here
+// instead of printed (#677). It is fully self-sufficient like
+// baselineSupervisor: it opens its own index/source connections per run
+// rather than reaching into the console's per-server bundle, so it needs
+// nothing from internal/console beyond the DTOs/interface it implements.
+//
+// One job at a time per server, tracked in-memory — like baseline, the
+// durable record (when there is one) is the baseline snapshots themselves;
+// a verify run's result has no artifact of its own, so a console restart
+// loses it (mirrors BaselineStatus's "idle if never run in this process").
+type verifySupervisor struct {
+	ctx context.Context // daemon lifecycle; cancels an in-flight run on shutdown
+
+	mu   sync.Mutex
+	jobs map[string]*verifyJob
+}
+
+// verifyJob is the mutable per-server job state: the pollable status PLUS the
+// bookkeeping Explain needs that must never reach the wire — the exact
+// BaselinePair each mismatched table was verified against, and enough of the
+// request to reopen a connection on demand. Re-deriving the pair via a fresh
+// FindBaselinePair at explain time would risk explaining a DIFFERENT pair
+// than the one the displayed verdict came from, if a new baseline landed
+// in between (see internal/verify.FindBaselinePair's doc comment).
+type verifyJob struct {
+	status console.VerifyStatus
+	mode   console.VerifyMode
+
+	indexDSN  string
+	noArchive bool
+
+	// pairs caches the BaselinePair behind every table this run reported as a
+	// mismatch, keyed "schema.table". Only populated for baseline-anchored
+	// runs (live-source has no explain support in the engine).
+	pairs map[string]verify.BaselinePair
+}
+
+// newVerifySupervisor builds a supervisor bound to the daemon context.
+func newVerifySupervisor(ctx context.Context) *verifySupervisor {
+	return &verifySupervisor{ctx: ctx, jobs: make(map[string]*verifyJob)}
+}
+
+// Trigger starts a verify run in the background; returns
+// console.ErrVerifyRunning if one is already in flight for this server.
+func (s *verifySupervisor) Trigger(req console.VerifyRequest) error {
+	s.mu.Lock()
+	if j, ok := s.jobs[req.ServerID]; ok && j.status.State == "running" {
+		s.mu.Unlock()
+		return console.ErrVerifyRunning
+	}
+	baselineSrc := req.BaselineDir
+	if baselineSrc == "" {
+		baselineSrc = req.BaselineS3
+	}
+	s.jobs[req.ServerID] = &verifyJob{
+		status:    console.VerifyStatus{State: "running", Mode: req.Mode, Since: nowStamp()},
+		mode:      req.Mode,
+		indexDSN:  req.IndexDSN,
+		noArchive: req.NoArchive,
+	}
+	s.mu.Unlock()
+
+	slog.Info("verify: starting in-process run", "server", req.ServerName, "id", req.ServerID, "mode", req.Mode)
+	go s.run(req, baselineSrc)
+	return nil
+}
+
+// Status returns a copy of the latest known run state (idle if never run here).
+func (s *verifySupervisor) Status(serverID string) console.VerifyStatus {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if j, ok := s.jobs[serverID]; ok {
+		return j.status
+	}
+	return console.VerifyStatus{State: "idle"}
+}
+
+// Explain re-runs the row-level drill-down for one table the last completed
+// baseline-anchored run reported as a mismatch, using the EXACT BaselinePair
+// that run verified — never a freshly re-derived one (see verifyJob's doc
+// comment). It opens its own short-lived index connection: the triggering
+// run's connection is already closed by the time an operator clicks Explain.
+func (s *verifySupervisor) Explain(serverID, schema, table string) (*console.VerifyExplanation, error) {
+	s.mu.Lock()
+	j, ok := s.jobs[serverID]
+	s.mu.Unlock()
+	if !ok || j.mode != console.VerifyModeBaselineAnchored {
+		return nil, console.ErrExplainUnavailable
+	}
+	pair, ok := j.pairs[schema+"."+table]
+	if !ok {
+		return nil, console.ErrExplainUnavailable
+	}
+
+	db, err := config.Connect(j.indexDSN)
+	if err != nil {
+		return nil, fmt.Errorf("connect index: %w", err)
+	}
+	defer db.Close()
+	resolver, err := metadata.NewResolver(db, 0)
+	if err != nil {
+		return nil, fmt.Errorf("load schema snapshot: %w", err)
+	}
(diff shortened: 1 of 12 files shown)


nethalo:
Ran a 3-agent review (code-reviewer + silent-failure-hunter + pr-test-analyzer, in parallel) against the initial commit. Two independent agents converged on the same critical bug; all four review findings are fixed in the follow-up commit:

**Critical (both agents caught it independently)**
- `Explain()` read the job's cached `BaselinePair` map after releasing the mutex, racing `cachePair`'s locked write from the still-running goroutine — reachable any time an operator clicks Explain on one mismatch while later tables are still verifying. Go treats concurrent map read/write as a fatal, unrecoverable error: this could kill the whole `watch` daemon (stream capture included), not just the request. Fixed by copying everything `Explain` needs out of the job inside the single critical section.
- The background `run()` goroutine had no panic recovery, unlike this codebase's own established convention for daemon-shared background work (`rotation.StartLoop`, the baseline-prune loop in `watch.go`). Added the same recover-and-log guard.

**Important**
- Live-source mode was admitted server-side with no baseline destination configured — `internal/verify.VerifyTable` still needs one and silently degrades every table to inconclusive, so a misconfigured live-source run would burn a full off-peak read of the live table for nothing. The CLI requires a baseline for both modes up front; the trigger endpoint now does too.
- `setNote` (the "only one baseline yet" no-op path) was setting `State` to succeeded itself, opening a narrow window where a second concurrent `Trigger` could be wrongly admitted before `run()`'s own terminal `finish()` call. `setNote` now only sets `Note`; `finish()` is the sole terminal-state transition.

**Also fixed**
- A malformed (non-empty) JSON trigger body's decode error was discarded, silently substituting the default mode/tables instead of 400ing.
- `pollVerify`'s poll loop now distinguishes a durable 403/404 (feature disabled, server deleted mid-run) from a transient blip, instead of retrying silently for the full ~20-minute cap.

Added the pr-test-analyzer's recommended zero-I/O unit tests for `verifySupervisor`'s lifecycle (Trigger collision, Explain's three refusal preconditions, appendResult/Summary accumulation, the setNote/finish state-ownership split) — this logic previously had zero coverage beyond the HTTP layer's stub-backed tests.

`go build/vet/test -race` clean across the whole repo after the fixes.

nethalo:
Ran the full `/pr-review-toolkit:review-pr` (comments + types, plus a re-verification pass of the first round's fixes by code-reviewer/silent-failure-hunter/pr-test-analyzer, 5 agents total). All fixes from the first round were independently confirmed correct — no regressions introduced. Two more real findings, both fixed in a follow-up commit:

- **`capabilitiesResponse.Verify`** was computed from `b.baselineSrc != ""` alone, but its comment claimed to mirror Reconstruct's gate (`b.baselineConfigured`, which also requires `!noArchive`). A no-archive server would've advertised `verify:true` and then reliably failed with a coverage-gap error on any window touching a rotated-out hour, since verify's own `query.FetchMerged` calls respect `NoArchive` too. Now uses `b.baselineConfigured` directly, genuinely matching Reconstruct.
- pr-test-analyzer confirmed empirically (by reverting the fix and re-running the suite) that the "live-source also needs a baseline destination" fix from the first round had no regression test — the existing suite passed even with the old, buggy per-mode-only validation restored. Added `TestVerifyTrigger_liveSourceRequiresBaselineDestination`.

Also documented the lock invariant directly on the `verifyJob` struct (every field is `mu`-protected) per the type-design-analyzer's suggestion, rather than leaving it scattered across individual method comments.

One process note for transparency: mid-review, the comment-analyzer agent transiently observed `verify.go` with the first round's `setNote` fix reverted, and (correctly, cautiously) flagged it rather than silently trusting a "this is fine" notice in its own tool output. I verified the committed file was never actually broken — the transient state was pr-test-analyzer (running concurrently in the same worktree) empirically reverting-then-restoring that exact code to prove the new test would catch a regression of it, which it does. No corruption occurred; confirmed via `git status`/`git diff` before and after.

`go build/vet/test -race` clean across the whole repo.

Issue #705, opened 2026-07-02
[forensics] Remote audit-log sources: RDS API (DownloadDBLogFilePortion) + CloudWatch Logs


nethalo:
Part of #701. Depends on #704 (parsers).

Port `agent/handler/rds_audit_log.go` (SaaS repo nethalo/dbtrail, ~560 LOC) and add the **CloudWatch Logs source the SaaS never built** — AWS itself recommends CWL over the file-download API because `DownloadDBLogFilePortion` consumes instance resources.

## Scope

- **RDS file API** (port): RDS-endpoint regex detection from the DSN host (`instance.xxx.region.rds.amazonaws.com`); `DescribeDBLogFiles` + `DownloadDBLogFilePortion` as a streaming `io.Reader` (Marker paging until `AdditionalDataPending=false`, 1MB-per-call cap, tail-mode via `NumberOfLines` on the first unmarked call, partial-trailing-line resume + de-dup); actionable error classification (missing `rds:DescribeDBLogFiles`/`rds:DownloadDBLogFilePortion` IAM actions, throttling, Aurora cluster-vs-instance id confusion, no credentials).
- **CloudWatch Logs source (new)**: log groups `/aws/rds/instance/{id}/audit` and `/aws/rds/cluster/{id}/audit` (one stream per Aurora instance) via `GetLogEvents`/`FilterLogEvents`; always correlate on the record-embedded timestamp, never CWL ingestion time (delivery lag is undocumented).
- Aurora file quirk: each instance stripes audit records across **4 concurrent files, 100MB rotation, non-sequential across files** ⇒ when reading files, fetch all + sort by timestamp (or steer users to CWL).
- Dependencies: add `aws-sdk-go-v2/service/rds` + `service/cloudwatchlogs` to go.mod (neither exists today). **Route all AWS config/region resolution through `storage.LoadAWSConfig`** (`internal/storage/s3url.go`) — it carries the #697 IMDS-region fix; do not call `awsconfig.LoadDefaultConfig` directly.
- Docs stub for the IAM policy (full guide is the docs child of #701): `rds:DescribeDBLogFiles`, `rds:DownloadDBLogFilePortion`, `logs:GetLogEvents`, `logs:FilterLogEvents`.

## Acceptance

- Port `rds_audit_log_test.go` (mocked SDK); unit tests for Marker pagination, partial-line resume, de-dup, tail-mode, and the CWL source; a live smoke against a real RDS instance can ride the e2e harness later.

PR #721, opened 2026-07-03
forensics: make who-changed on RDS/Aurora actually work (audit-tier wiring + full-scan Marker)

nethalo:
Fixes the two compounding defects from the post-merge review (`drafts/forensics-merge-review-2026-07-03.md`) that made the RDS/Aurora audit tier non-functional end-to-end. Together they gate the "who-changed on RDS/Aurora via audit logs" capability #705 shipped but that never actually ran.

## What was broken

**Blocker — full scan silently returned only the tail** (`internal/forensics/auditlog_rds.go`). `rdsLogReader.Read` never sent `Marker="0"` on the first `DownloadDBLogFilePortion` call of a non-tail read. Per the AWS API, a call with neither `Marker` nor `NumberOfLines` returns only the most-recent ~10000 lines / 1 MB (the tail). who-changed's audit read is a full scan (`auditReadOptionsFor` sets `TailLines:-1`), so any >1 MB RDS/Aurora audit log was read from the end only — older in-window records dropped with no warning, a confidently-wrong attribution on a normal input.

**High — the remote readers were dead code** (`internal/forensics/whochanged.go:436` + agent). No production caller set `AuditReadOptions.SourceHost`, so `rdsHostCandidate("")` was always false: the `AuditSourceAuto` local→RDS fallback never fired and the RDS-file-API + CloudWatch readers (~1080 LOC) were never reached.

## The fix

1. **Marker fix:** the first non-tail page now sends `Marker="0"` (start of file) and keeps following `resp.Marker`; tail mode is untouched (still `NumberOfLines`, single page).
2. **Wiring:** thread the source host through every surface that reaches the audit tier —
   - `WhoChangedDeps` gains `SourceHost`; `auditReadOptionsFor` sets it on the read; the who-changed CLI derives it from `--source-dsn`.
   - the agent gains `DefaultHandler.SourceHost` (from `--source-dsn`) and `ForensicsAuditLogRequest.{Source,SourceHost,CloudWatchLogGroup}`, so a BYOS agent on RDS/Aurora reaches the remote source without the caller supplying the endpoint; a per-request `SourceHost` overrides.

Local / non-RDS sources are unaffected: `rdsHostCandidate` matches only real RDS endpoints, and the RDS fallback is gated on `Source==Auto` **and** an RDS-shaped host **and** the local file being absent.

## Tests

- `TestRDSLogReader_FullScanStartsFromBeginning` — a mock that emulates the real AWS Marker semantics (no marker ⇒ tail-only). **Verified it fails on the pre-fix reader** (`Marker=<nil>` + oldest record dropped) and passes after.
- `TestAuditReadOptionsFor` — asserts the source host is threaded into `AuditReadOptions` (closes the dead-code gap at the engine).
- `TestResolveAuditSourceHost` — pins the agent's request-override-else-handler-default fallback so a refactor can't silently re-break the wiring.
- `go build ./... && go vet ./...` clean, full unit suite + forensics/agent/cli integration green against a live MySQL container, gofmt clean.

Reviewed by an adversarial pass over the diff: no blocker/high/medium defect introduced; tail mode and pagination-resume preserved; no wrong routing for local sources.

Deployment shapes tested: local MySQL (unaffected) + the RDS reader path via mocked AWS clients. **Hold on the RDS/Aurora claim is lifted once this lands** (per the review verdict).


nethalo:
## Verified against real AWS RDS 🔬

Ran a read-only probe against a live 50 MB MariaDB audit log on a real RDS instance (`us-west-2`, IAM creds holding `rds:DescribeDBLogFiles` + `rds:DownloadDBLogFilePortion`) — nothing provisioned, no instance touched beyond log reads. This validates the one thing the mocks can't: that the AWS API actually behaves as the fix assumes.

**Raw `DownloadDBLogFilePortion` on a 50 MB `server_audit.log` file, two ways:**

| Call | Bytes returned | First record timestamp | AdditionalDataPending |
|---|---|---|---|
| **no Marker, no NumberOfLines** (pre-fix full-scan) | 1,479,776 (~1.5 MB) | `20260703 12:25:59` | none (stops) |
| **`Marker="0"`** (post-fix) | 1,048,611 (~1 MB) | `20260703 11:58:41` | **true** + next marker |

The no-Marker read returned only the **tail** — it silently skipped the first **~27 minutes** of the file (11:58→12:25). Any who-changed query for an event in that window would have found nothing in the audit tier pre-fix. `Marker="0"` reads from the start and signals more pages, so the reader paginates the whole file. This confirms the blocker's premise (`no Marker ⇒ tail-only`) on real AWS.

**The actual `rdsLogReader` production code against the same 50 MB file:**

```
full-scan bytes read: 49,997,875           (the WHOLE file, ~50 paginated 1 MB calls)
first line: 20260703 11:58:41,...,QUERY,sbtest,'SELECT c FROM sb...   (start of file, not the 12:25 tail)
--- PASS: TestRealAWS_RDSReaderFullScanFromStart (47.63s)
```

So the fixed reader reads the entire file from the beginning and the Marker-follow pagination works end-to-end over a real multi-page RDS audit log (the real MariaDB `server_audit` format parses cleanly too). The throwaway `//go:build realaws` harness that produced this was run against `abi-test` and **not committed** (environment-specific).

Net: both the API-semantics assumption and the production reader code are confirmed correct on real RDS.

nethalo:
## Addressed the multi-agent PR review (`/pr-review-toolkit:review-pr`)

Ran the 5-specialist review (code / silent-failure / comments / tests / type-design). Verdict was clean on the core fix; the substantive items are now addressed in the follow-up commit:

**Substantive — silent-failure (medium):** `rdsLogReader.Read` treated an empty chunk as EOF *before* consulting `AdditionalDataPending`. An actively-written audit log can return an empty portion mid-file with `AdditionalDataPending=true`, so the old code would silently truncate the very full scan this PR makes load-bearing. Termination is now driven by `AdditionalDataPending`: an empty-but-pending page follows the marker; a pending response with **no** marker, or a **non-advancing** marker, errors instead of looping. Added 3 tests (`EmptyPageWithPendingContinues`, `PendingWithoutMarkerErrors`, `PendingNonAdvancingMarkerErrors`). This is the same silent-truncation class as the headline Marker fix, so it belongs here.

**Doc accuracy (code-reviewer + comment-analyzer):** the `HandleForensicsAuditLog` doc still said "local files only — remote RDS/CloudWatch out of scope", the exact opposite of what the PR does. Rewritten. Also dropped a CloudWatch over-claim from the three *who-changed auto-path* comments (`WhoChangedDeps.SourceHost`, `auditReadOptionsFor`, the CLI): auto mode only reaches the RDS file API — CloudWatch requires an explicit `Source="cloudwatch"` on the agent request.

**Host canonicalization (type + test + silent-failure all flagged):** the CLI stored `host:port` while the agent stored a bare host. Both normalize downstream, but a future raw reader could trip on the `:port`. The CLI now derives a bare host via `config.ParseSourceDSN`, matching the agent — one canonical value.

**Test gap (test-analyzer):** added `TestHandleForensicsAuditLog_ThreadsSource` pinning that the handler threads `req.Source` into the read options.

### Real-AWS re-verification after the reader restructure
Re-ran the restructured reader against a live 50 MB RDS audit log: read all **49,997,774 bytes** from the file's start; a no-Marker read of the same file returned only the tail starting **27 minutes later** (`12:53` vs the file start `12:26`). Pagination across ~50 pages intact; blocker premise still holds on real AWS.

`go build`/`vet`/full unit suite/forensics+agent+cli integration all green; my changed files gofmt-clean.

Issue #597, opened 2026-06-23
Epic: PostgreSQL-as-source (beta → GA)


nethalo:
Tracking issue for taking **PostgreSQL-as-source from beta → GA**. Successor to the alpha→beta epic #527, which reached **beta** (all 6 beta-BLOCKING data-safety gates closed; the silent-loss classes surfaced loud).

bintrail is a data-*recovery* tool, so the beta bar was *faithful capture + silent-data-loss classes closed + advertised surfaces covered + safe to point at a real DB*. The **GA bar** adds: *full recovery surface (reconstruct/time-travel), validated on managed PostgreSQL, with source-aware operator UX*.

## Where beta landed (context)

`recover` (delta-only) is the supported recovery surface; capture is type-faithful (#533), RI-FULL-enforced (#531), slot/WAL-monitored (#532), DDL-drift-safe (#591), and the silent-loss guards (#555 UNLOGGED, #556 cascade-child, #559 TimescaleDB, TOAST resolve-at-decode) are in. See #527 for the full beta record.

## GA gates — **ALL CLOSED as of 2026-08-02**

- [x] #593 — **Baseline at slot `consistent_point` LSN → reconstruct / time-travel for PG.** ✅ CLOSED.
- [x] #594 — **Managed-PostgreSQL smoke matrix** (RDS / Aurora / Cloud SQL / Azure). ✅ CLOSED (`managed-pg-smoke.yml` in CI).
- [x] #603 — **`SHOW TABLES FROM _flashback` returns only one table for PG.** ✅ CLOSED.
- [x] #595 — **Console source-aware presentation.** ✅ Shipped — subset via PR #598; live slot-health / RI-FULL signals via #599 → PR #604 (v0.20.1).
- [x] #558 — **Sequence cursor resume.** ✅ CLOSED.
- [x] #535 — **ICP survey.** CLOSED as NOT PLANNED — the product gate was discarded, not executed; downstream decisions (forensics attribution, managed-matrix breadth) must be taken without survey data.

## Smaller GA-bucket follow-ups (hardening / completeness; not launch-gating)

- [x] #592 — TOAST read-side fail-loud. ✅ CLOSED completed (the earlier author deferral was reversed; the guard is in).
- [x] #533 follow-up (closed parent) — empirical CI coverage of extension types. **Extracted to #1210** (still missing; tracked there).
[...]

PR #722, opened 2026-07-03
docs(pg): sequences after recovery — setval(MAX+1) recipe (#558)

### LINKED ISSUE #558: [pg] Sequence cursor not restored by reconstruct — document setval(MAX+1) + optional synthesis
**Finding from the 2026-06-21 source-semantics discussion (verified vs PG primary sources).**

Logical decoding does **not** replicate sequence advances. PG17 docs, verbatim: *"Sequence data is not replicated. The data in serial or identity columns backed by sequences will of course be replicated as part of the table, but the sequence itself would still show the start value on the subscriber."* (The "logical decoding of sequences" feature was reverted from PG15 — commit `2c7ea57` — and is absent through PG18.)

## Impact
- **Row DATA is fully captured**: the materialized `id` rides in `row_after`. `recover`/`reconstruct` restore row contents correctly. **No capture loss.**
- The sequence's `last_value` is a **separate catalog object we never see**. After a `reconstruct`/restore, the next `nextval` can collide with restored ids (the classic dump/restore gotcha). Same boundary as MySQL `AUTO_INCREMENT`.

## What to do
- **Docs (floor)**: document the post-reconstruct step `SELECT setval('seq', (SELECT MAX(id) FROM t))` for SERIAL/IDENTITY columns.
- **Optional**: synthesize `setval(...)` in `reconstruct --output-format` from `MAX(captured id)` for OWNED-BY sequences — the analog of mysqldump's `AUTO_INCREMENT=N`. Re-derive from MAX (robust); do NOT thread the baseline's stale `setval`.
- **Out of scope (document)**: a standalone sequence whose values never land in a captured table — never entered our history, irrecoverable.

## Refs
- Epic #527; docs
- `drafts/postgres-sequences-globals-extensions-2026-06-21.md`

nethalo:
Closes #558.

## What

Adds a dedicated **"Sequences after recovery"** section to `docs/postgres.md` (placed right after *Querying and recovering*), and re-points the existing beta-limitations bullet at it. Docs-only — no code changed.

The section covers, DBA-first:

- **Why**: logical decoding does not replicate sequence state (PG docs verbatim; the PG15 sequences-decoding feature was reverted — `2c7ea57` — and is absent through PG18). Row ids ARE fully captured; the sequence's `last_value` is a separate catalog object bintrail never sees. Same boundary as MySQL `AUTO_INCREMENT`.
- **When it matters** (verified, not hand-waved): fresh/restored targets and reset sequences (`TRUNCATE … RESTART IDENTITY`, manual `setval`) → always; applying `recover` output on the same live database where the delete happened → usually not needed, because PostgreSQL explicit-id INSERTs (which is what recover emits, incl. `OVERRIDING SYSTEM VALUE`) never advance the sequence AND the sequence already advanced past those ids at original insert time and never rewinds. "Cheap and harmless, run it when in doubt."
- **The recipe**: per-column `setval(pg_get_serial_sequence('t','id'), COALESCE(MAX(id),0)+1, false)` (empty-table safe, exact `MAX+1` next value), a note on multi-sequence tables and non-atomic MAX/setval under concurrent writers, plus a catalog-driven generator (`pg_depend` deptype `'a'`/`'i'` = SERIAL + IDENTITY) that emits the statements for a whole schema.
- **Out of scope, stated honestly**: standalone sequences whose values never land in a captured table are irrecoverable — they never entered the history; `pg_dump` records sequence state, snapshot them by other means.

## Why the optional setval synthesis is NOT included

The issue marks synthesizing `setval` into `reconstruct --output-format` output as optional. It would be dead code today:

- Full-table reconstruct is gated on a baseline (`FindBaseline` → `ErrNoBaseline` fails clean), and no PostgreSQL baseline producer exists yet — that is #593 in the GA epic #597 (`bintrail-pg` ships only `stream`/`doctor`/`reset`).
- The full-table output path (`internal/reconstruct/fulltable.go`) is mydumper/MySQL-dialect throughout; there is no PG-dialect restore artifact to attach a `setval` to.

Synthesis is deferred to #593's un-gating; the docs floor closes the issue as specified.

## Anchor check

New heading `## Sequences after recovery` → `#sequences-after-recovery`, linked from the beta-limitations bullet. No existing headings were retitled; all pre-existing anchors (`#querying-and-recovering`, `#type-support`, `#extensions-and-custom-types`, `#managed-postgresql`, `#5-the-replication-slot--created-for-you`) unchanged.


nethalo:
Validated the `pg_depend` generator against a live PostgreSQL 16 (`wal_level=logical` test container) — the flagged concern is cleared:

- SERIAL column, rows with MAX(id)=9 → generated `setval` set 10, `nextval` returned exactly 10 (MAX+1) ✓
- IDENTITY (GENERATED ALWAYS) with MAX=42 → next 43 ✓
- Empty table → `COALESCE(...,0)+1` → next value 1 ✓
- Quoted identifiers (`"Weird Table"."Odd Id"` SERIAL) → `format(%L/%I)` quoting held, next 6 ✓
- Standalone sequence in the same schema correctly NOT touched by the generator (`deptype IN ('a','i')` filter) ✓

Ran via `\gexec` exactly as the doc instructs.

PR #725, opened 2026-07-03
pg baseline: LSN anchor in Parquet metadata + flavor-aware baseline gap check (#593 slices A+B)

### LINKED ISSUE #593: [pg] Baseline at slot consistent_point LSN → enable reconstruct/time-travel for PostgreSQL (GA)

nethalo:
Part of #593 (slice A+B of the baseline-LSN blueprint). Does **not** close #593 — slices C/D (writing the anchor at dump time, reconstruct/verify enablement) follow.

## Slice A — LSN anchor in baseline metadata (pure-additive)

- New Parquet key-value metadata key `MetaKeyLSN = "bintrail.baseline_lsn"`: the decimal string of the uint64 WAL LSN at which a PostgreSQL-source baseline is consistent (the slot's consistent_point).
- Parsed by `ReadParquetMetadata` (local; warn-and-zero on a corrupt value, mirroring the `BinlogPos` branch) and by `ReadParquetMetadataAny`'s S3 `parquet_kv_metadata()` loop.
- Surfaced as `DumpMetadata.LSN` and `BaselineInfo.LSN` (populated by `DiscoverBaselines`).
- Zero behavior change for MySQL/MariaDB baselines — the key is simply absent, `LSN` stays 0.

## Slice B — flavor-aware baseline↔first-event gap check

The reconstruct CLI warned about a gap between the baseline anchor and the first indexed event with a MySQL-only compare on `binlog_file`/`start_pos`. For a PG source, `binlog_file` holds LSN **text** and lexical order is wrong:

> baseline at LSN `0/9`, first event at LSN `0/10` (numerically **after** — a real gap). Lexically `"0/10" < "0/9"`, so the old compare reports **no gap**. The mirror case would report a phantom gap.

PG events carry the numeric commit LSN in `StartPos` (`internal/pgcapture` `rowEvent`), so the fix compares numerics against the baseline's `MetaKeyLSN` anchor:

- `reconstruct.GapDetected(flavor, eventFile, eventStartPos, baselineFile, baselinePos, baselineLSN)` (new pure helper, `internal/reconstruct/gap.go`): flavor `"postgres"` → `eventStartPos > baselineLSN`, numeric only; anything else (mysql/mariadb/`""`) → the existing two-key file+pos compare, unchanged.
- Wired into `internal/cli/reconstruct.go`, replacing the inline comparison. The anchor-present / event-position-missing guards are flavor-aware too (PG anchor = `LSN != 0`; PG event position missing = `StartPos == 0`, since 0 is not a valid WAL position).

### `SourceFlavor` placement

Grep found exactly one existing copy of `SELECT flavor FROM stream_state WHERE id = 1` — inside `recovery.DialectForIndex`, which returns a `Dialect`, not the raw flavor. Rather than add a second copy, the read is **hoisted** to `query.SourceFlavor(db) string` (`internal/query/flavor.go`, nil-safe/best-effort → `""`), and `recovery.DialectForIndex` now delegates: `DialectForFlavor(query.SourceFlavor(db))`. No import cycle (`recovery` already imports `query`; `query` does not import `recovery`). One SELECT, two consumers.

### `baselineLSN == 0` semantics (pre-slice-A PG baseline)

Chosen: **unknown anchor → do not flag a gap; log the skip.** Rationale (documented on the helper): this check gates a *warning* only — it never blocks or alters the reconstruction — so "no flag" can't silently produce wrong data. It mirrors how the surrounding code already treats missing coordinates (the existing `baseline lacks position metadata` info-skip branch, and `#318`'s NULL-binlog_file warn-skip); failing closed on unknown is reserved for the explicitly-opted-in `status --fail-on-gap`. Flagging conservatively would instead emit a phantom "reconstruction may be incomplete" warning on every pre-existing PG baseline. The CLI logs an explicit skip (`gap detection skipped — baseline lacks position metadata`, with the flavor attr) so nothing is silent.

### Other gap-check call sites

Grep-audited `internal/reconstruct/fulltable.go`, `internal/verify`, `internal/shim`, `internal/console/reconstruct.go` (and repo-wide for `first.BinlogFile` / file+pos compares): the blueprint's finding holds — **the CLI reconstruct warning is the only baseline-vs-first-event coordinate comparison.** The `AllowGaps`/`query.GapError` machinery in shim/console/fulltable is the *planner's* coverage-gap check (rotated-but-unarchived hours), not a coordinate compare, and `verify`'s `query.BinlogPos`/`UntilPos` is a SQL upper bound on the baseline-anchored (MySQL-only today) path — both untouched.

## Tests

- Slice A: `TestReadParquetMetadata_LSN` (writer→reader round-trip via `NewWriter` metadata, same pattern as the existing metadata tests), `TestReadParquetMetadata_corruptLSN` (warn-and-zero, other keys unaffected), `TestDiscoverBaselines_LSN` + absent-key assertions on the existing MySQL-baseline tests.
- Slice B: table-driven `TestGapDetected` — MySQL two-key cases (same-file pos, file boundary, at-anchor), MariaDB + empty flavor fall back to MySQL semantics (and ignore a stray `baselineLSN`), PG numeric cases including **both directions of the lexical trap** (`0/10` vs `0/9`), and the `baselineLSN==0` edge. `TestSourceFlavorNilDB` pins the nil-guard contract `DialectForIndex` now relies on.
- `go build ./...`, `go vet ./...`, `go test ./... -count=1`: **all pass**.
- Integration (MySQL 13306): `go test -tags integration ./internal/baseline/... ./internal/reconstruct/... ./internal/query/... ./internal/recovery/... -count=1`: **all pass** (covers the `DialectForIndex` surfaces affected by the hoist).


nethalo:
Pushed dbe68b6 addressing the four review findings (additive — `GapDetected`/`SourceFlavor` signatures untouched for the stacked Slice C branch):

1. `query.SourceFlavor` now logs instead of silently flattening to `""`: `sql.ErrNoRows` → Debug (expected file-indexed/pre-stream shape); any other error → Warn with the error. Covers both the gap check and `DialectForIndex`.
2. `"flavor"` attr added to the gap warning (and the skip lines).
3. PG baseline without an LSN anchor now **WARNs** with an honest message ("this baseline predates LSN anchoring … a gap would go undetected") — no longer an INFO whisper, and no longer recommends `bintrail baseline`, which can't produce a PG anchor today. MySQL missing-anchor keeps its pre-existing Info level.
4. LSN-lineage contradiction guard: `flavor==""` + `bmeta.LSN!=0` → postgres semantics forced (with a Warn) — the lexical-on-LSN-text path is now structurally unreachable, not unreachable-by-convention.

The CLI glue decisions are extracted into `resolveGapCheck` and pinned by a table-driven test (guard case, PG anchor/position edges, #318 case, mariadb/empty-flavor fallbacks).

Re-ran: `go vet ./...`, `go test ./... -count=1`, and `-tags integration` for baseline/reconstruct/query/recovery/cli against MySQL 13306 — all pass.

Issue #633, opened 2026-06-27
Baseline: persist row-count + content digest at snapshot GTID, fail-loud on source mismatch


nethalo:
Part of #631. Depends on #632 (`consistency.Hasher` / `ConsistentTableChecksum`).

## Reshaped scope (file-based)

The `baseline` command operates **purely on files** — mydumper (a separate `dump` step) writes the dump + a metadata file with the snapshot GTID, and `baseline` converts those to Parquet with no live source connection. So the original "call consistentTableChecksum against the source at the dump's consistent snapshot" is **not achievable here**: a baseline-time live re-read is a *different* snapshot than mydumper's (the forbidden now-vs-now), would cry wolf on any active DB, and would duplicate #634. **The live source comparison stays in #634.** #633 persists the fingerprint and locks its canonical form.

## What

1. During the existing conversion pass, **tap the parser's row values** through the shared `consistency.Hasher` (mydumper dumps MySQL's text rendering — the same form `ConsistentTableChecksum` reads via the text protocol), and persist per-table **`bintrail.baseline_row_count`** + **`bintrail.baseline_content_digest`** into the Parquet key-value metadata (alongside the existing GTID), via the parquet-go writer's `SetKeyValueMetadata` before `Close`. No second pass; no DuckDB read-back (which would hit the #496 int>2^53 rounding); no Parquet→text renderer (that is #634's tool).
2. **Keystone:** the persisted digest is **byte-identical** to a live `ConsistentTableChecksum` of the same rows — locked by an integration test across unsigned>2^63, decimal, DATETIME(6), DATE, utf8mb4, NULL. That identity is what makes #634's compare meaningful.
3. `ReadParquetMetadata` (local + S3) reads the new keys back.

## Fix surfaced by the keystone (a real #632 bug)

With `parseTime=true` (which `config.Connect` and the test DSNs set), the MySQL driver decoded DATE/DATETIME/TIMESTAMP into `time.Time` and re-rendered them as **RFC3339** (`2021-01-01T00:00:00Z`) on a `RawBytes` scan — not MySQL's native text (`2021-01-01 00:00:00`) that mydumper dumps and the binlog carries. `ConsistentTableChecksum` now `CAST`s those columns `AS CHAR`, making the canonical form MySQL-native and **parseTime-independent**. (#632's own tests were unaffected — they compared #632 to itself.)

## Not in scope

[...]

Issue #557, opened 2026-06-21
[pg] recover: GENERATED ALWAYS AS IDENTITY needs OVERRIDING SYSTEM VALUE; omit STORED generated columns


nethalo:
**Finding from the 2026-06-21 source-semantics discussion — recover-path correctness on PG (bridges the sequence answer; bites the spike's own Part C output).**

Two PG-specific bugs in reversal-SQL generation:

1. **`GENERATED ALWAYS AS IDENTITY` rejects an explicit value.** A reverse-DELETE `INSERT INTO t (id,…) VALUES (12345,…)` **errors** on PG unless it emits `OVERRIDING SYSTEM VALUE`. MySQL AUTO_INCREMENT accepts explicit values freely → a real divergence. The spike's `reverse DELETE → INSERT INTO public.t_full (…id…)` would **fail** against a real identity-PK target.
2. **STORED generated columns** must be **omitted** from reverse-INSERT/UPDATE (PG rejects writing them). They are also **absent from the `pgoutput` stream in PG14–17** (`publish_generated_columns` is PG18+) → not in the row images. Fine for recovery (recomputed on insert), but a **visible gap for time-travel/query** that must be documented.

## What to do
PG-aware reversal SQL generation: emit `OVERRIDING SYSTEM VALUE` when the PK/target column is `GENERATED ALWAYS AS IDENTITY`; skip STORED generated columns in the column list. Needs the column-metadata oracle (identity/generated kind), available from the catalog / the type info #530 retains.

## Refs
- Extends #531 (PK-aware recovery) / #533 (rendering); epic #527
- Spike Part C: `drafts/postgres-spike-results-2026-06-20.md`
- `drafts/postgres-sequences-globals-extensions-2026-06-21.md`

PR #729, opened 2026-07-03
pg baseline: producer core — snapshot COPY into LSN-anchored Parquet (#593 slice C)

nethalo:
> **Stacked on #725 — do not merge before it; retarget to main after #725 merges.** This branch builds on `feat/593-slice-ab-lsn-anchor-gap-check` (needs slice A's `baseline.MetaKeyLSN`).

Part of #593 (does not close it). Slice C of the blueprint: the PostgreSQL baseline producer.

## What

New capture-side package **`internal/pgbaseline`** + **`bintrail-pg baseline`** command: takes a consistent baseline directly from a live PostgreSQL source (no dump step) and writes one Parquet file per published table under `<output>/<timestamp>/<schema>/<table>.parquet`, mirroring `internal/baseline.Run`'s marker/manifest discipline exactly (`_INCOMPLETE` before any conversion; integrity manifest + `_SUCCESS` only on full success).

## How

- **Ordering invariant**: the replication slot is ensured to exist **before** the snapshot transaction opens, so `slot consistent_point ≤ baseline anchor LSN` — no delta between them can be missed; the overlap is redelivered and is harmless because reconstruct's merge is last-write-wins idempotent. The integration test asserts `confirmed_flush_lsn ≤ anchor` explicitly.
- **Anchor**: `SELECT pg_current_wal_lsn(), now()` is the FIRST statement of a `REPEATABLE READ, READ ONLY` transaction, fixing (anchor LSN, snapshot time, visible data) atomically. Embedded per table as `baseline.MetaKeyLSN` (decimal uint64, slice A's contract). `CREATE_REPLICATION_SLOT`'s snapshot export is deliberately not used — it only exists at slot creation, and this design makes first-run and re-baseline the same code path (re-baseline needs no replication DSN at all).
- **Parallelism**: `pg_export_snapshot()` + one connection per worker adopting the snapshot via `SET TRANSACTION SNAPSHOT`; the anchor transaction stays open until all workers finish.
- **Slot seam**: `pgcapture.EnsureSlotExists` exports the existing TOCTOU-safe create logic (42710 → success; `wal_status=lost` → fail loud wrapping `ErrSlotLost`) with a lazy replication connect. Missing slot + no `--repl-dsn` = actionable fatal error, never a silent skip.
- **Values are raw PostgreSQL text**: new `baseline.Column.RawText` — unconditional optional-string Parquet column, values stored verbatim (`convertValue` short-circuits before the MySQL type switch; zero change to MySQL paths). This matches the pgoutput text rendering the delta path indexes, so the PK join stays an identity string match. No type conversion.
- **COPY TEXT parser is hand-rolled** (not encoding/csv): tab-split + the seven escapes `COPY TO` emits (`\b \f \n \r \t \v \\`), `\N` NULL distinguishable from empty string and from a literal `\N` (`\\N`). Unknown escapes / lone trailing backslash fail loud.
- **Discovery from `pg_publication_tables`** inside the snapshot transaction — the same view that defines what pgoutput streams, so partitioned-table naming (parent under `publish_via_partition_root=true`, leaves otherwise) matches the delta path by construction. Loud warnings on partition parents/leaves that flipping `pubviaroot` after the baseline silently breaks the merge join. `COPY (SELECT <cols> …)` is used (a partitioned parent rejects direct `COPY <table>`); identifiers quoted via `pgx.Identifier.Sanitize()`.
- **Generated-columns decision**: `GENERATED ALWAYS … STORED` columns are **excluded** (catalog filter `attgenerated <> ''`, logged per table). pgoutput on PG 14–17 never publishes them, so delta row images don't carry them; had the baseline included one, every touched row would lose it after the last-write-wins merge while untouched rows kept it — a silently inconsistent column set. PostgreSQL recomputes them on re-INSERT, and recovery already omits them (#557). Dropped columns excluded likewise.
- `MetaKeyCreateTableSQL` is deliberately absent (mydumper full-table reconstruct only; no pg_dump subprocess). Content digest is not stamped (its hasher is the MySQL live-checksum contract, #633); at-rest integrity is covered by the CRC-32C manifest.
- **CLI**: `bintrail-pg baseline --query-dsn … [--repl-dsn …] --slot --publication --output [--schemas/--tables] [--compression] [--row-group-size] [--parallelism] [--retry] [--upload s3://… --upload-region …]`, with `BINTRAIL_PG_*` env fallbacks exactly like `stream`; `--upload` reuses `internal/baseline.Upload` unchanged.

## Boundary

`internal/pgbaseline` links pgx/pglogrepl (capture side) and imports `internal/baseline` for Writer/markers/keys — the dependency never points the other way. `internal/event.TestReadLayerDoesNotLinkGoMySQL` passes; `go list -deps ./internal/baseline` links zero `jackc/*` packages.

## Tests

- Unit: COPY TEXT escape table (tabs/newlines/backslashes, `\N` vs empty vs literal `\N`, multibyte UTF-8, error cases), byte-at-a-time chunked sink reassembly; RawText writer round-trip (optional-string schema, verbatim values, NULL vs empty preserved).
- Integration (live PG 16, `//go:build integration`, `BINTRAIL_TEST_PG_DSN`): contents byte-exact including an out-of-line TOASTed value; `MetaKeyLSN` present, equal to `Stats.AnchorLSN`, ≤ `pg_current_wal_lsn()`; **slot `confirmed_flush_lsn` ≤ anchor (ordering invariant)**; `_SUCCESS` + manifest validation per file; generated-column exclusion; re-baseline over the existing slot with an empty `ReplDSN` (parallel path exercised on the first run, serial on the second); missing-slot and empty-publication fail-loud guards. All 3 pass locally; `go test ./... -count=1`, `go vet ./...`, and the read-layer link guard are green.


nethalo:
Review round applied in 6c921f1 — all findings addressed, nothing deferred except noted below.

**Blockers**
1. **Worker-pool deadlock** — work channel now buffered to `len(tables)`, fed and closed before any worker starts; an early-exiting worker (cancel, conn failure) can no longer wedge the feeder. `openWorkerConn` failures are logged at append time. Pinned by `TestPGBaseline_CancelMidRun_Integration` (5 tables / 2 workers / cancel mid-COPY → Run returns promptly, snapshot stays `_INCOMPLETE`, no `_SUCCESS`).
2. **Server row-count check** — `CopyTo`'s command tag `RowsAffected()` is compared against the parsed row count per table; mismatch fails loud. `Flush` now also **rejects** an unterminated trailing line as a protocol anomaly instead of accepting it as a row.

**Mediums**
3. `--retry` local skip **dropped** (unreachable + dangerous); the flag now only drives the S3 upload skip and its help text says so. Skips no longer counted anywhere.
4. `_INCOMPLETE` write failure is now **fatal** in both producers — `pgbaseline.Run` and the mydumper `baseline.Run` (`markIncomplete` removed) — closing the #467-class hole where an ENOSPC'd marker + uncatchable kill published a markerless partial snapshot.
5. `EnsureSlotExists` slot identity **scoped**: `queryScopedSlotState` requires `slot_type='logical' AND plugin='pgoutput' AND database=current_database()`; anything else under the same name is a loud actionable error. The 42710 race branch re-checks with the scoped query.

**New guarantee tests** (all green against live PG 16)
6. **Snapshot/delta boundary** via a test seam after the anchor read: a row committed post-anchor is absent from the baseline Parquet and present in the slot's deltas. One finding worth reviewer attention: the row's own WAL record can start **exactly at** the anchor (`pg_current_wal_lsn()` is the next-write position), so the test asserts the load-bearing form of the invariant — the transaction's **COMMIT record is strictly > anchor**, which is the LSN pgstreamrun stamps into indexed events. The row-record assertion is `>=`.
7. **Identity-join byte-equality**: bytea/timestamptz/numeric/bool/float8 rows byte-compared per column between baseline raw text and the real pgoutput decoder's output (`pgcapture.Capturer` end to end).
8. **Mid-run table failure** (restricted role, one table without SELECT): error names the table; no `_SUCCESS`, no manifest, `SnapshotComplete()==false` (the exact predicate discovery uses), partial file removed.
9. **Partitioned tables**: 2-partition parent under `publish_via_partition_root=true` (one parent-named file, 5 rows, parent warning) and `=false` (per-leaf files 2/3 rows, leaf warning). Plus `FuzzParseCopyTextLine` (round-trip vs a COPY-TO escaping oracle; 2.4M execs clean in 15s).

**Lows**: parse errors truncate quoted row content to 80 bytes; slot-seam connection-ownership doc fixed; `loadColumns` documents why PG15+ publication column lists cannot diverge under the RI FULL requirement.

**Results**: `go vet ./...` clean; full unit suite green; all 8 pgbaseline integration tests + pgcapture integration regression green against `localhost:15432`; read-layer link guard green.

PR #861, opened 2026-07-05
baseline: exclude GENERATED columns from ParseSchema, fail loud on arity mismatch

### LINKED ISSUE #767: baseline: SQL reader discards the INSERT column-list — GENERATED columns shift every row (silent corruption)
`ReadSQLFile` locates ` VALUES` and discards everything before it: the INSERT's column-list is never reconciled with `ParseSchema`. mydumper (and mysqldump) EXCLUDE GENERATED columns (STORED/VIRTUAL) from the column-list and values, but `ParseSchema` (`schema.go colRe`) INCLUDES them — so the positional mapping in `WriteRow` runs shifted.

Location: `internal/baseline/reader_sql.go:65`

Scenario: `orders(id INT, price DECIMAL, total DECIMAL GENERATED ... STORED, note VARCHAR)` → the value of `note` lands in `total` and `note` becomes NULL silently (`WriteRow` NULL-pads the missing tail, `writer.go:116`); if neighboring types differ it aborts with a confusing conversion error, if they coincide (varchar/varchar) the baseline is silently CORRUPT and feeds reconstruct/_snapshot/verify. The content-digest (#633) does NOT catch it: it hashes the same shifted values.

Fix: parse the INSERT column-list and map by name against `ParseSchema` (or at least exclude GENERATED in `ParseSchema` + validate arity). Related in spirit to #502 (` VALUES` substring) but a distinct, unfiled class.

---

nethalo:
closes #767

## Summary
- `ParseSchema` (`internal/baseline/schema.go`) previously included STORED/VIRTUAL generated columns in the positional column list it returns, but mydumper/mysqldump both **exclude** generated columns from the INSERT column-list and VALUES tuples they emit. Every column declared after a generated one landed in the wrong Parquet slot in `WriteRow` (or NULLed the trailing column) with no error — silent baseline corruption feeding `reconstruct`/`_snapshot`/`verify` downstream.
- `ParseSchema` now skips generated columns, detecting both MySQL's canonical `GENERATED ALWAYS AS (...) STORED`/`VIRTUAL` and MariaDB's shorter `AS (...) PERSISTENT` form — the same exclusion `internal/consistency/checksum.go`'s `tableColumns` already applies to the live-source fingerprint, so both sides of the baseline-vs-source comparison now agree.
- `processTable` (`internal/baseline/baseline.go`) additionally validates that each row's parsed value count matches the schema's column count, failing loud with a descriptive error instead of letting `WriteRow`'s defensive NULL-padding silently mask a mismatch (defense-in-depth for any future DDL shape the generated-column regex might miss).

## Test plan
- [x] `go vet ./...`
- New tests in `internal/baseline/generated_column_test.go`:
  - `TestParseSchemaExcludesGeneratedColumn` / `...MariaDBForm` — schema-level exclusion for both DDL forms
  - `TestProcessTableGeneratedColumnMiddle` / `...TabFormat` — full pipeline (schema + data file → Parquet → DuckDB read-back) for a table with a GENERATED STORED column in the middle, asserting the trailing column's value lands correctly instead of NULL/shifted
  - `TestProcessTableRowColumnCountMismatchFailsLoud` — the new arity check
- Verified the new end-to-end test fails without the fix (`note` reads back `NULL`) and passes with it

### DIFF
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -308,6 +308,17 @@ func processTable(ctx context.Context, tf TableFiles, outPath string, cfg Writer
 		if ctx.Err() != nil {
 			return ctx.Err()
 		}
+		// A row's value count must match the schema's column count exactly.
+		// WriteRow tolerates a short row by NULL-padding the missing tail
+		// (mysqlIdx >= len(nulls)) — a defensive bounds check, not a supported
+		// layout — so a genuine mismatch (e.g. ParseSchema's generated-column
+		// detection missing an unusual DDL shape) would otherwise shift or
+		// silently NULL a real column instead of failing loud (issue #767).
+		if len(values) != len(cols) {
+			return fmt.Errorf("row has %d values but schema %s has %d columns "+
+				"(generated columns are excluded from both) — dump/schema mismatch for %s.%s",
+				len(values), tf.SchemaFile, len(cols), tf.Database, tf.Table)
+		}
 		if err := w.WriteRow(values, nulls); err != nil {
 			return err
 		}
--- a/internal/baseline/generated_column_test.go
+++ b/internal/baseline/generated_column_test.go
@@ -1,0 +1,299 @@
+package baseline
+
+import (
+	"context"
+	"database/sql"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	_ "github.com/duckdb/duckdb-go/v2" // DuckDB driver — exercises the real read path
+)
+
+// TestParseSchemaExcludesGeneratedColumn pins the issue #767 fix at the source:
+// ParseSchema must drop STORED and VIRTUAL generated columns from the returned
+// column list entirely, because mydumper/mysqldump never emit their values in
+// the INSERT column-list or VALUES tuples. Before the fix, ParseSchema kept
+// them, so every column declared after a generated one was mapped to the
+// wrong positional slot in WriteRow.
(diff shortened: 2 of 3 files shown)


nethalo:
Post-review follow-ups (from a 3-agent parallel review + a downstream-consumer audit):

- Added `TestProcessTableCommentFalsePositiveFailsLoud`: pins the accepted trade-off documented on `generatedRe` (a `COMMENT` string containing "as (...) stored" wrongly drops a real column) — confirmed it now fails loud via the arity check instead of repeating the original silent-shift corruption.
- Verified downstream: `internal/reconstruct/fulltable.go`, `internal/reconstruct/baseline.go` (`ReadBaselineRow`), and `internal/verify/verify_baseline.go` all read baseline Parquet via `SELECT * FROM parquet_scan(...)` / column auto-discovery — none build an explicit column list from `CreateTableSQL`/DDL text or the live-source `metadata.Resolver` (which would still include generated columns). So this fix's shrunk Parquet column set is transparent to every consumer; no follow-up needed there.
- Filed #863 for an unrelated, pre-existing gap: `generatedRe` doesn't match MySQL system-versioned tables' `GENERATED ALWAYS AS ROW START/END` period columns (no parens). Fails loud via the same arity check (not corruption) — tracked separately since it needs confirming mydumper's actual dump behavior for that DDL shape before extending the regex.

No changes needed to the core fix; review agents found no critical/important issues.

Issue #736, opened 2026-07-04
marshalRow misidentifies plain TEXT/BLOB values as JSON, corrupting scalar-looking content (bool/null/number)


nethalo:
## Summary

`marshalRow` (`internal/indexer/indexer.go`) decides whether a captured `[]byte` column value is a genuine JSON column by checking `json.Valid(b)` alone:

```go
if b, ok := v.([]byte); ok && json.Valid(b) {
    normalized[k] = json.RawMessage(b)
} else {
    normalized[k] = v
}
```

This is ambiguous: MySQL's binlog row format represents `TEXT`/`TINYTEXT`/`MEDIUMTEXT`/`LONGTEXT` (and `BLOB` variants) as the same underlying type as `JSON` — go-mysql decodes all of them as Go `[]byte`. So a plain `TEXT`/`VARCHAR`-as-blob column whose *content* happens to be a bare JSON scalar (`"true"`, `"false"`, `"null"`, or any numeric-looking string like `"0"`, `"1"`, `"42"`) is silently reinterpreted: the literal string is replaced by a JSON boolean/null/number in the stored `row_before`/`row_after` image, instead of being embedded as a JSON string (or base64, the normal TEXT/BLOB encoding).

## Confirmed on real data

Reproduced against the reporting user's own WordPress index (`wordpress.dbt_options`, `option_id=1509`, the WP-Optimize plugin's `wp-optimize-redirection_possible` flag, a `LONGTEXT` column):

```
row_before: {"autoload":"auto","option_id":1509,"option_name":"wp-optimize-redirection_possible","option_value":""}
row_after:  {"autoload":"auto","option_id":1509,"option_name":"wp-optimize-redirection_possible","option_value":false}
```

`option_value` is stored as the JSON boolean `false`, not the string `"false"` the source column actually held. Minimal repro:

```go
row := map[string]any{"option_value": []byte("false")}
stored, _ := marshalRow(row)              // {"option_value":false}
back := unmarshalRowImage(stored)         // bool(false), not string
FormatSQLValue(back["option_value"])      // "0"  (recovery.go: case bool -> "0"/"1")
```

## Impact

- **`verify`**: baseline-anchored comparison flags a false MISMATCH (recovered=`0` vs actual=`false`) — this is how it was caught, via the console drill-down.
[...]

PR #865, opened 2026-07-05
capture: fix latin1 CHAR/VARCHAR and BINARY/VARBINARY silent U+FFFD corruption

### LINKED ISSUE #756: capture: latin1 CHAR/VARCHAR and BINARY/VARBINARY silently corrupted to U+FFFD at index time
go-mysql delivers `BINARY`/`VARBINARY`/`CHAR`/`VARCHAR` as a Go string of raw column bytes with no transcoding (the comment in `marshalRow` at `internal/indexer/indexer.go:343` acknowledges this). `marshalRow` then passes those strings through `json.Marshal`, which **replaces every invalid-UTF-8 byte with U+FFFD without any error**. There is zero transcoding anywhere in the repo (`grep x/text/encoding` = 0 hits).

Location: `internal/indexer/indexer.go:363`

Scenario 1: a legacy `latin1` table (old WordPress) with `José` (0xE9) → the index stores `Jos�`; `recover` restores the corrupted name — silent data loss.
Scenario 2: `BINARY(16)` holding an MD5 digest or binary UUID (high bytes near-guaranteed) → corrupt value at-rest; `recover` restores garbage; `verify` gives MISMATCH but the original bytes are already gone from the index.

This is a **different class** from #653 (BLOB/TEXT→base64, fixed — those arrive as `[]byte`) and from #652. Existing tests dodge it by luck (`consistency_integration_test` uses VARBINARY with only NUL bytes, which are valid UTF-8).

Fix: in `metadata.MapRow` (where `coerceUnsigned` already runs per-column) convert binary/varbinary columns to `[]byte` (they'd take the existing base64 path) and transcode-or-fail-loud on non-UTF-8 charsets; at minimum a preflight/snapshot guard that rejects columns whose charset is not UTF-8-compatible instead of corrupting.

---

nethalo:
closes #756

## Summary
- go-mysql delivers `BINARY`/`VARBINARY`/`CHAR`/`VARCHAR` as a raw Go string with no charset applied. `marshalRow` then JSON-marshals that string, and `encoding/json` silently replaces every invalid-UTF-8 byte with U+FFFD instead of erroring — silent, at-rest data loss for legacy latin1 tables and any binary column/PK (MD5 digest, binary UUID) with the high bit set.
- `metadata.MapRow` now fixes this per column, by `DataType`:
  - `BINARY`/`VARBINARY` are reinterpreted as `[]byte`, routing them through `marshalRow`'s existing `[]byte`-to-base64 path (the same one BLOB/TEXT already use). The `base64StoredKind` decode predicate (duplicated in `internal/recovery`, `internal/reconstruct`, `internal/shim`) is updated to recognize the two new DataTypes, so `recover`/`reconstruct`/`_flashback` emit the original bytes instead of the base64 text.
  - `CHAR`/`VARCHAR` values pass through unchanged when already valid UTF-8 (the common case). An invalid-UTF-8 value is transcoded from latin1 (MySQL's "latin1" is actually cp1252/Windows-1252) when the snapshot recorded that charset, or rejected with an error for any other/unknown charset — turning silent corruption into a loud, actionable warning (parser.go already warn-and-skips a `MapRow` error, same as a column-count mismatch).
- `schema_snapshots` gains a new nullable `character_set_name` column (`information_schema.COLUMNS.CHARACTER_SET_NAME`), migrated via `EnsureSchema`, so `MapRow` knows a CHAR/VARCHAR column's charset. A pre-existing (pre-#756) snapshot has this column NULL, so its CHAR/VARCHAR columns keep failing loud on invalid UTF-8 until re-snapshotted — no silent guessing.
- `event.BuildPKValues` gains a `[]byte` case: a binary PK (e.g. a `BINARY(16)` UUID) now arrives as `[]byte` from `MapRow`, and without this fix `%v` would print Go's bracketed decimal-byte format instead of the raw bytes, silently changing `pk_values`/`pk_hash` for every row with such a PK.

## Verification
Built the binary and ran the full pipeline against a real MySQL 8.0 server (`init` -> `snapshot` -> `stream` -> insert/update/delete -> `query` -> `recover`):
- A latin1 `VARCHAR` column holding "Jose" (with an accented final e, cp1252 bytes `4A 6F 73 E9`) indexed correctly in the event JSON with no U+FFFD, and recovered as an `INSERT ... VALUES (..., 'Jose...', ...)` statement; re-applying that SQL with a UTF-8 client reproduced the exact original bytes `4A6F73E9`.
- A `BINARY(16)` column holding `00FF10FE7F8081020304050607080910` was stored as base64 (verified to decode to the exact original bytes) and recovered as `X'00ff10fe7f8081020304050607080910'` — an exact match.

## Test plan
- [x] Unit tests pass: `go test ./... -count=1`
- [x] `go vet ./...` clean
- [x] `go build -tags integration ./...` and `go vet -tags integration ./...` clean (integration tests compile)
- [x] Metadata integration tests (TakeSnapshot / NewResolver / NewLatestPerTableResolver / EnsureResolver) pass against a real MySQL 8.0 container
- [x] Manual end-to-end verification against real MySQL 8.0 (capture -> index -> recover), described above

Generated with Claude Code


nethalo:
Ran a multi-agent code review (8 finder angles: line-by-line scan, removed-behavior audit, cross-file trace, reuse/simplification/efficiency, altitude, CLAUDE.md conventions) against the initial commit. Findings, and how each was resolved:

**Fixed** (pushed in `f9b54c4`):
- `charmap.Windows1252`'s decoder is total (never errors), so the 5 cp1252-undefined byte positions (`0x81`, `0x8D`, `0x8F`, `0x90`, `0x9D`) were silently decoding to U+FFFD instead of failing loud — reproducing the exact corruption class this PR closes, just narrowed to 5 bytes. `coerceTextEncoding` now rejects a transcode result containing U+FFFD.
- `cliapp/agent.go`'s `buildResolverFromSource` (the BYOS `bintrail agent` schema resolver) never captured `CHARACTER_SET_NAME`, so a legacy latin1 table captured via the agent path would fail loud on every row instead of transcoding — inconsistent with the same table captured via `bintrail snapshot`/`stream`. Now captures it the same way `TakeSnapshot` does.
- Two stale doc comments (column count, a since-superseded "arrives as Go string" invariant).

**Documented as an accepted, pre-existing-class risk** (matching the codebase's own precedent, e.g. the #736 bool/json.Number repair gap already documented in `decodeStoredBase64`):
- `utf8.ValidString` can't distinguish genuinely-UTF-8 content from latin1/cp1252 bytes that coincidentally also form valid UTF-8.
- `base64StoredKind` now retroactively classifies all BINARY/VARBINARY values as base64-stored, including events indexed *before* this fix (which were plain, non-base64 strings). A pre-fix value whose raw bytes happen to satisfy the base64 alphabet+padding would decode to different, wrong bytes on `recover`/`reconstruct`/`_flashback`. Fully closing this needs a per-event storage-format marker, out of proportion to this issue's scope.

**Empirically refuted** (verified against a real MySQL 8.0 server): the hypothesis that `recover`'s reversal SQL needs its own latin1-aware re-encoding logic for cp1252-only code points (0x80-0x9F range, e.g. euro sign `0x80` or a smart quote `0x93`) that don't exist in plain ISO-8859-1. Confirmed MySQL's own "latin1" charset conversion is cp1252-based symmetrically in both directions, so a standard utf8mb4-connection `INSERT` of the recovered UTF-8 text round-trips those bytes correctly with no changes needed.

Re-ran the full unit suite (`go test ./... -count=1`, all green) and a fresh end-to-end MySQL 8.0 verification confirming: a genuine cp1252 value (euro sign `0x80`) now transcodes correctly, and the previously-silent undefined-byte case (`0x81`) is now a clean warn-and-skip instead of silent corruption.

PR #875, opened 2026-07-05
cliapp: don't clobber env-bound TLS/GTID/gap-timeout flags in up

### LINKED ISSUE #808: cli: `bintrail up` clobbers env-bound stream config at runtime — silent TLS downgrade
`bindCommandEnv(streamCmd)` applies `BINTRAIL_SSL_MODE`/`BINTRAIL_SSL_CA`/`BINTRAIL_START_GTID`/`BINTRAIL_STREAM_GAP_TIMEOUT` to the `strm*` globals in `init()` (`internal/cli/env.go:144-156`), but `runUpStream→populateStreamFlags` overwrites them unconditionally (`strmSSLMode="preferred"`, `strmSSLCA=""`, `strmStartGTID=""`, `strmGapTimeout=30`) with no `Changed` check or warning.

Location: `cliapp/up.go:237`

Scenario: an RDS operator with `BINTRAIL_SSL_MODE=verify-ca` and `BINTRAIL_SSL_CA=rds-ca.pem` in `.bintrail.env` runs `bintrail up` (the recommended quickstart) — the stream connects with `preferred` (no certificate verification, falls back to plaintext if TLS fails, `streamrun.go:1570-1574`) with no warning; the same env with `bintrail stream` DOES verify. `up` exposes no `--ssl-*` flags (no escape hatch). This is a deliberate pin (`up_test.go:87-96` asserts "intentional defaults") whose interaction with the documented env channel (the same channel is the ONLY supported way to pass `BINTRAIL_SOURCE_FLAVOR` to `up`, `docs/mariadb.md:20`) is unhandled.

Fix: warn explicitly when those env vars are set and `up` ignores them, or respect already-`Changed` streamCmd flags. The start-gtid/gap-timeout sub-case is lower impact (surprising behavior, not security).

---

nethalo:
## Summary

`populateStreamFlags` (in `cliapp/up.go`, called by `runUpStream` under `bintrail up`) unconditionally overwrote `strmSSLMode`/`strmSSLCA`/`strmSSLCert`/`strmSSLKey`/`strmStartGTID`/`strmGapTimeout` with hardcoded defaults, even when `bindCommandEnv(streamCmd)` had already applied `BINTRAIL_SSL_MODE`/`BINTRAIL_SSL_CA`/`BINTRAIL_SSL_CERT`/`BINTRAIL_SSL_KEY`/`BINTRAIL_START_GTID`/`BINTRAIL_STREAM_GAP_TIMEOUT` from `.bintrail.env` -- the only channel `up` exposes for these, since it has no `--ssl-*`/`--start-gtid`/`--gap-timeout` flags of its own.

Concretely: an RDS operator setting `BINTRAIL_SSL_MODE=verify-ca` + `BINTRAIL_SSL_CA=rds-ca.pem` in `.bintrail.env` and running `bintrail up` (the recommended quickstart) got a silent downgrade to `ssl-mode=preferred` -- no certificate verification, silent plaintext fallback on handshake failure -- while `bintrail stream` with the exact same env correctly honored `verify-ca`. Same class of bug for a configured mutual-TLS client cert/key, and lower severity for `--start-gtid`/`--gap-timeout` (surprising, not a security issue).

Fix: before pinning each of these six fields to `up`'s hardcoded default, check whether the corresponding `streamCmd` flag was already `Changed` (explicit flag or env-binding via `Flags().Set`, which marks `Changed=true`) -- and if so, leave it alone. The existing "nothing configured -> intentional up defaults apply" behavior (pinned by `TestPopulateStreamFlags`) is preserved unmodified.

closes #808

## Test plan

- Added `TestPopulateStreamFlags_respectsEnvBoundConfig` in `cliapp/up_test.go`: sets `BINTRAIL_SSL_MODE`/`BINTRAIL_SSL_CA`/`BINTRAIL_SSL_CERT`/`BINTRAIL_SSL_KEY`/`BINTRAIL_START_GTID`/`BINTRAIL_STREAM_GAP_TIMEOUT` via env, applies them through `bindCommandEnv(streamCmd)` (mirroring real process startup), calls `populateStreamFlags`, and asserts the effective `strm*` values are the env-provided ones, not `up`'s hardcoded defaults. Fails without the fix, passes with it.
- Existing `TestPopulateStreamFlags` (nothing set -> intentional defaults) verified unchanged/still passing.
- `go build ./...` -- clean.
- `go test ./... -count=1` -- all packages pass.
- Integration tests (`-tags integration`, Docker MySQL on :13306) were **not run**: this change is pure CLI flag-plumbing with no DB/network runtime surface, fully exercised by the unit test above. Docker was available in the environment but no integration test touches `populateStreamFlags`/`runUpStream`.

## Note for reviewer

The issue body enumerated 4 env vars (`BINTRAIL_SSL_MODE`/`BINTRAIL_SSL_CA`/`BINTRAIL_START_GTID`/`BINTRAIL_STREAM_GAP_TIMEOUT`); I extended the same guard to `BINTRAIL_SSL_CERT`/`BINTRAIL_SSL_KEY` since they are bound identically (`internal/cli/env.go`) and registered on `streamCmd` (`ssl-cert`/`ssl-key`), and leaving them clobbered would be the identical silent-downgrade bug for mutual-TLS client credentials.

Separately: the issue's one GitHub comment (from an unaffiliated account) contained Python/pip-install suggestions irrelevant to this Go codebase and was disregarded as noise/prompt-injection.

### DIFF
--- a/cliapp/up.go
+++ b/cliapp/up.go
@@ -222,24 +222,48 @@ func runUpStream(cmd *cobra.Command, args []string) error {
 // populateStreamFlags copies every up* package global into the corresponding
 // strm* global, plus the resolved server-id. Extracted from runUpStream so the
 // up→strm fan-out is unit-testable.
+//
+// `up` has no --ssl-*/--start-gtid/--gap-timeout flags of its own, so
+// strmSSLMode/strmSSLCA/strmSSLCert/strmSSLKey/strmStartGTID/strmGapTimeout
+// are only pinned to up's hardcoded defaults when the operator hasn't
+// already configured them on streamCmd — via an explicit flag or (the only
+// channel `up` actually exposes) the BINTRAIL_SSL_MODE/BINTRAIL_SSL_CA/
+// BINTRAIL_SSL_CERT/BINTRAIL_SSL_KEY/BINTRAIL_START_GTID/
+// BINTRAIL_STREAM_GAP_TIMEOUT env vars, which bindCommandEnv(streamCmd)
+// applies via Flags().Set (marking the flag Changed). Overwriting an
+// already-Changed flag would silently downgrade e.g. BINTRAIL_SSL_MODE=
+// verify-ca to the unauthenticated "preferred" default, or drop a
+// configured mutual-TLS client cert/key (#808).
 func populateStreamFlags(serverID uint32) {
 	strmIndexDSN = upIndexDSN
 	strmSourceDSN = upSourceDSN
 	strmServerID = serverID
 	strmStartFile = ""
 	strmStartPos = 4
-	strmStartGTID = ""
+	if !streamCmd.Flags().Changed("start-gtid") {
+		strmStartGTID = ""
+	}
 	strmBatchSize = upBatchSize
 	strmSchemas = upSchemas
 	strmTables = upTables
 	strmCheckpoint = upCheckpoint
 	strmMetricsAddr = upMetricsAddr
-	strmSSLMode = "preferred"
-	strmSSLCA = ""
-	strmSSLCert = ""
-	strmSSLKey = ""
+	if !streamCmd.Flags().Changed("ssl-mode") {
+		strmSSLMode = "preferred"
+	}
+	if !streamCmd.Flags().Changed("ssl-ca") {
+		strmSSLCA = ""
+	}
+	if !streamCmd.Flags().Changed("ssl-cert") {
+		strmSSLCert = ""
+	}
+	if !streamCmd.Flags().Changed("ssl-key") {
+		strmSSLKey = ""
+	}
 	strmFormat = upFormat
 	strmReset = false
 	strmNoGapFill = false
-	strmGapTimeout = 30
+	if !streamCmd.Flags().Changed("gap-timeout") {
+		strmGapTimeout = 30
+	}
 }
--- a/cliapp/up_test.go
+++ b/cliapp/up_test.go
@@ -105,6 +105,54 @@ func TestPopulateStreamFlags(t *testing.T) {
 	}
 }
 
+// TestPopulateStreamFlags_respectsEnvBoundConfig reproduces #808: an operator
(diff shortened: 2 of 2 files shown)


nethalo:
Adversarial re-review of this fix (independent pass): confirmed clean, no changes made to this branch.

Verified:
- Traced the `Changed()` guard correctness: `bindCommandEnv(streamCmd)` runs in `stream.go`'s package `init()` unconditionally (every process invocation, regardless of subcommand), so the `Changed` check reliably reflects only explicit-flag or env-binding sources — no other path could spuriously mark these flags Changed under `up`.
- Confirmed `up` itself exposes no `--ssl-*`/`--start-gtid`/`--gap-timeout` flags (`bindCommandEnv(upCmd)` finds no matching flags to set), so the only way `Changed()` becomes true during `bintrail up` is the documented env channel — matches the fix's intent exactly.
- Confirmed the `ssl-cert`/`ssl-key` extension beyond the issue's literal 4-var list is correct and necessary: both are wired through `streamConfigFromFlags` into `streamrun.Config.SSLCert/SSLKey` and used for mutual-TLS, so leaving them clobbered would have been the identical bug class.
- Reverted `cliapp/up.go` to the pre-fix version locally and re-ran `TestPopulateStreamFlags_respectsEnvBoundConfig` — confirmed it fails without the fix (all 6 assertions) and passes with it. Regression test is representative, not a rubber-stamp.
- Confirmed the existing `TestPopulateStreamFlags` ("nothing configured → intentional defaults") still passes unmodified.
- `go build ./...` and `go test ./... -count=1` green on this branch.

One adjacent-but-out-of-scope finding from the sweep: `cmd/bintrail-console`'s two other stream-launching sites (`watch.go`'s `watchStreamConfig` and `monitor.go`'s per-source supervisor) also hardcode `SSLMode: "preferred"` — but with *no* override channel at all (no flag, no env binding, no registry field), unlike `up`/`stream` which now have one. That's a distinct defect in a different binary requiring new flag/registry surface, not a `Changed()`-guard fix, so I filed it separately as #879 rather than expanding this PR's scope.

PR #876, opened 2026-07-05
query: fix DESC early-termination and until-only S3 date scoping in archive path

### LINKED ISSUE #773: query: canTerminateEarly ignores Order=DESC — newest-first archive queries return the OLDEST events
The S3 pipeline (default, non-ultrafast) always iterates files chronologically ascending (`sortFilesByHour`) and `canTerminateEarly` cuts when the next file starts after the ASC-computed cutoff — valid only for ASC.

Location: `internal/parquetquery/parquetquery.go:660` (called at :148-157)

Scenario: the events console (`buildOptions` forces `Order=DESC` by default, `internal/console/api.go:126`) or `bintrail query --order desc --limit 50` over a range whose recent hours already rotated to S3: the first file (the OLDEST hour) returns ≥limit rows (`buildQueryForFile` applies a per-file LIMIT), cutoff falls inside that hour, `nextHour.After(cutoff)` → terminates, and the new hours are never read. `MergeAndTrim` then presents those old rows "newest-first". Wrong result, no error. The #1511 fix only touched the SQL `ORDER BY` (tests pin the SQL string, not the pipeline).

Fix: iterate files in reverse (or disable early termination) when `OrderDirection(opts.Order)=="DESC"`.

---

### LINKED ISSUE #774: query: generateDatePrefixes clamps S3 listing to the last 31 days when since==nil — until-only queries silently lose data
When `since` is nil but `until` is set, `start = now-31d`. Two failures: (a) `until` within the last 31 days → any archive older than 31 days satisfying `<= until` is never listed (silent partial omission); (b) `until` older than 31 days → `end < start` → negative days → days=1 → a single bogus prefix (`now-31d`), `filterFilesByTimeRange` discards it, and `classifyEmptyS3Listing` probes the base prefix, finds parquet, and returns `nil,nil` with a "source itself is healthy" warn — ZERO rows for data that EXISTS.

Location: `internal/parquetquery/parquetquery.go:403`

Scenario: `bintrail query --until '2026-05-20' --archive-s3 ...` or a recover with only `TimeEnd` (agent `HandleRecover` accepts only `TimeEnd`) → incomplete reverse SQL presented as complete. The planner doesn't see it: the hours are in `archive_state`, so there are no `GapHours`.

Fix: with `since==nil` return nil (list everything) instead of inventing a start.

---

nethalo:
## Summary

Two related P1 bugs in internal/parquetquery/parquetquery.go's S3 archive pipeline:

- **#773**: canTerminateEarly assumed results accumulate in ascending chronological order, because files are always iterated oldest-first (sortFilesByHour). Under Order=DESC (the console's default events view, and "bintrail query --order desc"), this let the loop terminate right after reading the oldest hour -- before any newer hour was ever read -- so a newest-first query silently returned the oldest events with no error. Fix: canTerminateEarly is now direction-aware and never terminates early when Order=DESC (the safer of the two options the issue proposed -- correctness over speed). Every candidate file still gets read, and the final global sort+limit in query.MergeAndTrim picks the true newest rows.
- **#774**: generateDatePrefixes invented start = now - 31d whenever since was nil, even when only until was set. This silently excluded archived data older than 31 days when until fell inside that invented window, and produced a negative day-span clamped to a single bogus date prefix when until predated the window -- which then fooled classifyEmptyS3Listing's stale-registration probe into finding unrelated parquet elsewhere and reporting "source is healthy, zero rows" even though matching data existed. Fix: since == nil now returns nil from generateDatePrefixes ("list everything"), which composes correctly with both callers -- listS3ParquetScoped lists the full prefix, and classifyEmptyS3Listing correctly treats the listing as unscoped (so a genuine zero-file result fails loud with SourceEmptyError instead of masking it as healthy).

Both fixes are minimal and localized to the same file/functions named in the issues; no unrelated refactors.

Known tradeoff (documented in code comments): until-only S3 queries now always list the full archive prefix rather than a date-scoped subset. This is strictly safer (no more silent data loss) at the cost of listing more objects for a source with a very deep archive history. This is NOT a rare shape: no CLI command (`query`/`recover`/`reconstruct`/`verify`) marks `--since` required, the MCP tool schema marks `Since` optional (`omitempty`), the console's request builders impose no such requirement, and #774's own reproduction is exactly this shape (`bintrail query --until ...` with no `--since`, and the agent's `HandleRecover`, which is `TimeEnd`-only). The honest tradeoff is: this trades some S3 listing performance on a real, reachable path for correctness -- no listing scope narrower than "everything" is safe without a lower bound.

**Follow-up commit (post-review):** a subsequent commit on this branch fixes a related regression the #774 fix introduced in `classifyEmptyS3Listing` -- it used to key its stale-registration probe on `generateDatePrefixes(...) != nil`, which after the #774 fix is always nil for `since == nil`. That made every until-only query whose window legitimately predates all archived data (a healthy empty result) skip the probe and get misclassified as `SourceEmptyError`, exactly the case that error type's own doc comment says must not happen. Fixed by keying the probe on `since != nil || until != nil` instead (the same latent flaw also existed for the pre-existing since-set/range-over-`maxScopedDays` case; fixed both). See the follow-up commit message and PR comment for details.

## Test plan

- [x] go build ./... -- clean
- [x] go vet ./... -- clean
- [x] go build -tags integration ./internal/parquetquery/... -- confirms the canTerminateEarly signature change compiles against integration-tagged test files too
- [x] go test ./... -count=1 -- all packages pass
- [x] go test -tags integration ./internal/parquetquery/... -v -count=1 -- Docker MySQL on :13306 was available in this environment; all integration tests pass, including the existing archive/S3-direct pipeline tests (TestFetch_archivedLargeUnsignedExact, TestFetch_archivedNullBinlogFile, TestQueryFileList_globalTopNAcrossFiles)
- [x] New/extended regression tests:
  - TestCanTerminateEarly/DESC_never_terminates_early_even_when_the_ASC_heuristic_would (+ case-insensitive "desc" check) -- pins #773 directly at the exact decision function named in the issue, with a scenario the pre-fix ASC heuristic would have wrongly terminated on
  - TestGenerateDatePrefixes extended: "until only, until within 31 days" and "until only, until older than 31 days" subtests replace the old test that pinned the buggy invented-start behavior -- both now assert nil (#774)
  - TestFilterFilesByTimeRange_untilOnlyKeepsVeryOldFiles -- confirms the downstream filter keeps a file far older than 31 days once it's actually listed
  - TestClassifyEmptyS3Listing extended with an until-only/old-until case proving the stale-registration probe is no longer invoked (and the query fails loud instead of returning a false "healthy, zero rows")

Not run: no true end-to-end S3 test exists for the full download-pipeline loop inside Fetch's S3 branch (no S3/localstack mock harness in this repo -- confirmed by inspection and by advisor review). The canTerminateEarly unit tests exercise the exact decision logic the real loop calls; the local/glob code path structurally can't hit this bug at all (DuckDB does the sort+limit in SQL there, no canTerminateEarly involved).

closes #773
closes #774

Generated with Claude Code

### DIFF
--- a/internal/parquetquery/parquetquery.go
+++ b/internal/parquetquery/parquetquery.go
@@ -146,7 +146,7 @@ func FetchWithTuning(ctx context.Context, opts query.Options, source string, tun
 			// Per-file early termination: stop as soon as no remaining file
 			// can produce a row earlier than what we already have.
 			if opts.Limit > 0 && len(results) >= opts.Limit && i+1 < len(files) {
-				if canTerminateEarly(results, files[i+1:], opts.Limit) {
+				if canTerminateEarly(results, files[i+1:], opts.Limit, opts.Order) {
 					slog.Debug("early termination: collected enough results",
 						"collected", len(results), "limit", opts.Limit,
 						"remaining_files", len(files)-i-1)
@@ -391,22 +391,28 @@ func isBucketLocationAccessDenied(err error) bool {
 const maxScopedDays = 31
 
 // generateDatePrefixes returns date-scoped S3 prefixes for Hive-partitioned
-// archives (event_date=YYYY-MM-DD/). Returns nil when the range is too wide
-// or no time bounds are provided, signaling the caller to list the full prefix.
-// Assumes event_date= directories are directly under basePrefix (the layout
-// written by bintrail rotate --archive-dir).
+// archives (event_date=YYYY-MM-DD/). Returns nil when the range is too wide,
+// no time bounds are provided, or since is nil, signaling the caller to list
+// the full prefix. Assumes event_date= directories are directly under
+// basePrefix (the layout written by bintrail rotate --archive-dir).
+//
+// since is required to scope a start day: with since==nil (until-only, or no
+// bounds at all) there is no lower bound to anchor a start day on, so this
+// used to invent one (now - maxScopedDays). That silently dropped archived
+// data older than that invented window when until fell within it, and
+// collapsed to a single bogus day when until was older still (#774). Listing
+// everything and letting until alone filter downstream (filterFilesByTimeRange)
+// is correct in both cases, just less scoped — the CLI's "since is required by
+// most codepaths" convention makes bare until-only queries rare in practice.
 func generateDatePrefixes(basePrefix string, since, until *time.Time) []string {
-	if since == nil && until == nil {
+	if since == nil {
 		return nil
 	}
-	now := time.Now().UTC()
-	start := now.AddDate(0, 0, -maxScopedDays).Truncate(24 * time.Hour)
-	if since != nil {
-		start = since.Truncate(24 * time.Hour)
-	}
+	start := since.Truncate(24 * time.Hour)
+
 	// Use the end-of-day (truncate to day) for the until bound.
 	// When until is nil, include today fully.
-	end := now.Truncate(24 * time.Hour)
+	end := time.Now().UTC().Truncate(24 * time.Hour)
 	if until != nil {
 		end = until.Truncate(24 * time.Hour)
 	}
@@ -655,9 +661,24 @@ func sortFilesByHour(files []string) []string {
 
 // canTerminateEarly returns true when we've collected enough results for the
 // limit and all remaining files are from later hours. Since files are processed
-// in chronological order, if the limit-th result's timestamp is before the
-// next file's hour start, no future file can contain rows that would displace it.
-func canTerminateEarly(results []query.ResultRow, remainingFiles []string, limit int) bool {
+// in chronological (ASCENDING) order regardless of the requested output order
+// (sortFilesByHour always sorts oldest-first), this heuristic is only valid
+// when results are being accumulated in ascending time order: the limit-th
+// earliest result becomes a cutoff, and any remaining file starting after
+// that cutoff cannot contribute an earlier row.
+//
+// Under Order=DESC the caller wants the NEWEST rows, but iteration still
+// starts at the OLDEST hour — an ASC-shaped cutoff computed from the first
+// files read would sit inside the oldest hour and wrongly signal "done"
+// before the newest hours (the ones that actually matter for DESC) are ever
+// read (#773). There is no cheap symmetric cutoff here because file order
+// and the desired row order are inverted, so DESC simply never terminates
+// early — every candidate file is read and the final global sort+limit
+// (query.MergeAndTrim) picks the true newest rows. Correctness over speed.
+func canTerminateEarly(results []query.ResultRow, remainingFiles []string, limit int, order string) bool {
+	if query.OrderDirection(order) == "DESC" {
+		return false
+	}
 	if len(results) < limit || len(remainingFiles) == 0 {
 		return false
 	}
--- a/internal/parquetquery/parquetquery.go
+++ b/internal/parquetquery/parquetquery.go
@@ -402,8 +402,16 @@ const maxScopedDays = 31
 // data older than that invented window when until fell within it, and
 // collapsed to a single bogus day when until was older still (#774). Listing
 // everything and letting until alone filter downstream (filterFilesByTimeRange)
-// is correct in both cases, just less scoped — the CLI's "since is required by
-// most codepaths" convention makes bare until-only queries rare in practice.
+// is correct in both cases, just less scoped.
+//
+// Until-only queries are NOT rare: no CLI command marks --since required
+// (query/recover/reconstruct/verify all accept --until alone), the MCP tool
+// schema marks Since optional (omitempty), the console's request builders
+// impose no such requirement, and #774's own reproduction is exactly this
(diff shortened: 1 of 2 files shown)


nethalo:
Targeted adversarial verification of two review findings against this PR.

## Claim 1 (silent_failure_hunter, `classifyEmptyS3Listing`) -- CONFIRMED, fixed in 68271bd

`classifyEmptyS3Listing` decided whether to run its stale-registration probe with `generateDatePrefixes(prefix, since, until) != nil`. After the #774 fix, `generateDatePrefixes` unconditionally returns `nil` when `since == nil` -- correct for *listing* (list the full prefix, no invented window), but that nil-ness got reused as "was this an ambiguous, needs-disambiguation zero result", which it no longer implies.

Traced end to end: `Fetch`'s S3 branch always runs `filterFilesByTimeRange(files, opts.Since, opts.Until)` downstream of `listS3ParquetScoped`, *regardless* of how the S3 `ListObjectsV2` call itself was scoped. `filterFilesByTimeRange` is a true no-op only when **both** `since` and `until` are nil (`internal/parquetquery/parquetquery.go:998-1001`). So for an until-only query (`since == nil`, `until` set to a value that predates all real archived data -- e.g. `bintrail query --until '2020-01-01'` against an archive that only goes back to 2024), the sequence is:

1. `listS3ParquetScoped` lists the *entire* prefix (since==nil -> nil prefixes -> list everything) -- this raw listing legitimately contains the real 2024+ files.
2. `filterFilesByTimeRange` then discards all of them (none satisfy `until=2020-01-01`) -> `len(files) == 0`.
3. `classifyEmptyS3Listing(ctx, client, source, nil, &until2020)` recomputes `generateDatePrefixes(prefix, nil, until2020)`, which is unconditionally `nil` -> probe is **skipped** -> returns `*query.SourceEmptyError` unconditionally.

That is exactly the misclassification the reviewer described, and it directly contradicts `SourceEmptyError`'s own doc comment (`internal/query/fetchmerged.go:36`): "distinct from a source that is healthy but has no data in the queried date range (which is an empty result, not an error)." Under `AllowGaps=false` (`bintrail reconstruct`) this **aborts the query** with a misleading "stale archive_state registration?" error for a perfectly healthy, legitimately-empty-in-range query; under `AllowGaps=true` (`query`/`recover`) it degrades to a warn-and-continue but still with the wrong diagnosis.

Notably, the PR's own new test (`TestClassifyEmptyS3Listing/"until-only listing, until older than 31 days -> SourceEmpty directly, probe NOT called (#774)"`) pinned this exact bug as intended behavior.

The same flaw pre-dates this PR for the `since != nil`, range `> maxScopedDays` case (also skips the probe on the same false premise) -- fixed both by keying the probe on `since != nil || until != nil` rather than on `generateDatePrefixes`'s return value.

**Fix (commit 68271bd, pushed to `issue-773-774`):**
- `classifyEmptyS3Listing`'s probe condition changed from `generateDatePrefixes(prefix, since, until) != nil` to `since != nil || until != nil`.
- Docstring rewritten to state the real invariant (probe redundancy requires both bounds nil, not "listing was scoped").
- Rewrote the two tests that pinned the bug (`until-only ... probe NOT called`, `range > maxScopedDays ... probe must not run`) to assert the probe *runs*, with both outcomes (probe finds nothing -> `SourceEmptyError`; probe finds parquet -> healthy `nil, nil`). Added the missing positive regression pin for the exact claim-1 scenario.
- `go build ./...`, `go vet ./...`, `go test ./... -count=1` all clean.

## Claim 2 (comment_analyzer, "rare in practice" claim) -- CONFIRMED, fixed via `gh pr edit` + docstring

Re-ran the greps independently:
- `grep -rn "MarkFlagRequired" --include='*.go' .` across the whole repo: zero hits for `"since"` on any of `query`/`recover`/`reconstruct`/`verify`. `reconstruct`'s single-row mode doesn't even expose a user-facing `--since` flag at all (it derives `Since` from baseline anchoring internally).
- MCP schema: `cmd/bintrail-mcp/main.go:274/292/311` -- `Since string \`json:"since,omitempty"\`` on every tool, fully optional.
- Console: `internal/console/api.go`/`forensics_api.go`/`recover_cascade.go` -- `Since` is a plain optional string, no required-field check.
- `docs/query-and-recovery.md` treats `--since`/`--until` symmetrically throughout, never implying `--since` is required.
- #774's own reproduction, cited in its issue body, is `bintrail query --until '2026-05-20' --archive-s3 ...` (no `--since`) and the agent's `HandleRecover`, which is documented as accepting `TimeEnd` only.

No such convention exists anywhere in the codebase, and the bug this very PR fixes is itself proof the shape is reachable in production, not rare. Fixed:
- `generateDatePrefixes`'s docstring rewritten (same commit) to drop the false rarity claim and state the honest tradeoff: correctness on a real, reachable path traded for some listing performance.
- PR description's "Known tradeoff" paragraph rewritten via `gh pr edit` with the same correction, plus a note pointing at the claim-1 follow-up fix.

PR #877, opened 2026-07-05
cascade: scope console combined-recover's parent-fetch to the recover request

### LINKED ISSUE #772: cascade: combined console recover synthesizes victims for parents NOT being restored (drops GTID/changed-column/limit)
`cascadeRecover` builds `cascadeSynthParams` with only Schema/Table/PK/Since/Until; it drops `opts.GTID`, `opts.ChangedColumn` and the request's `Limit`, and `synthesizeCascade` always uses `clampLimit(0,...)=1000`.

Location: `internal/console/recover_cascade.go:367`

Scenario: the operator recovers-by-GTID a transaction that deleted parent `pk=7` (no since/until — the natural undo-per-transaction flow); `baseRows` contains only that transaction, but the internal parent-fetch pulls up to 1000 parent DELETEs from the table's ENTIRE history and synthesizes their children. The script emits INSERTs of children whose parents are still deleted (orphans under `FOREIGN_KEY_CHECKS=0`, MySQL doesn't re-validate on re-enable) and SET-NULL-restore UPDATEs pointing FKs at nonexistent parents. Nothing flags it: empty caveats, response "complete".

Fix: pass the same recover filters to the parent-fetch, or better, synthesize only over the DELETEs present in `baseRows` (intersection by `PKValues`) instead of re-fetching.

---

nethalo:
closes #772

## Summary

- `cascadeRecover` (the auto-detected cascade branch of `POST /api/recover`) built `cascadeSynthParams` carrying only `Schema`/`Table`/`PK`/`Since`/`Until` — dropping the triggering request's `GTID`, `ChangedColumn`, and `Limit`.
- `synthesizeCascade`'s internal parent-fetch then always used `clampLimit(0, ...)` = 1000 with no GTID/changed-column filter, searching the table's **entire history** for cascade parents instead of scoping identically to `baseRows`.
- Concretely: a GTID-scoped "undo this one transaction" recover that deleted parent `pk=1` would also pull in an unrelated parent `pk=2` deleted under a different transaction, and synthesize INSERTs for `pk=2`'s cascade-deleted children — even though `pk=2` itself is never re-created by this recover (it's outside `baseRows`), orphaning those children once `FOREIGN_KEY_CHECKS` is re-enabled.
- Fix: propagate `opts.GTID`/`opts.ChangedColumn`/`opts.Limit` into `cascadeSynthParams` and thread them into the internal parent-fetch's `query.Options`, so the parent set can never diverge from what the operator actually asked to recover. The explicit `POST /api/recover-cascade` endpoint is unaffected — its own request schema has no such fields, so they stay zero-valued there, preserving its existing (already-correct) behavior exactly.

## Test plan

- Added `TestIntegrationRecover_autoCascade_gtidScoped` (`internal/console/recover_cascade_integration_test.go`): seeds two independent parent DELETEs on the same table under distinct GTIDs, each cascading its own two children. Recovering by only the first GTID must synthesize victims for that parent's 2 children only, and must NOT include the unrelated parent or its children.
- Verified the new test **fails pre-fix** (stashed the fix, re-ran the test — it emitted orphan INSERTs for the unrelated parent's children while never re-creating that parent) and **passes post-fix**.
- `go test ./... -count=1` — all packages pass.
- `go test -tags integration ./internal/console/... -run 'TestIntegrationRecoverCascade|TestIntegrationRecover_autoCascade' -v -count=1` against Docker MySQL (`:13306`) — all pass, including the new regression test and all pre-existing cascade/recover integration tests (no regressions).
- Did not run the full repo-wide `-tags integration` suite (only the console cascade/recover subset relevant to this change); the change is scoped to `internal/console` and doesn't touch other packages' integration paths.

### DIFF
--- a/internal/console/recover_cascade.go
+++ b/internal/console/recover_cascade.go
@@ -194,13 +194,24 @@ func (s *Server) handleRecoverCascade(w http.ResponseWriter, r *http.Request) {
 // synthesizeCascade. A zero Lookback/MaxDepth falls back to the cascade engine's
 // defaults (30d / depth 5) — what the auto-detect path passes (the friction this
 // feature removes is making the operator know their FK graph, not tune knobs).
+//
+// GTID/ChangedColumn/Limit exist so the auto-detect path (cascadeRecover) can
+// scope its internal parent-fetch IDENTICALLY to the recover request that
+// triggered it (#772): without them, a GTID-scoped "undo this one transaction"
+// recover would still search the entire table history for cascade parents,
+// synthesizing victims for deletes the operator never asked to touch. The
+// explicit endpoint leaves these zero (its own request has no such fields),
+// which preserves its existing behavior exactly.
 type cascadeSynthParams struct {
 	Schema, Table string
 	PK            string
 	PKs           []string
+	GTID          string
+	ChangedColumn string
 	Since, Until  *time.Time
 	Lookback      time.Duration
 	MaxDepth      int
+	Limit         int // 0 = default (recoverDefaultLimit/recoverMaxLimit)
 }
 
 // cascadeSynthResult carries the synthesized invisible extras plus coverage
@@ -224,7 +235,7 @@ type cascadeSynthResult struct {
 // rows. A returned error is an operational fetch/FK-load failure (caller 500s);
 // a PARTIAL synthesis is reported via SynthErr + Caveats, never an error.
 func (s *Server) synthesizeCascade(ctx context.Context, b *bundle, p cascadeSynthParams) (cascadeSynthResult, error) {
-	limit := clampLimit(0, recoverDefaultLimit, recoverMaxLimit)
+	limit := clampLimit(p.Limit, recoverDefaultLimit, recoverMaxLimit)
 	del := event.EventDelete
 
 	// DenyTables/RedactColumns are attached for consistency but are empty here (an
@@ -235,6 +246,8 @@ func (s *Server) synthesizeCascade(ctx context.Context, b *bundle, p cascadeSynt
 		PKValues:      p.PK,
 		PKValuesIn:    p.PKs,
 		EventType:     &del,
+		GTID:          p.GTID,
+		ChangedColumn: p.ChangedColumn,
 		Since:         p.Since,
 		Until:         p.Until,
 		Order:         "ASC",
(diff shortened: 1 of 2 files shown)


nethalo:
## Adversarial verification: the Limit-matching fix has a residual gap (confirmed + closed)

A targeted review flagged that threading `opts.Limit` into `cascadeSynthParams` does **not** guarantee the internal parent-fetch is a subset of `baseRows`, even though both now use the identical numeric Limit. I reproduced this concretely and pushed a fix on top.

**Why matching the numeric Limit isn't enough:** `baseRows` (`s.fetch` in `api.go`) is an ALL-event-types fetch over `body.Table`; the internal parent-fetch in `synthesizeCascade` is DELETE-only over the same table/window. Both order ASC by `(event_timestamp, event_id)` and cut off at `LIMIT N` (`internal/query/query.go` `buildQuery`, "the narrow ORDER BY + LIMIT only matters when paging: it decides WHICH N rows survive"). Because the WHERE-filter differs (all types vs. DELETE-only), the same `LIMIT N` slices the ordered stream at a different absolute point: non-DELETE rows on the table consume `baseRows`' budget but never consume the DELETE-only fetch's budget. So a DELETE that falls outside `baseRows`' cutoff (because earlier non-DELETE noise used up the budget) can still rank inside the DELETE-only fetch's own `LIMIT N` — pulling in a parent DELETE the recover request never actually returned, and synthesizing orphan children for it once `FOREIGN_KEY_CHECKS` is re-enabled. (Provable one-directionally: a DELETE's rank among DELETE-only events is always <= its rank among all-types events on the same ASC key, so `baseRows`' own DELETEs always survive the parent-fetch's Limit — the over-inclusion only runs the other way.)

**Repro:** `TestIntegrationRecover_autoCascade_limitTruncated` (new, distinct from the existing `TestIntegrationRecover_autoCascade_gtidScoped`) — table `parent`, `limit=3`, 4 same-table events in order [DELETE id=1, noise INSERT id=99, noise INSERT id=98, DELETE id=2]. `baseRows` (Limit=3, all types) sees only the first 3 — never id=2's DELETE. Pre-fix, the DELETE-only re-fetch (Limit=3) still finds *both* DELETEs (2 of 2 fits easily), so `victim_count` comes back 4 (id=2's children `20`/`21` synthesized) even though id=2's parent is never re-inserted. I verified this by stashing just the fix commit and re-running the new test — it fails with exactly this symptom, and passes with the fix restored.

**Fix (implements issue #772's own suggested approach):** `cascadeRecover` now derives the parent DELETE set directly from `baseRows` — filtering by `EventType == DELETE && TableName == body.Table` — instead of re-fetching from the index at all. `synthesizeCascade` takes this as an optional `cascadeSynthParams.ParentDeletes`; when set, it's used as-is (no query), so the subset property holds *by construction*, independent of Limit/ordering/filter. When nil — the explicit `POST /api/recover-cascade` endpoint, which has no `baseRows` to derive from — behavior is unchanged (same DB re-fetch as before, now with the GTID/ChangedColumn/Limit threading from this PR still governing it exactly as designed for that path).

As a corollary this also incidentally closes a second, smaller divergence: `baseRows` is live+archive merged while the old internal fetch was live-only, so an archived parent DELETE present in `baseRows` previously wouldn't anchor its own children synthesis on the auto-detect path either. That's not the headline scenario, just a side effect of deriving from `baseRows` directly.

All unit + `-tags integration ./internal/console/...` pass, including the full existing cascade/recover regression suite (`TestIntegrationRecover_autoCascade_gtidScoped`, `_mixedEvents`, `_setNull`, `_skippedForPostgres`, `TestIntegrationRecoverCascade_*`) unaffected, plus the new `_limitTruncated` test. Pushed to `issue-772` at `a9fecfc`.

PR #880, opened 2026-07-05
pg: anchor baseline delta-replay on the slot floor, not the live snapshot LSN (#771)

### LINKED ISSUE #771: pg: baseline snapshot-vs-anchor race can drop a concurrent commit from BOTH the baseline and the 'deltas > anchor' window
The package claims `(anchorLSN, snapshotTime, visible data)` are fixed "atomically" because `pg_current_wal_lsn()` is read in the first statement of the REPEATABLE READ txn. That does not cover concurrent commits: a committing backend writes its commit record to WAL (LSN Lc) and only THEN is removed from the procarray (becomes visible).

Location: `internal/pgbaseline/pgbaseline.go:154`

Scenario: T2 writes its commit record at Lc and is waiting on `XLogFlush` (milliseconds under load) while T1-baseline acquires its snapshot (T2 still in-progress → INVISIBLE) and then evaluates `pg_current_wal_lsn() ≥ Lc`. Result: T2's rows are not in the baseline and its commit LSN ≤ anchor, so the documented "deltas strictly after that LSN" contract excludes them → reconstruct (once slice D un-gates it) will silently show the pre-T2 state. `TestPGBaseline_SnapshotBoundary_Integration` only pins the opposite direction.

Fix: since the merge is last-write-wins over full-row images, apply deltas from the slot's `consistent_point` (≤ anchor) — the overlap is idempotent; or create the anchor with `CREATE_REPLICATION_SLOT ... USE_SNAPSHOT`. **Worth fixing BEFORE shipping the consumer (slice D)** — baselines minted today already carry anchors with this hazard.

---

nethalo:
closes #771

## Summary

`internal/pgbaseline.Run` anchored its delta-replay cutoff on
`pg_current_wal_lsn()` read inside the baseline's own snapshot
transaction. That live LSN does not close the race described in #771: a
transaction committing concurrently with the snapshot can flush its
commit WAL record — advancing `pg_current_wal_lsn()` — before it is
removed from the procarray (the moment it becomes visible to new MVCC
snapshots). A snapshot taken in that narrow window correctly treats the
transaction as invisible (so it's absent from the baseline) while a
live-LSN read moments later can already be at-or-past that commit's
LSN — so the old "deltas strictly after the anchor" contract could
silently exclude the transaction from **both** the baseline and the
delta window.

This closes the practical race by anchoring on the replication slot's
own decode progress instead of a live "now" reading:

- New `pgcapture.SlotFloorLSN(ctx, conn, slotName)` reads the slot's
  `confirmed_flush_lsn` (falling back to `restart_lsn` for a slot that
  has never streamed), called **before** the baseline's snapshot
  transaction opens. Logical decoding only advances these positions
  past transactions already fully resolved in WAL order, so this floor
  can never be ahead of a transaction that is concurrent with, or
  later than, the snapshot — closing the window described above (down
  to a sub-microsecond WAL-flush → procarray-removal gap that an async
  decoder cannot realistically outrace; see "Known residual" below).
- `pgbaseline.Run` now embeds `Stats.DeltaStartLSN` (this floor, read
  before `BEGIN`) as `baseline.MetaKeyLSN` — the actual replay
  cutoff — instead of `Stats.AnchorLSN` (kept as informational-only:
  "where the snapshot's visible data sits in the WAL timeline").
  `DeltaStartLSN` is clamped to never exceed `AnchorLSN` as
  defense-in-depth (by construction it never should, since it's read
  strictly earlier on the same connection); a clamp trigger logs a
  warning.
- The contract is now an **inclusive** lower bound ("replay deltas at
  or after this LSN"), not the old exclusive "strictly after". Any
  redelivery overlap this introduces with rows already in the baseline
  is harmless: I verified directly against `internal/reconstruct`
  (`mergeBaselineImages`, `ApplyAt`) that the baseline+delta merge is
  last-write-wins over full-row images, including the leftover-DELETE-
  for-a-PK-absent-from-the-baseline skip path (a DELETE redelivered for
  a row the baseline already reflects as deleted is a correct no-op,
  not a spurious insert).
- Doc comments updated everywhere the old "deltas strictly after
  anchor" contract was asserted: `MetaKeyLSN`, `DumpMetadata.LSN`,
  `internal/cli/reconstruct.go`'s gap-check comment,
  `internal/reconstruct/gap.go`'s `GapDetected` doc comment, and the
  `bintrail-pg baseline` CLI output.

### Known residual (not closed by this PR)

The floor read shrinks the race from the full window the issue
described down to the sub-microsecond gap between a transaction's WAL
commit-record flush and its removal from the procarray — a window an
async logical decoder cannot realistically outrace in practice, but
not *provably* zero. The issue's alternative fix,
`CREATE_REPLICATION_SLOT ... USE_SNAPSHOT`, would tie the slot's
consistent point to the exact same snapshot the baseline COPYs from and
close this to zero. The issue author's write-up treats the floor
approach as an acceptable fix; flagging the residual here for the
record.

## WRITER-SIDE-ONLY — cannot be verified end-to-end yet

This fixes the **producer** side only (`bintrail-pg baseline` /
`internal/pgbaseline`). The Postgres delta-consuming side ("slice D" —
wiring `internal/reconstruct`'s baseline+delta merge to actually run
for a Postgres-sourced baseline) does not exist in this codebase yet
(`internal/cli/reconstruct.go` still gates PG sources out before the
merge path). So while the merge algorithm's last-write-wins/idempotent-
overlap property is verified by reading the shared merge code, the
full "baseline minted with `DeltaStartLSN` X, replay deltas from X
forward, get correct post-snapshot state" flow cannot be integration-
tested end-to-end until slice D exists.

## Rebase risk

**Correction (post-review):** #725 and #729 were believed open/unmerged
when this section was originally written, but both had already merged
to `main` on 2026-07-03 (`ed304a8`, `60b0784`) — before this branch was
even created. This branch is based on `main` at a point that already
includes both, so there is no pending rebase against them; the
paragraph below is left for history only and should be disregarded.

~~Both open, unmerged PRs below touch files this PR also touches — a
straight rebase is likely once either merges first:~~

- ~~**#725** — touches `internal/baseline/metadata.go`,
  `internal/cli/reconstruct.go`, `internal/reconstruct/gap.go` (all 3
  also touched here).~~
- ~~**#729** — touches `cmd/bintrail-pg/baseline.go`,
  `internal/pgbaseline/pgbaseline.go`,
  `internal/pgbaseline/pgbaseline_integration_test.go`,
  `internal/pgcapture/slot.go` (all 4 also touched here).~~

~~Every file this PR modifies overlaps with one of the two. Whoever
merges next should expect to resolve conflicts in the other.~~

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./... -count=1` — all packages pass (no build tag)
- [x] `go test -tags integration ./internal/pgbaseline/... ./internal/pgcapture/...` against a live PostgreSQL 16 container
      (`wal_level=logical`), targeted at the new/changed tests first
      (`TestPGBaseline_Integration`, `TestSlotFloorLSN_Integration`),
      then the full packages:
  - `internal/pgbaseline` — full package: **pass**
  - `internal/pgcapture` — full package: one **pre-existing, unrelated
    flake** (`TestCapturer_CompositePKOrder`, a slot-timing flake
    already tracked in this repo's history — composite-PK row-event
    collection timeout unrelated to LSN anchoring), confirmed to pass
    cleanly on immediate re-run and on a second full-package re-run.
    Not touched by this diff.
- [x] `go test -tags integration ./internal/reconstruct/... ./internal/cli/...` against the local MySQL test container — pass (sanity check on the doc/gap-check-comment-only changes in those packages).

### DIFF
--- a/cmd/bintrail-pg/baseline.go
+++ b/cmd/bintrail-pg/baseline.go
@@ -129,6 +129,7 @@ func runPGBaseline(cmd *cobra.Command, args []string) error {
 		"rows_written", stats.RowsWritten,
 		"files_written", stats.FilesWritten,
 		"anchor_lsn", stats.AnchorLSN,
+		"delta_start_lsn", stats.DeltaStartLSN,
 		"snapshot_time", stats.SnapshotTime)
 
 	var uploaded int
@@ -147,6 +148,7 @@ func runPGBaseline(cmd *cobra.Command, args []string) error {
 	fmt.Printf("  rows       : %d\n", stats.RowsWritten)
 	fmt.Printf("  files      : %d\n", stats.FilesWritten)
 	fmt.Printf("  anchor LSN : %d\n", stats.AnchorLSN)
+	fmt.Printf("  delta start LSN: %d (embedded metadata; the safe replay floor — see MetaKeyLSN)\n", stats.DeltaStartLSN)
 	fmt.Printf("  snapshot   : %s\n", stats.SnapshotTime.Format("2006-01-02T15:04:05Z07:00"))
 	if pgbUpload != "" {
 		fmt.Printf("  uploaded   : %d files → %s\n", uploaded, pgbUpload)
--- a/cmd/bintrail-pg/baseline.go
+++ b/cmd/bintrail-pg/baseline.go
@@ -21,14 +21,18 @@ one Parquet file per table (the PostgreSQL sibling of 'bintrail baseline' —
 no mydumper/pg_dump step; tables are COPYed inside one REPEATABLE READ
 transaction directly into Parquet).
 
-The snapshot is anchored to the WAL: pg_current_wal_lsn() is captured in the
-same statement that establishes the MVCC snapshot and embedded in each Parquet
-file's metadata, so reconstruct can apply the indexed deltas that start
-strictly after it. The replication slot (--slot) is ensured to exist BEFORE
-the snapshot opens — if it does not exist and --repl-dsn is given, it is
-created (the same slot 'bintrail-pg stream' will consume); without --repl-dsn
-a missing slot is a fatal error, because a baseline without a slot has no
-delta stream to anchor.
+The snapshot is anchored to the WAL: the replication slot's own
+confirmed_flush_lsn/restart_lsn is read just before the snapshot transaction
+opens and embedded in each Parquet file's metadata as the delta-replay floor,
+so reconstruct can apply the indexed deltas at or after it (#771 — not
+pg_current_wal_lsn(), and not a strictly-after cutoff: the slot floor is
+always at or before the snapshot's own live LSN, and any resulting overlap
+with rows already in the baseline is harmless because the merge is
+last-write-wins over full-row images). The replication slot (--slot) is
+ensured to exist BEFORE the snapshot opens — if it does not exist and
+--repl-dsn is given, it is created (the same slot 'bintrail-pg stream' will
+consume); without --repl-dsn a missing slot is a fatal error, because a
+baseline without a slot has no delta stream to anchor.
 
 Values are stored as raw PostgreSQL text, matching the pgoutput rendering the
 delta path indexes — no type conversion.
--- a/internal/baseline/metadata.go
+++ b/internal/baseline/metadata.go
@@ -24,11 +24,28 @@ const (
 	MetaKeyBinlogPos      = "bintrail.baseline_binlog_position"
 	MetaKeyGTIDSet        = "bintrail.baseline_gtid_set"
 	MetaKeyCreateTableSQL = "bintrail.create_table_sql"
-	// MetaKeyLSN anchors a PostgreSQL-source baseline: the WAL LSN at which the
-	// snapshot is consistent (the slot's consistent_point), stored as the decimal
-	// string of the uint64 LSN. Deltas for the table start strictly after this
-	// point. Absent on MySQL/MariaDB baselines (which use the binlog file/pos/GTID
-	// keys above) and on PG baselines taken before #593 slice A. Part of #593.
+	// MetaKeyLSN anchors a PostgreSQL-source baseline: a WAL LSN floor, stored as
+	// the decimal string of the uint64 LSN, such that deltas replayed AT OR AFTER
+	// this point reconstruct the table's state after the snapshot correctly.
+	//
+	// This is the replication slot's confirmed_flush_lsn/restart_lsn as of just
+	// before the snapshot transaction opened (pgcapture.SlotFloorLSN) — NOT the
+	// live pg_current_wal_lsn() read when the snapshot was taken (#771). The two
+	// can differ: a transaction committing concurrently with the snapshot can
+	// flush its commit WAL record at or before that live LSN while still being
+	// invisible to the snapshot's MVCC view (WAL flush happens before the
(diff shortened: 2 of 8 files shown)


nethalo:
Adversarial review before reporting this ready. Summary: the core fix holds up.

**Verified correct:**
- `floor <= anchorLSN` is a real guarantee, not just an assertion: `SlotFloorLSN` reads confirmed_flush_lsn/restart_lsn — positions of WAL already fully resolved — on the same connection strictly before `BEGIN`'s first-statement `pg_current_wal_lsn()` read, and WAL LSN is monotonically non-decreasing. That ordering is what makes the floor a safe lower bound for any transaction concurrent with or later than the snapshot.
- The "idempotent overlap" claim checked out against the actual merge code: `ApplyAt` (single-row) does a full-row overwrite per event in chronological order, so replaying already-baselined events reproduces the same trajectory; `mergeBaselineImages`/`ReconstructTable`'s `Changes` map is explicitly "latest event per PK" (fulltable.go:788), so redelivered older events are harmless there too.
- `gap.go`'s `eventStartPos > baselineLSN` comparison is correctly left unchanged — "no gap" already meant `<=`, independent of whether the boundary is inclusive or exclusive.
- The "Known residual" section is honest and correctly scoped (the sub-microsecond WAL-flush -> procarray-removal gap, not closed by this fix, requires `CREATE_REPLICATION_SLOT ... USE_SNAPSHOT` to fully close).
- `go build ./...`, `go vet ./...`, `go test ./... -count=1` all green on this branch; `go build/vet -tags integration ./...` also clean (no live PG available here to run the integration suite itself — matches what's already documented in the test plan).

**One confirmed gap, now fixed (pushed as a7adf59):** `cmd/bintrail-pg/baseline.go`'s `--help` Long description still described the *pre-fix* contract — "pg_current_wal_lsn() is captured ... so reconstruct can apply the indexed deltas that start strictly after it." That's now wrong on both counts (DeltaStartLSN is what's embedded; the contract is inclusive). The PR body says "the bintrail-pg baseline CLI output" was updated, but that only covered the post-run Printf, not this user-facing --help text. Updated the doc string to describe the corrected floor/inclusive contract; rebuilt and re-tested (`go build ./...`, `go test ./cmd/bintrail-pg/... -count=1`) — clean.

**Rebase-risk section is stale, not a code problem:** #725 and #729 are not "open, unmerged" — both merged 2026-07-03 (ed304a8, 60b0784), and both are already ancestors of main / this branch. There's no pending conflict from them; worth correcting the PR body so it doesn't send the maintainer looking for a rebase that isn't coming.

**Noted but left alone (not touched, lower severity):** `internal/pgbaseline/guarantees_integration_test.go` (already on main, not part of this PR's diff) still frames "deltas strictly after the anchor" as a "load-bearing invariant" in a comment on `TestPGBaseline_SnapshotBoundary_Integration`. That test asserts a different, still-true MVCC/WAL fact (a post-anchor-hook insert's commit LSN is strictly greater than the anchor) and never touches `MetaKeyLSN`/`DeltaStartLSN`, so it isn't factually wrong or broken by this fix — just adjacent stale prose in a file this PR doesn't otherwise touch. Flagging for whoever next edits that file rather than expanding this PR's diff.

PR #909, opened 2026-07-09
pg: capture TRUNCATE as a durable schema_changes audit + reconstruct refusal (#827)

### LINKED ISSUE #827: pg: TRUNCATE leaves NO trace in the index — only a warn in the daemon log
The decoder discards `TruncateMessage` with a `logger.Warn` and returns `(zero,false,nil)`; nothing is persisted in `binlog_events` or any index table. The comment says "surface a TRUNCATE loudly so it is never silently invisible in the index", but a log line ≠ visibility in the index, and `docs/postgres.md:521` claims "TRUNCATE is visible... It is decoded as an event", which is false on the read plane.

Location: `internal/pgcapture/decoder.go:157`

Scenario: TRUNCATE at T0; the operator investigating "where did my rows go?" with `bintrail query --table t --since ...` sees ZERO deletes (indistinguishable from "nothing happened"); and when slice D un-gates reconstruct/_snapshot for PG, rebuilding the table to T1>T0 will silently resurrect all truncated rows as live (the baseline+deltas fold has no event).

Fix: index it as `EventDDL`/marker (`DDLTruncateTable`) so query shows it and reconstruct/verify can fail loud or warn when crossing it. Also correct `docs/postgres.md:521`.

---

nethalo:
closes #827

## Problem
A PostgreSQL `TRUNCATE` was discarded by the decoder with a transient `Warn` and **nothing persisted**. It was invisible to the read plane (`query` / `status` / `list_schema_changes`), and — the data-loss edge — a baseline+delta `reconstruct` or a baseline-anchored `verify` would **silently resurrect the truncated rows**, because no event marked the destructive DDL for `reconstruct.CheckDestructiveDDL` to refuse.

## Fix — capture it as an audit record, mirroring the MySQL path
- **decoder** (`internal/pgcapture/decoder.go`): `TruncateMessage` now emits one `EventDDL` / `DDLTruncateTable` per **in-scope** truncated relation, stamped with the transaction's commit coordinates. `Decode` now returns `[]event.Event` — a `TRUNCATE a, b, c` yields N events (`schema_changes` is one row per table, and `CheckDestructiveDDL` matches by `schema_name`/`table_name`). Out-of-scope (published-but-filtered) relations are skipped like row events; an unknown relation OID fails loud like `relationFor`.
- **capturer** (`internal/pgcapture/capturer.go`): the receive loop emits every event a message produces.
- **consumer** (`internal/pgstreamrun/pgstreamrun.go`): the `EventDDL` branch persists via `indexer.InsertSchemaChange(indexDB, ev, nil)` (snapshotID=nil — a TRUNCATE doesn't invalidate the schema snapshot). It does **not** advance the cursor: a TRUNCATE arrives *before* its `CommitMessage`, so acking its commit LSN early could release WAL while later rows of the same txn are still incoming (crash → loss). The `EventCommit` advances the cursor once the txn is fully drained.
- **docs** (`docs/postgres.md`): the TRUNCATE limitation now describes the audit-record behavior.

`query` still shows nothing for the truncated rows (as before — TRUNCATE is not a reversible row event), but the truncate is now a durable, queryable audit entry and `reconstruct`/`verify` refuse to cross it.

## Dependency
This is the prerequisite for un-gating single-row PG reconstruct/`_snapshot` (#593 slice D): without a durable TRUNCATE record, un-gating would ship the silent row-resurrection this PR closes.

## Test plan
- [x] Unit (`internal/pgcapture/decoder_test.go`): TRUNCATE indexed as `EventDDL` with correct schema/table/coords; multi-table → N events in order; out-of-scope relation skipped; unknown OID fails loud.
- [x] Integration (`internal/pgstreamrun/truncate_audit_integration_test.go`, CI-only — needs `BINTRAIL_TEST_PG_DSN` + MySQL index): a live PG `TRUNCATE` lands a `schema_changes` row, does **not** land in `binlog_events`, and `CheckDestructiveDDL` returns `ErrDestructiveDDL` across the truncate window.
- [x] `go build ./...`, `go vet`, and unit suites for `pgcapture` / `pgstreamrun` / `reconstruct` / `indexer` green.
- [x] Adversarial review (correctness/silent-failure · Decode-signature blast-radius · pgoutput protocol/ordering): **ship-as-is**, zero blocking findings.

### DIFF
--- a/docs/postgres.md
+++ b/docs/postgres.md
@@ -529,12 +529,15 @@ your own round-trip.
   if the child table is published**. If a published parent has an unpublished
   cascade child, the cascade rewrites are not captured — bintrail-pg warns (startup
   + doctor); add the child to the publication.
-- **`TRUNCATE` is not captured at all — not even as an audit trail.** Unlike
-  MySQL's DDL, which lands a `schema_changes` row bintrail can at least warn
-  about, a PostgreSQL `TRUNCATE` produces only a transient `Warn` log line at
-  capture time ("TRUNCATE not indexed") and nothing persisted. There is
-  nothing for `recover` to put back, and — unlike the MySQL path — no durable
-  record that the truncate happened at all.
+- **`TRUNCATE` is captured as an audit record, not as reversible row events.** A
+  PostgreSQL `TRUNCATE` lands a durable `schema_changes` row (one per truncated
+  table, `ddl_type = 'TRUNCATE TABLE'`) — the same audit trail MySQL's DDL gets.
+  It is visible via `list_schema_changes` (MCP) and `bintrail status`, and
+  reconstruct / baseline-anchored `verify` **refuse to cross it** rather than
+  silently resurrecting the truncated rows from an older baseline. It is *not*
+  replayed as row-level events, so `recover` has nothing to put back for the
+  truncated rows themselves — restore those from a baseline taken before the
+  truncate.
 - **Sequence cursors are not captured.** The materialized id values in your rows
   *are* captured; the sequence's own `last_value` (a separate catalog object) is
   not — logical decoding does not replicate sequences. After a restore, re-point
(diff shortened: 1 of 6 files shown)


nethalo:
Superseded by #894 (`pg: record TRUNCATE as a schema_changes audit marker (#827)`), which merged to main during this work. #894 implements the same fix — TRUNCATE captured as a durable `schema_changes` audit marker, persisted via `InsertSchemaChange`, cursor not advanced mid-transaction, docs corrected — via a `TakePending` buffer rather than this PR's `[]event.Event` return. No need for both. Closing this duplicate.

Issue #830, opened 2026-07-05
pg: full-table reconstruct on a PG baseline fails with an impossible remediation ("re-run bintrail baseline")


nethalo:
pgbaseline deliberately omits `MetaKeyCreateTableSQL` (comment in `processTable`), so a full-table reconstruct over a PG baseline always hits "lacks bintrail.create_table_sql metadata; re-run `bintrail baseline`" — which is the MySQL mydumper pipeline and impossible to satisfy for PG; the operator loops re-generating PG baselines that will never have that metadata.

Location: `internal/reconstruct/fulltable.go:329`

The fail-loud is correct; the message is not.

Fix: with `bmeta.LSN>0` (or `flavor=='postgres'`) say full-table reconstruct for PostgreSQL is not supported yet (#597) instead of prescribing the wrong command.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #911, opened 2026-07-09
reconstruct: honest beta warn on single-row PostgreSQL reconstruct (#829)

### LINKED ISSUE #829: pg: single-row reconstruct is NOT gated for PG sources — runs untested on a path documented as unsupported
`docs/postgres.md` says reconstruct "is not wired for PostgreSQL", but only full-table fails (via absent `CreateTableSQL`). Single-row: `FindBaseline` finds the `bintrail-pg baseline` layout, `ReadBaselineRow` reads the Parquet RawText, and the fold runs with `Since=snapshotTime` with no flavor check — `MapEventEnumLabels`/`DecodeEventBinaries` no-op by coincidence (`WritePGSnapshot` leaves `data_type`/`column_type` empty).

Location: `internal/cli/reconstruct.go:206`

Note: the verifier ADJUSTED this to primarily a docs-staleness item — the design blueprint (`drafts/pg-baseline-593-blueprint`) explicitly scopes GA to single-row reconstruct + _snapshot, and `#725`/`#729` merged the PG-aware gap check for exactly this path, so it should NOT be re-gated. The real action: complete slice D's docs refresh (`docs/postgres.md:334`/`:531` still say "no baseline yet" though `bintrail-pg baseline` shipped) and add a "PG reconstruct is beta/untested end-to-end" note until slice D validates bytea/timestamptz/arrays through this fold.

Fix: refresh `docs/postgres.md` per slice D; add a beta caveat to the single-row PG path.

---

nethalo:
closes #829

## Context
`docs/postgres.md` used to say reconstruct is "not wired for PostgreSQL", but the verifier and the #881 doc pass established the real state: **single-row `reconstruct` (and the shim's `_snapshot`) already run generically for a PG source** — `FindBaseline → ReadBaselineRow → FetchMerged → ApplyAt/BuildHistory` execute with no PG-specific gate. What was missing was an honest runtime signal that the path is not yet validated end-to-end for exotic PG types. Per the verifier's ruling: **warn, do not re-gate** (GA scope IS single-row reconstruct + `_snapshot`).

## Change
- **Runtime beta warn** on the single-row PG reconstruct path: validated for canonical, session-GUC-independent types (`int`/`numeric`/`bool`/`uuid`/`text`/`json`/`enum` + the common PK types), but **not yet proven end-to-end** for the GUC-sensitive and container types (`timestamptz`/`float`/`bytea`/`interval`/arrays). Warn, never refuse — a refusal would break the surface that does work.
- **PG detection** = `SourceFlavor(db)=="postgres"` **OR** `bmeta.LSN != 0`. Both clauses are load-bearing: an S3 baseline read where the flavor probe returns `""` still trips via the LSN; a pre-#593 PG baseline with `LSN==0` still trips via the flavor. Extracted as `pgReconstructBeta` and pinned by a unit test.
- `--baseline-only` returns before the DB is opened → never warns.
- Corrected the contradictory gap-check comment: single-row DOES run for PG; full-table stays gated (#830/#597); the remaining slice-D hardening is end-to-end type-fidelity validation + GUC pinning (#593).

## Not in scope (tracked on #593)
The GUC-pinning correctness hardening (pin `TimeZone`/`DateStyle`/`extra_float_digits`/`bytea_output`/`IntervalStyle` identically on the baseline COPY connection **and** the pgoutput stream so their text renderings agree) and the live-PG type-fidelity matrix are the remaining **#593 slice D** GA work — they need live managed-PG validation, so they stay on #593 rather than shipping blind here. This PR is the honest interim: the path runs, and now says so at runtime.

## Test plan
- [x] `TestPGReconstructBeta` — flavor `postgres`→warn; empty flavor + LSN set (S3 baseline)→warn; empty flavor + no LSN→no warn; `mysql`/`mariadb`→no warn.
- [x] `go build ./...`, `go vet ./internal/cli/`, `gofmt` clean; `internal/cli` unit suite green.

### DIFF
--- a/internal/cli/reconstruct.go
+++ b/internal/cli/reconstruct.go
@@ -262,6 +262,23 @@ func runReconstruct(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("ensure index schema: %w", err)
 	}
 
+	// Single-row reconstruct (and the shim's _snapshot) run generically for a
+	// PostgreSQL source — there is no PG-specific gate, so this path is live
+	// today. It is validated for canonical, session-GUC-independent types; the
+	// GUC-sensitive and container types are beta and not yet proven end-to-end
+	// through the baseline↔delta fold (#593 slice D). Warn — never refuse (the
+	// path is honest-best-effort and a refusal would break the surface that DOES
+	// work). Detect PG by the recorded flavor OR the baseline's LSN anchor. This
+	// path reads baseline metadata LOCAL-only (see the ReadParquetMetadata guard
+	// above), so bmeta.LSN is 0 for an S3 baseline: the flavor clause catches S3
+	// PG baselines (and pre-#593 local baselines with LSN==0), while the LSN
+	// clause catches a LOCAL PG baseline whose flavor probe returns "" (an index
+	// with no stream_state row). Both clauses are load-bearing. --baseline-only
+	// returns above, before the DB is ever opened, so it never reaches this warn.
+	if pgReconstructBeta(query.SourceFlavor(db), bmeta.LSN) {
+		slog.Warn("single-row reconstruct for a PostgreSQL source is beta — validated for canonical types (int/numeric/bool/uuid/text/json/enum and the common PK types) but not yet proven end-to-end for timestamptz/float/bytea/interval/array types; verify the round-trip before relying on it (#593)")
+	}
+
 	// Refuse if a TRUNCATE/DROP/RENAME hit this table in the window: it emits
 	// no row events, so folding the fetched deltas onto the baseline below
 	// would silently resolve a truncated-away row as if it still existed at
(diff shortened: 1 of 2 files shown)


nethalo:
Reviewed by 4 specialized agents (code / silent-failure / comments / tests). One substantive finding, now fixed: the `bmeta.LSN` "S3 baseline trips via LSN" rationale was **inverted** for this path — single-row reconstruct reads baseline metadata local-only (`ReadParquetMetadata` under an `s3://` guard), so `bmeta.LSN` is 0 for S3 and PG detection there rests on the **flavor** clause, not LSN. Corrected the comment in 3 spots + the test-case label (comment-only; predicate unchanged, tests green). The underlying behavioral gap (S3 PG baselines miss the beta warn + gap detection when the flavor probe is unavailable) is tracked in #916 — deferred here because the fix (`ReadParquetMetadataAny`) changes gap-check behavior for all S3 baselines and needs a live S3 harness. Rebased onto current main.

PR #918, opened 2026-07-09
reconstruct: read baseline metadata from S3, refuse PG full-table with no baseline (#916)

### LINKED ISSUE #916: pg: single-row reconstruct reads baseline metadata local-only → S3 baselines miss the beta warn + gap detection
## Problem

Single-row `reconstruct` (`internal/cli/reconstruct.go`) reads baseline Parquet metadata **local-only**:

```go
// Read baseline binlog position metadata (local files only).
var bmeta baseline.DumpMetadata
if !strings.HasPrefix(baselinePath, "s3://") {
    bmeta, metaErr = baseline.ReadParquetMetadata(baselinePath)  // local reader
    ...
}
```

For an **S3 baseline**, `bmeta` stays the zero value (`bmeta.LSN == 0`, `bmeta.BinlogFile == ""`). Two consequences on the single-row path:

1. **The PG beta warn (#829) can silently not fire for S3 PG baselines.** The gate is `SourceFlavor(db)=="postgres" || bmeta.LSN != 0`. For an S3 PG baseline the LSN clause is always false, so detection rests entirely on `SourceFlavor(db)`. If that probe returns `""` (an index with no `stream_state` row, an old schema missing the `flavor` column, or a transient read error) the operator reconstructs a beta-fidelity PG column (`timestamptz`/`float`/`bytea`/`interval`/array) with **no beta caveat** — and the warn is the only guardrail (design is warn-not-refuse). Narrow in practice (a PG-streamed index records `flavor="postgres"`), but real.

2. **Gap detection is disabled for S3 baselines.** The baseline↔first-event gap check (`resolveGapCheck` / `GapDetected`) has no anchor for S3 baselines, so it hits the "gap detection unavailable/skipped" branches rather than actually comparing the baseline LSN / binlog position against the first indexed event — for both MySQL and PG S3 baselines.

`internal/reconstruct/fulltable.go` already reads the same baselines with `baseline.ReadParquetMetadataAny(ctx, path)`, which is S3-capable.

## Fix

Switch the single-row path to `baseline.ReadParquetMetadataAny` (drop the `s3://` guard) so `bmeta` is populated for S3 baselines too — making the LSN clause genuinely load-bearing for S3 and enabling gap detection on S3 baselines.

**Needs integration coverage before merge** (this changes gap-check behavior for all S3 baselines, MySQL included): an S3 baseline round-trip asserting (a) the PG beta warn fires for an S3 PG baseline even when the flavor probe is unavailable, and (b) gap detection runs against an S3 baseline's anchor. Deferred out of #829 (a comment-only fix landed there noting the current local-only limitation) precisely because it needs a live S3 + index harness.

## Related (secondary, from the same review)

`internal/reconstruct/fulltable.go`: the `ErrNoBaseline` → `reconstructBinlogOnly` fallback runs **before** the PG full-table gate (#830), so a full-table `reconstruct` of a PG table that is absent from the baseline source produces binlog-only output instead of the "#597 not supported for PostgreSQL" refusal. Narrow (full-table mode + baseline source given + requested table missing from it) and pre-existing degrade parity with the shim, but it means the PG full-table gate isn't airtight.

---
_Found during the PR review of #830 / #829 (post-full-audit reconstruct-surface work)._

nethalo:
closes #916

Two PG reconstruct gaps surfaced during the #829/#830 review (both shipped-as-deferred there).

## 1. Single-row reconstruct read baseline metadata **local-only**
`internal/cli/reconstruct.go` read metadata under `if !strings.HasPrefix(baselinePath, "s3://")` → for an S3 baseline `bmeta` stayed zero:
- **PG beta warn (#829) could not fire for an S3 PG baseline** — the gate is `SourceFlavor(db)=="postgres" || bmeta.LSN != 0`; with `bmeta.LSN==0` for S3, detection rested entirely on the flavor probe, so a `""` probe (index with no `stream_state` row / old schema / transient blip) left beta-fidelity PG output with **no caveat** (the warn is the only guardrail — warn-not-refuse).
- **Gap detection was disabled for every S3 baseline** (MySQL and PG) — `resolveGapCheck` had no anchor.

**Fix:** use `baseline.ReadParquetMetadataAny` (S3-capable — the exact reader `fulltable.go` already uses). An S3 baseline now populates `bmeta`, so the LSN clause is load-bearing for S3 and gap detection runs against its anchor. Read failure stays non-fatal. **Local behavior is unchanged** (`ReadParquetMetadataAny` delegates to `ReadParquetMetadata` for non-`s3://`). #829's beta-warn / `pgReconstructBeta` / test comments updated to match.

## 2. Full-table PG reconstruct with no baseline fell through instead of refusing
`fulltable.go`'s `ErrNoBaseline → reconstructBinlogOnly` fallback ran **before** the PG gate (#830), so a PG full-table reconstruct of a table absent from the baseline produced a binlog-only (delta-only) report mislabeled as full-table. Added the PG-flavor refusal in that branch too (no baseline → no LSN, detect by flavor), mirroring the with-baseline gate.

## Testing
- Local metadata read is a **proven no-op** — covered by the existing reconstruct integration test.
- The **S3-transport** behavior is validated-by-reuse: the same `ReadParquetMetadataAny` S3 branch `fulltable.go` already ships. This repo has no local S3/minio harness (a documented repo-wide gap — see `parquetquery_s3direct_integration_test.go`).
- The two `SourceFlavor(db)=="postgres"` branches need a live PG-flavored index (the same accepted gap as #830's existing flavor clause).
- `go build ./...` / `go vet` / `gofmt` clean; `reconstruct` + `cli` unit suites green.

### DIFF
--- a/internal/cli/reconstruct.go
+++ b/internal/cli/reconstruct.go
@@ -209,14 +209,19 @@ func runReconstruct(cmd *cobra.Command, args []string) error {
 	}
 	slog.Debug("found baseline snapshot", "path", baselinePath, "snapshot_time", snapshotTime.UTC().Format(time.RFC3339))
 
-	// Read baseline binlog position metadata (local files only).
+	// Read baseline position metadata (binlog file/pos for MySQL, the WAL LSN
+	// anchor for PG). ReadParquetMetadataAny reads S3 too — the same reader the
+	// full-table path uses (fulltable.go) — so an S3 baseline populates bmeta,
+	// which the old local-only read silently skipped: without it, gap detection
+	// and the PG beta warn below were disabled for every S3 baseline (#916). A
+	// read failure is non-fatal — warn and proceed with a zero bmeta (gap
+	// detection then reports "unavailable", the pre-#916 S3 behavior).
 	var bmeta baseline.DumpMetadata
-	if !strings.HasPrefix(baselinePath, "s3://") {
-		var metaErr error
-		bmeta, metaErr = baseline.ReadParquetMetadata(baselinePath)
-		if metaErr != nil {
-			slog.Warn("could not read baseline metadata", "error", metaErr)
-		} else if bmeta.BinlogFile != "" {
+	if bm, metaErr := baseline.ReadParquetMetadataAny(cmd.Context(), baselinePath); metaErr != nil {
+		slog.Warn("could not read baseline metadata", "error", metaErr)
+	} else {
+		bmeta = bm
+		if bmeta.BinlogFile != "" {
 			slog.Debug("baseline binlog position",
 				"file", bmeta.BinlogFile, "pos", bmeta.BinlogPos, "gtid", bmeta.GTIDSet)
 		}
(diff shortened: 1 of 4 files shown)


nethalo:
Reviewed by silent-failure-hunter + code-reviewer: **sound / clean**, net-reduces silent failures (removes the S3 metadata skip, closes the mislabeled-full-table gap). Applied the code-reviewer's low-severity suggestion — added `TestReconstructTable_pgNoBaselineRefused` (sqlmock flavor='postgres' + empty baseline dir → asserts the #597 refusal). One noted LOW residual (not blocking, out of scope): the no-baseline PG refusal fails-open if the flavor probe *errors* transiently (no LSN backstop there, unlike the with-baseline gate) — consistent with the codebase's documented `SourceFlavor` fail-open contract; closing it fully would need a `SourceFlavor` API change to distinguish error from a clean non-PG result.

Issue #594, opened 2026-06-23
[pg] Managed-PostgreSQL smoke matrix in CI (RDS/Aurora/Cloud SQL/Azure) (GA)


nethalo:
## Summary

Add a **managed-PostgreSQL smoke matrix** to CI (RDS / Aurora / Cloud SQL / Azure). Today managed setup is documented but **not validated by any automated test**. **GA gate** (epic #527 Beta-DEFERRABLE — "beta targets self-hosted + one managed; expand at GA").

## Current state (verified)

- **Doc yes:** `docs/postgres.md:424-437` has a "Managed PostgreSQL" table (RDS/Aurora/Cloud SQL/Azure `wal_level` + replication-grant setup) plus a troubleshooting row (added by PR #565).
- **Test no:** `.github/workflows/ci.yml` job `integration-postgres-source` runs **stock** `postgres:["14","15","16","17"]` container images only — pure self-hosted. No RDS/Aurora/Cloud SQL/Azure smoke anywhere in `cmd/bintrail-pg`, `internal/pgcapture`, `internal/pgstreamrun`.

## Why managed differs (and is worth a real smoke)

Managed flavors gate/alter logical replication: RDS needs `rds.logical_replication=1` + the `rds_replication` grant; Aurora has its own switch; Cloud SQL uses `cloudsql.logical_decoding`; Azure Flexible Server uses `wal_level=logical` via server params. The `CREATE_REPLICATION_SLOT` / `pgoutput` path, RI-FULL enforcement, and slot-lag/WAL-retention monitoring (`max_slot_wal_keep_size`) should be smoke-tested against at least one managed flavor before claiming managed support.

## Work

- [ ] Pick the first managed target (RDS Postgres is the cheapest/most-common ICP) and add a gated smoke job (provision → publication + slot → capture an INSERT/UPDATE/DELETE → assert it lands in the index → recover) — likely opt-in / scheduled, not per-PR, given cost + credentials.
- [ ] Expand to Aurora / Cloud SQL / Azure as ICP evidence warrants (see #535).

Part of epic #527 (GA).

PR #920, opened 2026-07-09
pg: pin rendering GUCs on capture + baseline sessions; fold-validated type matrix (#593 slice D)

### LINKED ISSUE #593: [pg] Baseline at slot consistent_point LSN → enable reconstruct/time-travel for PostgreSQL (GA)

nethalo:
Closes #593 — slice D, the last open GA-gating item of epic #597.

## The correctness hole this closes

The PostgreSQL baseline+delta design rests on one identity: **baseline COPY text ≡ pgoutput delta text** — reconstruct's PK join and last-write-wins merge are exact string operations over that text. But both sides render every value through the type's server-side **output function**, whose text is **session-GUC-dependent** (`TimeZone`, `DateStyle`, `extra_float_digits`, `bytea_output`, `IntervalStyle`), and neither `internal/pgbaseline` (COPY connections) nor `internal/pgcapture` (walsender session) pinned any GUC. The same logical value could render as different text on the two sides — silently breaking the merge for `timestamptz`/`timestamp`/`date`/`time`, `real`/`double precision`, `bytea`, `interval`.

## The fix

**`internal/pgcapture/gucpin.go`**: every connection that renders row text is pinned to `TimeZone=UTC, DateStyle=ISO, extra_float_digits=3, bytea_output=hex, IntervalStyle=postgres` via **startup-packet RuntimeParams** — the same channel `replication=database` itself travels, so it provably reaches the walsender backend. Startup params apply as `PGC_S_CLIENT` (beats `ALTER DATABASE/ROLE` defaults and the DSN `options` string — both empirically proven by the test below); an invalid value FATALs at connect (never a silently-unpinned session); startup placement leaves pgbaseline's REPEATABLE READ anchor / exported-snapshot workers untouched. Wired: the walsender, the pgbaseline anchor conn, every parallel COPY worker, the slot-create conn. Catalog-only connections deliberately stay unpinned.

**Pre-pin baseline detection**: pgbaseline stamps the pinned set into the Parquet metadata (`bintrail.render_gucs`); single-row reconstruct warns (with re-baseline guidance) when an LSN-anchored baseline lacks the stamp.

**#829 beta warn narrowed**: scalars including the GUC-sensitive set are now fold-validated; the residual beta surface is container types only.

## Fail-before / pass-after evidence

The PR is two commits, per protocol:

- **Commit 1 (`272317b`) — fail-before**: the fold matrix (`TestOne_PGTypeMatrixThroughReconstructFold`: full type matrix through `FindBaseline → ReadBaselineRow → Fetch → ApplyAt`, three rows per type, under a two-sided GUC skew — `ALTER DATABASE` defaults hitting the walsender + baseline conns, and an operator `options=-c` string on the baseline DSN) ran **without the pin**. Result — [run 29055029997](https://github.com/dbtrail/dbtrail/actions/runs/29055029997): red on **exactly the GUC-sensitive set** (`timestamptz`, `timestamp`, `date`, `interval`, `real`, `double`, `bytea`) on all four PG versions (14/15/16/17); everything else green, including the newly-CI-enabled `internal/pgbaseline` integration suite. Two additional reds (`boolean`, `inet`) were **oracle-cast artifacts**, not fidelity bugs: the fold output was internally consistent (`t`/`t`, `192.168.1.10`/`192.168.1.10`), but the oracle's `::text` CAST diverges from the output function (`bool::text`="true" vs `boolout`="t"; `inet::text` appends `/32`). Commit 2 fixes the oracle to read the output-function rendering via the simple protocol (`RawValues`) — a no-op for the GUC-sensitive types (their cast IS the output function), so the recorded fail-before evidence stands.
- **Commit 2 (`c850720`) — pass-after**: the pin + the oracle fix + docs. The CI run at this commit must be green on all four PG cells — that green IS the empirical proof that (a) the walsender accepts startup GUCs on PG 14–17, and (b) the pin beats both `ALTER DATABASE` defaults and the DSN `options` string.

## Upgrade note (also in CHANGELOG)

On servers whose defaults differ from the pin, baselines taken before this release were rendered unpinned and will not text-join post-upgrade deltas for GUC-sensitive types — re-run `bintrail-pg baseline` after upgrading (the reader now warns on a pre-pin baseline). Events indexed pre-upgrade keep their old rendering.

## Out of scope (unchanged)

Full-table `reconstruct` / baseline-anchored `verify` stay gated for PG (#830/#916); container types (arrays beyond `integer[]`/`text[]`, composite, range/multirange, hstore, geometric) remain beta; `lc_monetary` deliberately not pinned (`money` keeps its locale caveat). Stock PG 14–17 in CI validates the GUC mechanics; the gated #594 managed-smoke workflow (RDS+Aurora, billed) can be triggered on the merged commit as additional closing evidence — owner's call.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -339,6 +339,8 @@ jobs:
             ./internal/pgcapture/ ./internal/pgstreamrun/ ./internal/pgbaseline/ 2>&1 | tee /tmp/pg-it.out
           grep -q -- '--- PASS: TestOne_EndToEnd_PostgresToIndexToRecovery' /tmp/pg-it.out \
             || { echo "FATAL: PG end-to-end integration test did not run — false-green tripwire" >&2; exit 1; }
+          grep -q -- '--- PASS: TestPGBaseline_Integration' /tmp/pg-it.out \
+            || { echo "FATAL: pgbaseline integration suite did not run — false-green tripwire" >&2; exit 1; }
 
   console-e2e:
     # Headless-Chrome regression guard for the console SPA: go test never
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,7 +336,7 @@ jobs:
         run: |
           set -o pipefail
           go test -tags=integration -race -p 1 -count=1 -v \
-            ./internal/pgcapture/ ./internal/pgstreamrun/ 2>&1 | tee /tmp/pg-it.out
+            ./internal/pgcapture/ ./internal/pgstreamrun/ ./internal/pgbaseline/ 2>&1 | tee /tmp/pg-it.out
           grep -q -- '--- PASS: TestOne_EndToEnd_PostgresToIndexToRecovery' /tmp/pg-it.out \
             || { echo "FATAL: PG end-to-end integration test did not run — false-green tripwire" >&2; exit 1; }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **PostgreSQL rendering GUCs pinned on capture and baseline sessions** (#593 slice D). The baseline+delta reconstruct design rests on the identity *baseline COPY text ≡ pgoutput delta text* — the PK join and last-write-wins merge are exact string operations — but both sides render values through PostgreSQL's session-GUC-dependent output functions, and neither session pinned any GUC: the same logical value could render as different text on the two sides, silently breaking the merge for `timestamptz`/`timestamp`/`date`/`time`, `real`/`double precision`, `bytea`, and `interval`. Every connection that renders row text — the logical-decoding (walsender) session and the baseline COPY connections — is now pinned via startup parameters to `TimeZone=UTC`, `DateStyle=ISO`, `extra_float_digits=3`, `bytea_output=hex`, `IntervalStyle=postgres`; the pinned set is stamped into the baseline Parquet metadata (`bintrail.render_gucs`), and single-row `reconstruct` warns when it reads a pre-pin baseline. The full type matrix is now validated **through the single-row reconstruct fold** under deliberately skewed session GUCs on the PG 14–17 CI matrix (`TestOne_PGTypeMatrixThroughReconstructFold`); the single-row PG beta warning narrows to container types only.
+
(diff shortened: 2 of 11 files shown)


nethalo:
Adversarial review (2 lenses completed; pin-blast-radius + test-integrity): **1 must-fix found and applied in `97deba1`** — `PinRenderGUCs` compared keys case-sensitively, but PostgreSQL GUC names are case-insensitive: an operator DSN `?timezone=...` landed under a different RuntimeParams key than the pinned `TimeZone`, both reached the startup packet in random map order, and the server applies them last-writer-wins → **~half of all connections silently unpinned, nondeterministically**. Fixed with a case-insensitive purge + `TestPinRenderGUCs_caseInsensitiveCollision`. Also applied: stamp **value** comparison via `baseline.RenderGUCsPinned` (read-layer constant, cross-pinned by unit test — `internal/cli` must not import pgcapture/pgx), RESET-cleanup ordering in the fold test, a second false-green tripwire for the newly-enabled pgbaseline suite, and the event-side PK-discontinuity docs note. Verified non-issues with evidence: reverse-SQL re-parses under any operator session (explicit offsets / unambiguous ISO); all unpinned conns are catalog-only; no verify false-MISMATCH surface. **Deferred (warn-parity, not corruption):** shim `_snapshot` / console reconstruct fold baselines without the pre-pin stamp warn the CLI has.

PR #924, opened 2026-07-10
stream: refuse position-mode resume when saved binlog position exceeds uint32

### LINKED ISSUE #845: capture: saved binlog position truncated to uint32 when resuming a >4GB binlog
`stream_state.binlog_position` is BIGINT UNSIGNED and `streamState.binlogPos` is uint64, but `resolveStartForFlavor` does `uint32(saved.binlogPos)` because `gomysql.Position.Pos` is uint32 (COM_BINLOG_DUMP protocol limit). A binlog can exceed 4GB when a single transaction exceeds `max_binlog_size` (the file only rotates at transaction end): a 5GB bulk load leaves checkpoints with pos > 2^32.

Location: `internal/streamrun/streamrun.go:374`

On resume, the position wraps silently to an arbitrary earlier offset: most likely a noisy decode error, but if the wrapped offset lands on a valid event boundary, history is re-indexed/duplicated with no warning.

Fix: detect `saved.binlogPos > math.MaxUint32` and fail loud with instruction to use GTID mode (which has no such limit) instead of casting.

---

nethalo:
## What

`resolveStartForFlavor` resumed position mode with `uint32(saved.binlogPos)`. `stream_state.binlog_position` is BIGINT UNSIGNED (uint64 in Go), but `gomysql.Position.Pos` is uint32 because COM_BINLOG_DUMP carries the offset as a 4-byte field. A binlog can exceed 4GiB when a single transaction exceeds `max_binlog_size` (the file only rotates at transaction end), leaving a checkpoint whose position wraps silently on resume — most likely a noisy decode error, but if the wrapped offset lands on a valid event boundary, history is re-indexed/duplicated with no warning.

## Fix

Guard the position-mode resume path: `saved.binlogPos > math.MaxUint32` now fails loud with an actionable error telling the operator to switch to GTID mode (`--start-gtid` with the source's `gtid_executed`), which has no such limit. The suggested remediation works by construction — the `--start-gtid` mode-switch branch runs before the guard, so an oversized saved position is always escapable. Exactly `MaxUint32` still resumes. This is the only `uint32(` narrowing in the package.

## Tests

- `TestResolveStart_resumePositionOver4GiBFailsLoud` — overflow errors, message names the file and points at GTID mode (fails on the pre-fix code: the cast wrapped to 0 and returned nil error)
- `TestResolveStart_resumePositionExactlyMaxUint32` — boundary still resumes
- `TestResolveStart_positionOver4GiB_gtidFlagStillSwitches` — the suggested escape hatch works with an oversized saved position

```
go build ./...                                                ok
go vet ./internal/streamrun/                                  ok
go test ./internal/streamrun/... -count=1                     ok (0.8s)
go test -tags integration ./internal/streamrun/... -count=1   ok (32.9s)
```

Closes #845

### DIFF
--- a/internal/parser/querytext_test.go
+++ b/internal/parser/querytext_test.go
@@ -91,7 +91,11 @@ func TestStreamParser_rowsQueryTextAttachedToRows(t *testing.T) {
 		makeQueryEvent("BEGIN"),
 		makeRowsQueryEvent("INSERT INTO shop.orders VALUES (1, 10)"),
 		makeOrdersInsertEvent(1, 10),
-		makeXIDEvent(300),
+		// XID logPos matches makeOrdersInsertEvent's hardcoded LogPos:200 (not
+		// a realistic gap) — this test asserts QueryText propagation, not
+		// position tracking; must stay >= the preceding insert's LogPos so it
+		// doesn't trip the #845 same-file-backward-position guard.
+		makeXIDEvent(200),
 		// Transaction 2: variable toggled off — no ROWS_QUERY event.
 		makeGTIDEvent(2),
 		makeQueryEvent("BEGIN"),
--- a/internal/parser/stream.go
+++ b/internal/parser/stream.go
@@ -30,6 +30,12 @@ type StreamParser struct {
 	// decoded. See SetSyncDDLHook for why this must not move off the parse
 	// path, and why the ordering (hook, then emit) is load-bearing for #760.
 	onDDL atomic.Pointer[func(Event) error]
+	// flavor is the source flavor ("mysql" or "mariadb"; see SetFlavor), used
+	// only to word remediation text (GTIDExecutedHint) in the position-
+	// wraparound error inside Run. Never consulted for parsing decisions —
+	// MySQL vs MariaDB event types (GTIDEvent vs MariadbGTIDEvent) already
+	// disambiguate that.
+	flavor string
 }
 
 // NewStreamParser creates a StreamParser that resolves column names via
@@ -53,6 +59,30 @@ func NewStreamParser(resolver *metadata.Resolver, filters Filters, logger *slog.
 func (sp *StreamParser) SwapResolver(r *metadata.Resolver) {
 	sp.schemaVersion.Store(uint32(r.SnapshotID()))
 	sp.resolver.Store(r)
+}
+
+// SetFlavor sets the source flavor ("mysql" or "mariadb", matching
+// gomysql.MySQLFlavor/MariaDBFlavor) used to word the position-wraparound
+// error (see Run) flavor-appropriately. Must be called before the goroutine
+// that runs Run is spawned — Run reads it without synchronization, relying
+// on the happens-before edge from the spawning statement (mirrors how
+// resolver/filters are fixed before Run begins). Empty (never called)
+// defaults to MySQL wording.
+func (sp *StreamParser) SetFlavor(flavor string) {
+	sp.flavor = flavor
+}
+
+// GTIDExecutedHint returns the flavor-appropriate system variable an
+// operator should read to obtain the source's current executed GTID set, for
+// remediation messages that direct the operator to --start-gtid. MySQL has
+// gtid_executed; MariaDB has no such variable — its analog is
+// gtid_binlog_pos (see detectMariaDBGTIDGap in internal/streamrun, which
+// queries the same variable for gap detection).
+func GTIDExecutedHint(flavor string) string {
+	if flavor == "mariadb" {
+		return "SELECT @@gtid_binlog_pos"
+	}
+	return "SELECT @@GLOBAL.gtid_executed"
 }
 
 // SetSyncDDLHook registers fn to run synchronously inside Run, for each DDL
@@ -105,6 +135,10 @@ func (sp *StreamParser) Run(ctx context.Context, streamer *replication.BinlogStr
 	var currentFile string
 	var currentGTID string
 	var currentConnectionID uint32 // pseudo_thread_id from most recent QueryEvent
+	// lastLogPos tracks the highest EventHeader.LogPos seen since the last
+	// RotateEvent, to detect a same-file position wraparound (#845) live,
+	// during streaming — see the check at the top of handleEvent below.
+	var lastLogPos uint32
 	// currentQueryText holds the original SQL statement from the most recent
 	// ROWS_QUERY_EVENT (MySQL, binlog_rows_query_log_events=ON) or
 	// ANNOTATE_ROWS event (MariaDB, binlog_annotate_row_events=ON; the syncer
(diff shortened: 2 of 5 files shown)


nethalo:
Two independent adversarial reviews converged on the same finding, so I dug in and fixed it. Posting the gap + the fix for the record.

**The gap in `152d9167`:** the guard (`saved.binlogPos > math.MaxUint32` before `uint32(saved.binlogPos)` in `resolveStartForFlavor`) is placed correctly but is **unreachable**. Every writer of `binlogPos`/`stream_state.binlog_position` derives it from `replication.EventHeader.LogPos`, which is itself a `uint32` wire field (`internal/parser/stream.go` sets `EndPos: uint64(hdr.LogPos)` — a uint32 upcast). So `saved.binlogPos` can never exceed `math.MaxUint32` through this codebase's own capture path; the guard checks a condition that provably can't happen. The real corruption happens **upstream, at the source**: when a single oversized transaction delays rotation past 4GiB (`max_binlog_size` is only enforced between transactions), MySQL/MariaDB itself truncates `end_log_pos` on the wire once the true offset crosses 2^32 — bintrail receives an already-wrapped small position with zero signal, and on resume it would silently start re-indexing from the wrong offset.

**The fix (`e3807035`):** detection moved to where it can actually fire — live, during streaming, not at resume time.
- `StreamParser.Run` (`internal/parser/stream.go`) now tracks the highest `LogPos` seen since the last `RotateEvent`. A same-file position going **backward with no intervening rotation** is the unmistakable signature of a 32-bit wraparound (legitimate positions are strictly increasing within a file). On detection it fails loud immediately, before the offending event's rows are captured and before any checkpoint can advance past it — closing the window the resume-time guard left open.
- I verified there's no way to recover a true 64-bit position instead: `RotateEvent.Position` is an 8-byte field but is only meaningful at rotation, and `COM_BINLOG_DUMP` resume is itself capped at 4 bytes — so even an internally-tracked running offset couldn't be used to resume position-mode past the wrap point. GTID mode remains the only escape, same as the original guard's remediation.
- Kept the resume-time guard as cheap defensive insurance (documented as such — it's not the primary fix anymore) in case a future writer stops deriving from `LogPos`.

**Also fixed (should-fix from review):** the remediation text hardcoded `SELECT @@GLOBAL.gtid_executed`, which errors on MariaDB (`Unknown system variable`). Both guards now go through a new `parser.GTIDExecutedHint(flavor)` helper, which recommends `SELECT @@gtid_binlog_pos` for MariaDB — the same variable `detectMariaDBGTIDGap` already queries for gap detection.

**Tests:** added `internal/parser/wraparound_test.go` — three tests exercising the real `StreamParser.Run` capture-time path (via `AddEventToStreamer` + `Run`, not a hand-built state struct):
- `TestStreamParser_positionWraparoundFailsLoud` — reproduces the issue's own scenario (a transaction commit near the uint32 ceiling, followed same-file/no-rotate by a commit at a small wrapped value) and asserts `Run` fails loud, naming the file and pointing at GTID mode.
- `TestStreamParser_positionWraparound_mariadbHint` — same scenario under `SetFlavor("mariadb")`, asserts the message recommends `gtid_binlog_pos` and never suggests `gtid_executed`.
- `TestStreamParser_positionResetByRealRotate` — negative test: the same backward jump across a genuine `RotateEvent` does NOT false-positive (fresh files legitimately restart small).

One existing test (`TestStreamParser_rowsQueryTextAttachedToRows`) hardcoded a synthetic XID position that happened to go backward relative to a shared fixture's fixed `LogPos:200` — a test-fixture artifact, not a real regression; fixed with a one-line position bump + comment.

**Residual limitation, stated plainly:** this is fail-loud detection, not true 64-bit position recovery — there is no such value to recover from the wire once MySQL itself has wrapped it, so a stream that hits this must restart in GTID mode. That's the same remediation the original guard already prescribed; the change is *where* the check actually fires.

`go build ./...`, `go vet ./...`, full unit suite, and `-tags integration` for `internal/streamrun` and `internal/parser` all pass.

nethalo:
Follow-up after a second pass: verified the new guard doesn't false-positive on the two realistic scenarios that matter most (idle heartbeats, reconnect), and want to state two residual limitations plainly rather than leave them implicit.

**Verified safe (checked against go-mysql v1.13.0 source, not just assumed):**
- **Heartbeats on an idle connection**: `HEARTBEAT_EVENT`/`HEARTBEAT_LOG_EVENT_V2` aren't given a type-specific case in `BinlogParser.parseEvent` — they decode as `GenericEvent`, and their `EventHeader.LogPos` comes from the same generic wire decode every event uses. go-mysql's own reconnect bookkeeping (`binlogsyncer.go: handleEventAndACK`, `if e.Header.LogPos > 0 { b.nextPos.Pos = e.Header.LogPos }`) trusts that value as the master's current, monotonic position for *every* event type uniformly — it has to, since that's what drives its own resume-position tracking. So the guard's assumption (LogPos is monotonic-or-zero) is load-bearing for go-mysql's own correctness too, not just mine.
- **FormatDescriptionEvent's LogPos=0**: go-mysql's own comment right next to that line — `// Some events like FormatDescriptionEvent return 0, ignore.` — confirms real, non-Transaction_payload events can legitimately carry `LogPos=0` on the wire. That's exactly what my `hdr.LogPos != 0` skip-guard is for (originally added for Transaction_payload inner events per `rewriteInnerHeader`'s doc comment); good to have independent confirmation it's the right general guard, not a narrow one.
- **Reconnect**: `retrySync`/`prepareSyncPos` always force `pos.Pos >= 4` and resume via `COM_BINLOG_DUMP`/`COM_BINLOG_DUMP_GTID`, and `handleEventAndACK`'s `case *RotateEvent:` is the only place `b.nextPos.Name` (the tracked current file) gets updated — meaning go-mysql's own correctness already depends on the server sending a (possibly fake/artificial) `RotateEvent` at the start of every dump session. My guard resets `lastLogPos = 0` on any `RotateEvent`, so a reconnect's first real event afterward is never compared against stale pre-reconnect state.
- Also confirmed `FillZeroLogPos` (a MariaDB-11.4+-only opt-in that makes go-mysql itself synthesize a `LogPos` via `nextPos.Pos + EventSize`, still `uint32`-typed) is **not** set in bintrail's `BinlogSyncerConfig` (`internal/streamrun/streamrun.go`), so that path is inert for us — one less interaction to reason about.

**Two residual limitations, stated plainly (neither is new — both are inherent to detecting this live rather than fully recovering a true 64-bit position):**

1. **Reconnect exactly during an in-progress wraparound is a false negative.** go-mysql's `BinlogSyncer` resumes using its *own* internally tracked `nextPos.Pos`, which is itself `uint32` — if a connection drops mid-transaction right as the true offset crosses 4GiB, that internal value has already wrapped by the time of reconnect. The fake `RotateEvent` the server sends on resume would carry that already-wrapped (small) position, my guard resets its baseline to it, and subsequent events "look" monotonic from that wrong baseline — undetected. This is bounded by go-mysql's own `uint32` position type and isn't fixable without patching/forking the dependency. The common case the issue actually describes — a continuous connection through one oversized transaction, no drop — **is** caught, which is the scenario `#845` was filed against.
2. **MariaDB wire events can legitimately carry `LogPos=0`** outside the Transaction_payload/FDE cases (per go-mysql's own `shouldCalculateDynamicLogPos` doc comment: "MariaDB 11.4+ ... event has LogPos=0, indicating server didn't set it"). Our guard treats `LogPos=0` as "no signal" — never a false positive, but no detection power for that specific event either. Coverage is best-effort on such connections; this doesn't regress anything (there was zero wraparound detection before this PR at all), it's just not airtight everywhere.

Net: this remains fail-loud detection of the common/documented failure mode, not full 64-bit recovery — consistent with what I said in the previous comment, just making the edges explicit instead of leaving them implicit.

Issue #511, opened 2026-06-17
ultrafast Tier 2: S3 httpfs-direct + parallel per-file DuckDB queries


nethalo:
Part of #509 — the big S3 win, with **real OOM risk**. Do this only after Tier 1, and only behind the explicit `--ultrafast` opt-in (or a narrower `--ultrafast-s3`). Default stays the safe download-then-query path.

## What

Two S3-path levers that the container-safe defaults deliberately give up:

### 1. httpfs-direct (skip the local download)

Today, S3 Parquet files are downloaded to local disk via the AWS SDK, then queried by DuckDB from disk (`internal/parquetquery/parquetquery.go:62`; baseline equivalent `materializeBaselineLocal` in `internal/reconstruct/fulltable.go:712`). Ultrafast → let DuckDB read S3 directly via `parquet_scan('s3://…')` / httpfs, removing the double I/O (download-to-disk then read-from-disk).

**⚠️ Caveat from our own code comment (`parquetquery.go:62-65`):** httpfs holds the entire S3 file in memory **outside `memory_limit` tracking**. So even the Tier 1 sane `memory_limit` does NOT protect this path — this is why it is the riskiest lever and must be explicit. Mitigation to document and ideally pre-check: `available_RAM ≥ largest_file_size × concurrency`.

### 2. Parallel per-file DuckDB queries

Today queries are strictly sequential — only one DuckDB query active at a time (`parquetquery.go:83-97`), which keeps peak working set constant. Ultrafast → run per-file queries concurrently (multiplies peak memory by the degree of parallelism, but much faster).

**Note:** the `buildUnsortedQuery` machinery (`parquetquery.go:605`, currently latent / tests-only) was built for exactly the "sort after merging all files" path that parallel per-file queries need — the per-file `ORDER BY` in `buildQueryForFile` assumes sequential merge. Wiring parallel queries means moving the final sort to post-merge and switching those calls to the unsorted variant.

## Risk gating

- Must be explicit (`--ultrafast`); never the default.
- Consider a pre-flight RAM check that refuses (or warns loudly) when `largest_file × concurrency` exceeds available RAM, since `memory_limit` cannot catch the httpfs path.
- Keep the early-termination (`parquetquery.go:122`) and date-scoped listing (`parquetquery.go:195`) intact — those are pure wins, unaffected by this change.

## Tests

- httpfs-direct path returns results identical to the download path against a test S3 (minio) archive.
[...]

PR #925, opened 2026-07-10
archive: share one region-pinned DuckDB session across reconcile --deep footer probes

### LINKED ISSUE #807: archive: reconcile --deep opens a fresh DuckDB session (INSTALL httpfs + aws + CREATE SECRET) per S3 object, with no region pin
`s3ParquetRowCount` does `sql.Open('duckdb')+LoadHTTPFS+EnableS3CredentialChain` per object. A monthly `reconcile --deep` over a year of hourly archives = ~8760 DuckDB sessions, each with extension setup and credential resolution (IMDS/SSO) — tens of minutes and thousands of IMDS hits where a single reused session would do millisecond footer probes.

Location: `cliapp/archive.go:305`

Also: `EnableS3CredentialChain` without a region pin (unlike `EnableS3CredentialChainRegion`, #511) makes every probe on cross-region buckets fail with 301 → all objects fall to `DeepUnverified` and the cron monitor is permanently red with no explanation.

Fix: one shared DuckDB session for the whole scan, with the region pinned.

---

nethalo:
## What

`archive reconcile --deep` opened a fresh DuckDB session per S3 object in `s3ParquetRowCount` — `sql.Open("duckdb")` + `INSTALL/LOAD httpfs` + `INSTALL/LOAD aws` + `CREATE SECRET` for every footer probe — and created the credential-chain secret with no region pin.

Two consequences:

- **Cost**: a monthly `--deep` over a year of hourly archives ≈ 8760 sessions, each paying extension setup and credential-chain resolution (IMDS/SSO round trips) for a millisecond footer read.
- **Correctness**: without a region pin the chain secret resolves region from the AWS SDK config (e.g. `AWS_REGION`), not the bucket. On a cross-region bucket every probe fails with 301/PermanentRedirect, every object falls to `DeepUnverified`, and the cron monitor is permanently red with no explanation.

## How

- `scanS3Archive` now opens **one** DuckDB session for the whole scan (`openS3FooterSession`: single pooled connection, httpfs loaded once) and passes it to each probe (`duckdbParquetRowCount`).
- The session's chain secret pins the scan's resolved region via `duckdbutil.EnableS3CredentialChainRegion` (the #511 precedent). The region comes from the same `storage.LoadAWSConfig` resolution (flag > env/config > IMDS fallback) that backs the `ListObjectsV2` listing — a successful listing already proves that region matches the bucket, since the SDK does not follow S3's cross-region redirects.
- Failure semantics unchanged: if the shared session cannot be opened, the scan warns once and completes with row counts unset, and the decision layer counts each object deep-unverified (#469) exactly as a per-object failure did before.

## Tests

- `TestDuckDBParquetRowCountSharedSession` — multiple footer probes (incl. a quote-escaped path) on one caller-owned session, over local parquet written by DuckDB itself.
- `TestOpenS3FooterSessionRegionPin` — the shared session comes up usable and the chain secret carries `region=eu-west-1` (stripping the pin fails this test); offline hosts skip, mirroring the `internal/duckdbutil` convention.

```
go build ./...                                  ok
go vet ./cliapp/                                ok
go test ./cliapp/... -count=1                   ok (8.2s)
go test -tags integration ./cliapp/... -count=1 ok (12.2s)
```

Both new tests ran non-skipped locally (extensions cached).

Closes #807

### DIFF
--- a/cliapp/archive.go
+++ b/cliapp/archive.go
@@ -249,9 +249,29 @@ func scanS3Archive(ctx context.Context, s3URL, region string, deep bool) ([]arch
 	if err != nil {
 		return nil, err
 	}
-	client, err := storage.NewS3Client(ctx, region)
+	cfg, err := storage.LoadAWSConfig(ctx, region)
 	if err != nil {
 		return nil, err
+	}
+	client := s3.NewFromConfig(cfg)
+
+	// ONE DuckDB session serves every --deep footer probe of the scan (#807):
+	// extension install and credential-chain resolution happen once, not per
+	// object (a year of hourly archives = thousands of sessions and IMDS
+	// round-trips otherwise), and the chain secret pins the listing's resolved
+	// region (#511) so probes on a cross-region bucket don't 301 every object
+	// into DeepUnverified. A failed open degrades exactly like a failed probe:
+	// row counts stay unset (warned) and the decision layer counts each object
+	// deep-unverified (#469) — the scan itself still completes.
+	var footerDB *sql.DB
+	if deep {
+		fdb, err := openS3FooterSession(ctx, cfg.Region)
+		if err != nil {
+			slog.Warn("reconcile: cannot open DuckDB session for --deep footer reads; row counts left unset", "error", err)
+		} else {
+			footerDB = fdb
+			defer footerDB.Close()
+		}
 	}
 
 	var out []archive.ScannedFile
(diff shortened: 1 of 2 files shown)


nethalo:
Applied a confirmed review finding on top of this branch (commit 942d6c0):

`openS3FooterSession` creates the AWS credential-chain secret **once** for the whole `--deep` scan session, but `duckdbutil.EnableS3CredentialChain`'s own doc comment warns: "The secret resolves credentials at CREATE time, not per request ... do not reuse this on long-lived pooled sessions under expiring roles." A long scan (thousands of hourly archive objects, 15-45min) under an expiring IMDS/STS role could have credentials expire mid-scan -- every remaining probe 403s, falls to `DeepUnverified`, and the cron monitor goes red, i.e. the same symptom class #807 was filed to fix. The pre-#807 per-object sessions picked up rotated credentials naturally; the shared session regressed that.

Fix: `scanS3Archive` now tracks when the shared session's chain secret was last issued and re-issues it (`CREATE OR REPLACE SECRET` via `EnableS3CredentialChainRegion`) every `s3FooterSecretRefreshInterval` (10min, well under common IMDS/STS credential lifetimes) as it walks the object list, instead of freezing it for the whole scan. The single-shared-session performance win (one extension load/install for the whole scan) is kept -- this doesn't revert to a fresh DuckDB session per object.

Added:
- `dueForS3SecretRefresh` -- pure threshold function, unit tested (`TestDueForS3SecretRefresh`).
- `TestS3FooterSessionSecretReissue` -- exercises the actual mechanism: re-issuing the secret on an already-open session stays usable and still carries the pinned region (mirrors `TestOpenS3FooterSessionRegionPin`).

Verified: `go build ./...`, `go vet ./cliapp/...`, `go test ./cliapp/... -count=1`, `go test -tags integration ./cliapp/... -count=1` all pass.

PR #926, opened 2026-07-10
recover: emit LIMIT 1 on all-columns-fallback DELETE/UPDATE so duplicate rows aren't all hit

### LINKED ISSUE #789: recover: exact-duplicate rows in PK-less tables — the reversal hits ALL copies, not one (no LIMIT 1)
In PK-less tables (or with no snapshot) the WHERE uses all columns WITHOUT `LIMIT`: if byte-identical duplicate rows exist, the reverse DELETE of ONE INSERT deletes ALL duplicates (loss of pre-existing rows) and the reverse UPDATE modifies them all.

Location: `internal/recovery/recovery.go:846`

MySQL supports `DELETE/UPDATE ... LIMIT 1` (PG would need `ctid`) → buildable. Partially documented: `docs/query-and-recovery.md:313` says "always correct for tables without duplicate rows" and the code comment admits the assumption, but there is no detection or warning when the assumption fails — the script applies and over-deletes silently.

Fix: emit `LIMIT 1` on the fallback DELETE/UPDATE for MySQL (or detect duplicates and warn).

---

nethalo:
## What

`pkWhereClause` now reports when it fell back to the all-columns WHERE (PK-less table, unresolvable table, or PK column missing from the row image), and `buildDelete`/`buildUpdate` append ` LIMIT 1` to those statements in the MySQL dialect.

## Why

The all-columns fallback WHERE matches **every byte-identical duplicate row**, while the row event being reversed touched exactly one row. Reversing one INSERT therefore deleted ALL duplicate copies (silent loss of pre-existing rows), and a reverse UPDATE modified them all. `DELETE`/`UPDATE ... LIMIT 1` caps the reversal at one row; which physical copy is affected is undefined and irrelevant, since the matching rows are identical on every column the WHERE references.

- PK-scoped WHEREs (resolver present, PK resolves): unchanged, no LIMIT.
- PostgreSQL dialect: unchanged — PG has no `UPDATE`/`DELETE ... LIMIT`, so the fallback stays unbounded there and `docs/query-and-recovery.md` now documents that residual caveat explicitly (the old "assumes no duplicate rows" claim is replaced by the LIMIT 1 behavior for MySQL).

## Tests

New unit tests: fallback DELETE and UPDATE end with ` LIMIT 1`; PK-scoped DELETE/UPDATE carry no LIMIT; PostgreSQL-dialect fallback carries no LIMIT (invalid syntax there).

- `go build ./...` — ok
- `go vet ./internal/recovery/` — ok
- `go test ./internal/recovery/... -count=1` — ok
- `go test ./internal/cascaderecover/ ./internal/cli/ ./internal/console/ ./internal/query/ ./internal/baseline/ -count=1` — ok (dependent packages)
- `go test -tags integration ./internal/recovery/... -count=1` — ok (MySQL 13306)

Closes #789

### DIFF
--- a/docs/query-and-recovery.md
+++ b/docs/query-and-recovery.md
@@ -362,16 +362,17 @@ For `UPDATE` and `DELETE` reversals, the generator needs a `WHERE` clause to ide
 UPDATE `mydb`.`orders` SET `status` = 'draft' WHERE `id` = 42
 ```
 
-**Without a snapshot (fallback)**: Uses every column in the row image. This is verbose, and is **not** a guarantee of correctness even for tables without duplicate rows:
+**Without a snapshot (fallback)**: Uses every column in the row image, capped at one row:
 
 ```sql
 UPDATE `mydb`.`orders`
 SET `status` = 'draft'
-WHERE `id` = 42 AND `status` = 'published' AND `created_at` = '2026-02-19 14:01:00' AND `notes` IS NULL
+WHERE `id` = 42 AND `status` = 'published' AND `created_at` = '2026-02-19 14:01:00' AND `notes` IS NULL LIMIT 1
 ```
 
-- A `NULL` column renders as `IS NULL` (not `= NULL`, which never matches in SQL) — but a row whose *other* columns are all non-`NULL` and identical to another row (no natural key, no snapshot) will match **both** rows, and the reversal over-applies. There is no runtime detection of this: the generator has no way to know the table has (or doesn't have) duplicate rows.
-- "Without duplicate rows" is a precondition the generator assumes, not something it checks. If you rely on this fallback (no schema snapshot available), verify the table has a natural uniqueness property before applying the generated script.
+- A `NULL` column renders as `IS NULL` (not `= NULL`, which never matches in SQL).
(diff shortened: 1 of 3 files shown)


nethalo:
Applied the confirmed should-fix finding from review: the LIMIT 1 cap from #789 only applies to the MySQL dialect (PostgreSQL has no `UPDATE`/`DELETE ... LIMIT`). On the PostgreSQL-dialect all-columns fallback (PK-less table, no snapshot) the reversal stayed unbounded with zero runtime signal beyond a docs paragraph - reversing one INSERT against a PK-less table with duplicate rows would generate a DELETE matching every byte-identical copy.

Fix: `internal/recovery/recovery.go` now emits a leading `-- WARNING: unbounded all-columns WHERE, no per-statement row cap on this dialect ...` comment immediately above each affected statement when the PostgreSQL-dialect all-columns fallback is taken (`allColsFallbackWarningComment`, mirrors `rowLimitSuffix`'s dialect gate). A comment was chosen over a `slog.Warn`-only signal because it travels with the script into `--dry-run`/`--output` review, which is where an operator actually looks before applying. Also updated `docs/query-and-recovery.md` to note the inline warning, and added tests pinning the warning on both DELETE and UPDATE PG fallback paths, plus a negative test confirming a PK-resolved PG statement stays clean.

Verified:
- `go build ./...`
- `go vet ./...`
- `go test ./internal/recovery/... -count=1` (incl. new PG-warning tests)
- `go test -tags integration ./internal/recovery/... -count=1`
- `go test ./internal/cascaderecover/... ./internal/cli/... ./internal/console/... ./internal/agent/... ./internal/reconstruct/... ./cmd/bintrail-mcp/... -count=1` (all downstream consumers of `GenerateSQLFromRows`)

Commit: 8a72f70

PR #927, opened 2026-07-10
recover: --limit keeps the most recent events; truncated flag in JSON output

### LINKED ISSUE #785: recover: --limit truncates keeping the OLDEST events — the partial script maps to no consistent state
`recover` doesn't set `opts.Order` → `Fetch` orders ASC and `LIMIT` (default 1000) keeps the OLDEST N events of the window. Reverting only those leaves the newest events applied: the result is no historical consistent state (e.g. the reverse UPDATE of an old event uses `WHERE row_after` that a later un-reverted event already changed → 0 rows matched silently; or the reverse DELETE removes a row a later event modified).

Location: `internal/cli/recover.go:306`

Current mitigation: a truncation warning to stderr (`recover.go:306/346`) — but only if `n >= rLimit` (if a statement failed to generate, `n < limit` and there's no warning), and the `--format json` branch returns before the warning, so automated consumers never see it.

Fix: select the most recent SUFFIX when truncating (a partial rollback to a consistent intermediate point) or refuse, and carry `truncated` into the JSON payload.

---

nethalo:
## What

`recover` never set `query.Options.Order`, so the fetch ordered ASC and `--limit` (default 1000) kept the **oldest** N events of the window. Reversing only that prefix leaves the newer events applied — the partial script maps to no historical state (a reverse UPDATE's `row_after` WHERE silently matches 0 rows that a later un-reverted event rewrote; a reverse DELETE removes a row a later event modified).

- Set `Order: "DESC"` on the recover fetch so the SQL `LIMIT` keeps the most recent **suffix** of the window — reverting a newest-suffix is a rollback to a consistent intermediate point. The whole pipeline (`buildQuery`, `sortResults`, `MergeAndTrim`, `parquetquery`) already honors `Order=DESC`, so the change is confined to the recover command layer.
- Restore ascending order after the fetch (`query.MergeResults(rows, 0, "ASC")`) before generation — `GenerateSQLFromRows` expects ASC input and reverses internally to undo most-recent first.
- Make the truncation warning reliable: truncation is detected on the **fetched** row count before generation (fires even when generation later refuses) and emitted as a `slog.Warn` for every format; both `--format json` payloads (dry-run and file-output) now carry a `truncated` field. Previously the stderr-only warning was emitted after the JSON branch returned, so automated consumers never saw it.

## Why

Issue #785 (full-codebase audit finding): a truncated reversal script that undoes the oldest prefix is a silently inconsistent recovery — the worst failure mode for a recovery tool.

## Tests

New integration tests (`internal/cli/recover_limit_integration_test.go`) seed a 4-event UPDATE chain (v0→v1→v2→v3→v4) and pin both halves:

- `TestRecover_limitKeepsNewestEvents`: with `--limit 2` the script must reverse only the newest two events, most-recent first. **Fails on the old ASC ordering** (verified by temporarily flipping `Order` back: `script reversed the OLDEST events`).
- `TestRecover_jsonCarriesTruncatedFlag`: `--format json --dry-run` payload reports `truncated=true, statements=2` when capped and `truncated=false, statements=4` when not.

```
go build ./...                                        ok
go vet ./internal/cli/                                ok
go test ./internal/cli/... -count=1                   ok (1.5s)
go test -tags integration ./internal/cli/... -count=1 ok (19.6s, MySQL 13306)
```

## Notes

The MCP `recover` tool (`cmd/bintrail-mcp/main.go` recoverTool) has the same oldest-prefix truncation shape (no `Order` on its fetch, warning as a trailing SQL comment) — sibling surface, left out of this PR to keep it scoped to the cited finding.

Closes #785

### DIFF
--- a/internal/cli/recover.go
+++ b/internal/cli/recover.go
@@ -179,6 +179,15 @@ func runRecover(cmd *cobra.Command, args []string) error {
 		Flag:       rFlag,
 		Limit:      rLimit,
 		LimitPerPK: rLimitPerPK,
+		// When --limit truncates the window it must keep the most RECENT
+		// events (#785): reversing a newest-suffix rolls the data back to a
+		// consistent intermediate point, whereas the ASC default would keep
+		// the OLDEST prefix — undoing old events underneath later un-reverted
+		// ones maps to no state that ever existed (the reverse UPDATE's
+		// row_after WHERE no longer matches, or the reverse DELETE removes a
+		// row a later event rewrote). The rows are re-sorted ascending after
+		// the fetch, before generation.
+		Order: "DESC",
 	}
 
 	// ── Connect to index database ─────────────────────────────────────────────
@@ -300,6 +309,18 @@ func runRecover(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// The fetch above ran Order=DESC so --limit kept the newest suffix of the
+	// window (#785). Detect truncation on the FETCHED row count — before
+	// generation, so the warning fires even when generation later refuses —
+	// then restore ascending order: GenerateSQLFromRows expects ASC input and
+	// reverses it internally to undo most-recent first.
+	truncated := rLimit > 0 && len(rows) >= rLimit
+	if truncated {
+		slog.Warn("matched events truncated at --limit; only the most recent events of the window are reversed",
+			"limit", rLimit)
+	}
+	rows = query.MergeResults(rows, 0, "ASC")
+
 	// ── Generate recovery SQL ─────────────────────────────────────────────────
 	// Select the SQL dialect from the source flavor recorded in the index
 	// (single-source per stream_state) — PostgreSQL needs double-quoted identifiers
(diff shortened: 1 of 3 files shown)


nethalo:
Applied a confirmed review finding: the comment on `internal/console/api.go:251-255` ("Forcing ASC here also makes the LIMIT select the oldest N, matching the CLI") became factually wrong once this PR changed the CLI to fetch DESC + re-sort ASC so `--limit` truncation keeps the **newest** N events. The console recover handler still forces `Order = ""` (ASC), so it still keeps the **oldest** N when truncated — the same inconsistent-partial-state bug #785 fixed for the CLI, just not fixed here.

Changes in `e48acae`:
- Corrected the comment to state the console's actual (unfixed) behavior instead of the now-false "matches the CLI" claim, and pointed at the follow-up issues.
- No behavior change — the console/MCP recover paths are a different subsystem/release surface and are intentionally out of scope for this PR.

Filed as follow-ups (not fixed here, to avoid scope creep):
- dbtrail/dbtrail#981 — console recover truncation-order bug
- dbtrail/dbtrail#982 — MCP `recover` tool, same bug (`opts.Order` never set → `MergeResults` defaults ASC)

Verification: `go build ./...`, `go vet ./internal/console/...`, `go test ./internal/console/... -count=1` (ok, 11.6s), `go test -tags integration ./internal/console/... -count=1` (ok, 30.5s, MySQL 13306).

PR #928, opened 2026-07-10
shim: interpret MySQL escape sequences in string PK literals

### LINKED ISSUE #826: shim: string literals lack MySQL escape semantics — backslash values produce a false "the row didn't exist"
`stripQuotes` (`parser.go:433`) only removes the outer quotes; it doesn't process `\'`, `''`, `\n`, `\\`. A string PK containing a backslash — `WHERE path = 'C:\\temp\\new'` escaped MySQL-style — searches for the literal bytes `C:\\temp\\new` (with double backslash) against `pk_values`, doesn't match, and returns an empty resultset: in the documented AS OF semantics, an empty resultset means "the row didn't exist at that instant" — a FALSE forensic conclusion with no signal.

Location: `internal/shim/parser.go`

Values containing single quotes don't even parse (the regex `'[^']*'` cuts at the first quote → 1064, fail-loud, the benign case). Irrelevant for numeric PKs (the vast majority), but the shim explicitly supports CHAR/VARCHAR/TEXT PKs.

Fix: interpret standard MySQL escapes in `stripQuotes`, or document that literals are raw-bytes without escapes.

---

nethalo:
## What

`stripQuotes` in `internal/shim/parser.go` only removed the outer quotes of a captured WHERE value. A string PK containing a backslash — `WHERE path = 'C:\\temp\\new'` escaped MySQL-style — was matched against `pk_values` as the literal escaped bytes (`C:\\temp\\new`), missed, and returned an empty resultset. Under the documented AS OF semantics an empty resultset means "the row didn't exist at that instant" — a false forensic conclusion with no signal.

## Fix

Quoted literals now pass through `unescapeStringLiteral`, which applies standard MySQL string-literal semantics (MySQL 8.0 reference §9.1.1):

- `\0` `\b` `\n` `\r` `\t` `\Z` `\\` `\'` `\"` translate to their character
- `\%` and `\_` keep the backslash (only special in LIKE patterns)
- a backslash before any other character is dropped (escapes are case-sensitive: `\Z` is Ctrl+Z, `\z` is `z`)
- `''` collapses to `'`

Applied at the single `stripQuotes` seam, so all four call sites benefit: `_flashback`/`_snapshot`, `_diff`, the `DBTRAIL_AT` hint form, and the bare `AS OF` form. Numeric (unquoted) values pass through unchanged.

Unchanged by design: literals containing a quote character (`'O\'Brien'`, `'O''Brien'`) still cannot be captured by the value regexes (`'[^']*'`) — they miss the full matcher and keep the fail-loud ER_PARSE_ERROR (1064) path, pinned by `TestParseQuoteInStringPKStaysFailLoud`. The `\'`/`''` branches in the unescaper are for MySQL-semantics completeness (unit-tested directly).

## Tests

- `TestParseStringPKEscapes` — escape matrix through `Parse` (backslash path, `\n`/`\t`, `\r`/`\0`/`\b`/`\Z`, `\"`, LIKE escapes, case-sensitive `\z`, plain value)
- `TestParseStringPKEscapesAllForms` — `_diff`, hint form, bare AS OF form all unescape
- `TestUnescapeStringLiteral` — `''`, `\'`, trailing lone backslash, empty
- `TestParseQuoteInStringPKStaysFailLoud` — 1064 behavior preserved

Verified fail-before: with the unescape call reverted, the new tests fail with `PKValue = "C:\\\\temp", want "C:\\temp"`.

```
go build ./...                                        ok
go vet ./internal/shim/                               ok
go test ./internal/shim/ -count=1                     ok (1.5s)
go test -tags integration ./internal/shim/... -count=1  ok (12.2s, MySQL 13306)
```

Closes #826

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -133,12 +133,22 @@ func (f *Filters) Matches(schema, table string) bool {
 func BuildPKValues(pkColumns []metadata.ColumnMeta, row map[string]any) string {
 	parts := make([]string, 0, len(pkColumns))
 	for _, col := range pkColumns {
-		val := formatPKValue(row[col.Name])
-		val = strings.ReplaceAll(val, `\`, `\\`)
-		val = strings.ReplaceAll(val, `|`, `\|`)
-		parts = append(parts, val)
+		parts = append(parts, EscapePKValue(formatPKValue(row[col.Name])))
 	}
 	return strings.Join(parts, "|")
+}
+
+// EscapePKValue applies the pipe/backslash escaping BuildPKValues uses for
+// each individual PK column value. Exported so callers holding a raw,
+// already-decoded single-column PK value (e.g. internal/shim, after
+// MySQL-unescaping a SQL string literal) can re-encode it into the same
+// at-rest form stored in binlog_events.pk_values before using it as a match
+// filter — the string-literal unescaping and this pipe-delimiter escaping
+// are two unrelated encodings, and neither implies the other.
+func EscapePKValue(val string) string {
+	val = strings.ReplaceAll(val, `\`, `\\`)
+	val = strings.ReplaceAll(val, `|`, `\|`)
+	return val
 }
 
 // formatPKValue renders a single PK value for BuildPKValues. []byte needs its
--- a/internal/shim/handler.go
+++ b/internal/shim/handler.go
@@ -452,12 +452,18 @@ func (h *Handler) runPointInTime(q TimeTravelQuery) (*mysql.Result, error) {
 	// ROW_NUMBER OVER (PARTITION BY pk_values ORDER BY event_timestamp
 	// DESC, event_id DESC) so the kept event for each PK is the most
 	// recent one — exactly what _flashback "row state at AsOf" needs.
+	// q.PKValue is the raw, MySQL-unescaped literal value (stripQuotes /
+	// unescapeStringLiteral, #826). Options.PKValues is matched against
+	// binlog_events.pk_values, which is stored in event.BuildPKValues-encoded
+	// form (backslash/pipe escaped). Re-encode with event.EscapePKValue here
+	// so a backslash- or pipe-containing PK still matches instead of
+	// silently missing (#826 follow-up).
 	engine := query.New(h.indexDB)
 	rows, _, err := query.FetchMerged(ctx, h.indexDB, engine, query.FetchMergedOptions{
 		Opts: query.Options{
 			Schema:     q.Schema,
 			Table:      q.Table,
-			PKValues:   q.PKValue,
+			PKValues:   event.EscapePKValue(q.PKValue),
 			Until:      &q.AsOf,
 			LimitPerPK: 1,
 		},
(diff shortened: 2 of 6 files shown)


nethalo:
Two independent adversarial reviews of this PR (each with an empirical repro) found the same critical regression, so I pushed a follow-up fix on this branch (`2926c22`).

**The bug.** `stripQuotes`/`unescapeStringLiteral` (this PR) correctly MySQL-unescapes a quoted PK literal — e.g. `'C:\\temp\\new'` → `C:\temp\new`. But that raw value then flowed unchanged into `query.Options.PKValues`/`PKValuesIn` at every match seam:
- `internal/shim/handler.go` `_flashback` (`runPointInTime`)
- `internal/shim/handler.go` `_diff` (`runDiff`)
- `internal/shim/snapshot.go` `_snapshot`'s post-baseline delta fetch

Those seams compare against `binlog_events.pk_values`, which is stored in `event.BuildPKValues`-encoded form (backslash doubled, pipe escaped). Raw `C:\temp\new` no longer matches encoded `C:\\temp\\new` → 0 rows, where the parent (pre-#826-fix) commit coincidentally returned 1 row for the backslash case. That's a straight regression of the exact "false empty resultset = row didn't exist" scenario #826 was filed to eliminate — and it's worse on `_snapshot`: `ReadBaselineRow` correctly gets the raw value (Parquet stores actual column values, not encoded pk_values) so the baseline lookup still matches, but the same raw value used for the post-baseline delta fetch missed — silently serving a **stale baseline image** instead of applying a post-baseline UPDATE/DELETE.

Both reviews also flagged the pre-existing sibling gap: pipe-containing PKs (`a|b`, stored as `a\|b`) miss the same way.

**The fix.** Extracted `event.EscapePKValue` (the per-value half of `BuildPKValues`'s escaping) and applied it at each of the three match seams, right before the value becomes `Options.PKValues`/`PKValuesIn`. `ReadBaselineRow` (snapshot.go) intentionally keeps receiving the raw, unescaped value — Parquet baseline rows store actual column values, not the encoded form.

**Why the original tests went green over this.** PR #928's tests (`TestParseStringPKEscapes`) only assert `q.PKValue` at the parser layer — they never cross into a real `pk_values`-populated table. Added `internal/shim/pk_escape_seam_integration_test.go`, which seeds a REAL `event.BuildPKValues`-encoded row, drives it through `Parse()` + the actual handler methods, and independently checks `_flashback`, `_diff`, and the `_snapshot` split-brain case as separate subtests (so one seam's regression can't hide behind another's `Fatalf`).

**Evidence — ran the new tests before and after, on both a backslash PK and a pipe PK:**

Pre-fix (handler.go/snapshot.go reverted, only the parser-layer fix + `event.EscapePKValue` helper present):
```
--- FAIL: TestPKValueEscapeSeam_BackslashAndPipe/backslash_path/_flashback
    _flashback: expected 1 row for PK "C:\\temp\\new" (stored pk_values="C:\\\\temp\\\\new"), got 0 — encoding seam regression
--- FAIL: TestPKValueEscapeSeam_BackslashAndPipe/backslash_path/_diff
    _diff: expected 1 row for PK "C:\\temp\\new", got 0 — encoding seam regression
--- FAIL: TestPKValueEscapeSeam_BackslashAndPipe/pipe-containing_PK/_flashback
    _flashback: expected 1 row for PK "a|b" (stored pk_values="a\\|b"), got 0 — encoding seam regression
--- FAIL: TestPKValueEscapeSeam_BackslashAndPipe/pipe-containing_PK/_diff
    _diff: expected 1 row for PK "a|b", got 0 — encoding seam regression
--- FAIL: TestSnapshotBaseline_BackslashPK_DeltaAppliesAfterBaseline
    _snapshot row = [C:\temp\new alice], want [C:\temp\new bob] — post-baseline UPDATE not applied (stale baseline served)
```

Post-fix: all pass, `_flashback`/`_diff`/`_snapshot` subtests green for both the backslash and pipe cases.

Full suite: `go build ./...`, `go vet ./internal/shim/... ./internal/event/...`, `go test ./internal/shim/... ./internal/event/... -count=1`, and `go test -tags integration ./internal/shim/... -count=1` all pass.

Both must-fix findings (backslash-PK false-empty on `_flashback`/`_diff`, and the `_snapshot` stale-baseline split-brain) and the should-fix pipe case are closed.

PR #929, opened 2026-07-10
byos: close EnsurePartitionKey check-then-write race with S3 conditional put

### LINKED ISSUE #806: archive (BYOS): EnsurePartitionKey check-then-write race can re-arm the #198 cutover the marker exists to prevent
`Exists→Put` is not atomic: two agents starting simultaneously against the same empty prefix (duplicate deploy, or migration with a different serverID — exactly the #198 UUID-vs-numeric-id scenario) both see `!ok` and write their own marker; last-write-wins with no error. The agent whose marker was overwritten keeps writing payloads under a partition key the surviving marker doesn't know → those objects fall outside the new-layout queries, silently.

Location: `internal/byos/partition_key.go:45`

S3 supports conditional PutObject (`If-None-Match: '*'`) since 2024 and the SDK exposes it; with that the second Put fails with `PreconditionFailed` and the agent re-reads and compares — closing the race in one line.

Fix: conditional PutObject on the marker write.

---

nethalo:
The Exists→Put sequence in `EnsurePartitionKey` was not atomic: two agents starting simultaneously against the same empty prefix (duplicate deploy, or a migration with a different serverID) could both observe the marker as absent and each write their own, last-write-wins. The losing agent kept writing payloads under a partition key the surviving marker doesn't know, silently re-arming the #198 cutover the marker exists to prevent.

Adds an optional `storage.ConditionalPutter` seam (`PutIfAbsent` + sentinel `ErrObjectExists`). `S3Backend` implements it with a conditional PutObject (`If-None-Match: *`): the losing writer gets `PreconditionFailed`, and `EnsurePartitionKey` then falls through to the existing read-and-compare path instead of overwriting. Backends without the interface keep the prior non-atomic behavior; S3-compatible services that answer `NotImplemented` to conditional writes degrade to a plain Put.

Unit tests cover the winning/losing writer paths and the NotImplemented degradation.

Closes #806

### DIFF
--- a/internal/byos/partition_key.go
+++ b/internal/byos/partition_key.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"time"
@@ -56,10 +57,17 @@ func EnsurePartitionKey(ctx context.Context, b storage.Backend, serverID string)
 		if err != nil {
 			return fmt.Errorf("marshal partition-key marker: %w", err)
 		}
-		if err := b.Put(ctx, partitionKeyMarker, bytes.NewReader(body)); err != nil {
+		created, err := putMarker(ctx, b, body)
+		if err != nil {
 			return fmt.Errorf("write partition-key marker: %w", err)
 		}
-		return nil
+		if created {
+			return nil
+		}
+		// Lost the create race: another agent wrote the marker between the
+		// Exists check and our conditional write. Fall through to
+		// read-and-compare so the surviving marker is validated instead of
+		// silently overwritten (which would re-arm the #198 cutover).
 	}
 
 	rc, err := b.Get(ctx, partitionKeyMarker)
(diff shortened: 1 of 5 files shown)


nethalo:
Applied a confirmed should_fix finding from adversarial review on `internal/storage/s3.go`'s `PutIfAbsent`:

1. **409 ConditionalRequestConflict was unmapped.** AWS documents this response (alongside 412 PreconditionFailed) for a conditional write that races another write to the same key during upload -- exactly the two-agents race this PR closes. Unmapped, it fell through to the generic error branch, so the losing writer got a hard `PutIfAbsent` error instead of `ErrObjectExists`, and `EnsurePartitionKey` propagated that straight up to a startup abort (`cliapp/agent.go:352`) instead of taking the intended read-and-compare fallback. Now both the API error code and the generic-HTTP-409 form map to `ErrObjectExists`, same as 412.
2. **Silent NotImplemented/501 degrade.** Falling back to an unconditional `Put` on backends without conditional-write support had zero operator signal -- the atomicity guarantee this PR adds goes inactive with nothing in the logs. Added an `slog.Warn` naming the degraded guarantee.

Added `TestPutIfAbsentConditionalRequestConflictRefused`, `TestPutIfAbsentHTTP409Refused` (API-code and generic-HTTP forms, mirroring the existing 412 tests), and extended `TestPutIfAbsentNotImplementedFallsBack` to assert the warning fires.

`go build ./...`, `go vet ./internal/storage/... ./internal/byos/...`, `go test ./internal/storage/... ./internal/byos/... -count=1`, and `go test -tags integration ./internal/byos/... -count=1` (MariaDB 13307) all pass. `internal/storage` has no integration-tagged tests.

Issue #766, opened 2026-07-05
reconstruct: ErrNoBaseline emits an EMPTY table and doesn't even replay binlog-only; prefers local dir over live S3 copy


nethalo:
With `ErrNoBaseline` the worker does `slog.Warn("treating table as empty at dump time")` and returns an empty report **without fetching events**: a table created after the last snapshot with 100k inserted rows (all in the binlog) comes out of `reconstruct --output-format mydumper` with no file at all, exit 0. The shim in the same case degrades to binlog-only (`snapshot.go:117-120`) and DOES recover those rows — full-table is strictly worse.

Location: `internal/reconstruct/fulltable.go:315`

Compound scenario with prune (#616): prune deletes old local snapshots (durable in S3), `console/manager.go:226` prefers `BaselineDir` over `BaselineS3`, so a reconstruct/time-travel to a date X older than local retention hits `ErrNoBaseline` even though the S3 copy exists → restore that drops the whole table with only a log Warn.

Fix: apply binlog-only events in this path (parity with the shim) and/or require an explicit `--allow-missing-baseline` flag; ideally fall back to `BaselineS3` when the local dir has no candidate.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

Issue #602, opened 2026-06-23
reconstruct: --output-format mydumper projects onto baseline schema, silently dropping a column added after the baseline


nethalo:
## Summary

Full-table `bintrail reconstruct --output-format mydumper` derives its output column set from the **baseline** Parquet (its physical columns + embedded `CREATE TABLE`). A column **added** after the baseline is not in that set, so its values — which *are* present in the delta events applied on top — are **never emitted**. Silent column-data loss in the full-table restore output.

## Mechanism

- `internal/reconstruct/fulltable.go` — `mergeBaselineIntoWriter` reads the column list via `readBaselineColumns` (`SELECT * ... LIMIT 0` on the baseline Parquet) and writes the baseline's `CreateTableSQL` as the schema header.
- `rowAfterOrdered` (~fulltable.go:804) iterates the **baseline** `colNames`; a delta event's `row_after` key not in `colNames` is never looked up → a post-baseline-added column is dropped for every row.

Counterpart to #600: there the fixed schema is the *latest* snapshot; here it's the *baseline era*. Same family — output columns come from a fixed schema that differs from the as-of-T data.

## Confidence

Code-read (per architecture trace); **repro pending**. Repro before fixing — repro > cita.

## Design tension (decide on triage)

`reconstruct` full-table is deliberately **baseline-anchored** — it emits the baseline's `CREATE TABLE` as the mydumper schema header. Surfacing a post-baseline-added column means reconciling the output schema/DDL with the target-time columns, a larger change than the shim union fix in #600 (it touches the mydumper schema header, not just row assembly). Scope on triage:

- **minimum (fail loud):** warn / error when delta events carry columns absent from the baseline schema — don't drop silently;
- **fuller:** union the post-baseline columns into the output and reconcile the emitted `CREATE TABLE`.

## Relationship

Schema-drift family with #600 (shim display) and the `recover` apply-time sibling. Distinct path, distinct fix.

Issue #782, opened 2026-07-05
reconstruct: PK-changing UPDATE can duplicate the row or resurrect a deleted one; single-row reports a misleading "no row found"


nethalo:
`PKValues` for an UPDATE is built from the BEFORE image (`parser.go:537`) and the change-map collapses by that key.

Location: `internal/reconstruct/fulltable.go`

Scenario A (resurrection): `UPDATE pk 1→2` + `DELETE pk=2`: `changes={1:update(row_after pk=2), 2:delete}`; the baseline row pk=1 matches the update and EMITS the pk=2 row that was actually deleted → silent resurrection.
Scenario B (duplicate): `UPDATE 1→2` + `UPDATE pk=2`: emits the stale after of the first update AND the leftover of the second → pk=2 duplicated in the dump (blows up with 1062 only at restore).
Single-row searching pk=2 doesn't fetch the event keyed '1' and fails with a misleading "no row found in baseline".

Only warned in the Go doc-comment of `FullTableConfig` ("Re-snapshot the baseline"), not in `docs/`. Buildable: `row_before` and `row_after` are both stored, so old-PK→new-PK transitions can be chained in the merge.

Fix: chain PK-changing UPDATE transitions in the merge using the stored before/after images.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #931, opened 2026-07-10
reconstruct: refuse full-table merge on post-baseline dropped column instead of mixed-epoch NULL-fill

### LINKED ISSUE #843: reconstruct: dropped column post-baseline emits a mixed stale/NULL epoch state with a per-row warn (remaining #600/#601/#602 class)
`rowAfterOrdered` NULL-fills baseline columns absent from `row_after` and warns per row: after a DROP COLUMN, rows touched post-drop come out with NULL and never-touched rows keep the old value → the dump mixes two schema epochs in one table (a state that never existed), with the baseline's CREATE TABLE still declaring the column.

Location: `internal/reconstruct/fulltable.go:1087`

Deliberately-unresolved asymmetry: added column = fail-loud refusal (#602, line 527), dropped column = warn-and-emit. Also the warn is per row×column: 1M rows updated post-drop = 1M log lines. Scenario: drop a PII column Monday, `reconstruct --at Tuesday` → the "table as of Tuesday" re-exposes the PII column with stale values in the cold rows.

Fix: match the #602 posture (refuse with a re-baseline message) or project out the dropped column by consulting the schema epoch at `--at`.

---

nethalo:
## What

`rowAfterOrdered` (`internal/reconstruct/fulltable.go`) NULL-filled baseline columns absent from a delta event's `row_after` image, warning once per row x column. After a `DROP COLUMN` post-baseline, a full-table reconstruct emitted a dump mixing two schema epochs: rows touched after the drop carried NULL while never-touched baseline pass-through rows kept the pre-drop value — under a `CREATE TABLE` header that still declared the column. That state never existed on the source, and it silently re-exposes a dropped (e.g. PII) column's stale values. 1M post-drop rows also meant 1M warn lines.

This was the deliberately-unresolved asymmetry of #602: added column = fail-loud refusal, dropped column = warn-and-emit.

## Fix

Match the #602 posture. New `droppedBaselineColumns` helper (symmetric counterpart of `postBaselineColumns`) scans the change map once, up front, and `mergeBaselineIntoWriter` refuses with an aggregated re-baseline error before the writer opens — no per-row spam, no partial chunk files on disk.

Column ABSENCE from the image is the signal: `binlog_row_image=FULL` (validated at index time) guarantees a complete after-image, so a missing key means the column no longer existed at event time. A genuinely-NULL value is present in the map with a nil value and merges exactly as before (pinned by test).

The NULL-fill-with-warn in `rowAfterOrdered` remains only for the binlog-only fallback path (`writeBinlogOnlyChanges`, #766), whose `colNames` come from a resolver snapshot rather than a baseline — re-baselining is not an answer there.

## Tests

- `TestReconstruct843_columnDroppedAfterBaselineFailsLoud` — refusal names the column, points at `bintrail baseline`, leaves no partial `.sql` output. Verified fail-before: with the guard neutered, the test fails (`got nil` + the old per-row WARN fires).
- `TestReconstruct843_nullValueIsNotADrop` — `row_after` carrying `status: nil` still merges and emits `(2, NULL)`.
- `TestDroppedBaselineColumns` — pure-helper matrix (no drift, NULL-not-a-drop, sorted/dedup across events, DELETE/nil-RowAfter/nil-entry ignored).

```
go build ./...                                          ok
go vet ./internal/reconstruct/                          ok
go test ./internal/reconstruct/... -count=1             ok (4.2s)
go test ./internal/cli/... ./internal/verify/... -count=1   ok
go test -tags integration ./internal/reconstruct/... -count=1   ok (4.9s, MySQL 13306)
go test -tags integration ./internal/shim/... -count=1          ok (11.0s)
```

Out of scope (same class, different surface): the shim's in-memory `_snapshot` full-table path (`SnapshotFullTableImages`) has no drift guard in either direction — #602 chose the mydumper writer path only, and this PR keeps that boundary.

Closes #843

### DIFF
--- a/internal/reconstruct/fulltable.go
+++ b/internal/reconstruct/fulltable.go
@@ -1424,25 +1424,45 @@ func postBaselineColumns(changes map[string]*query.ResultRow, colNames []string)
 
 // droppedBaselineColumns is the symmetric counterpart of postBaselineColumns:
 // it returns, sorted and de-duplicated, the baseline column names ABSENT (key
-// missing — not value NULL) from some non-DELETE event's row_after image,
-// i.e. columns DROPPED from the source table after the baseline snapshot.
-// binlog_row_image=FULL (validated at index time) guarantees a complete
-// after-image, so key absence means the column no longer existed at event
-// time; a genuinely-NULL value is present in the map as a nil value and is
-// never flagged. mergeBaselineIntoWriter calls this up front to refuse the
-// run instead of letting rowAfterOrdered NULL-fill the column row by row
-// (#843). DELETE events carry no row_after and are skipped.
+// missing — not value NULL) from some event's row image, i.e. columns DROPPED
+// from the source table after the baseline snapshot. binlog_row_image=FULL
+// (validated at index time) guarantees a complete image, so key absence means
+// the column no longer existed at event time; a genuinely-NULL value is
+// present in the map as a nil value and is never flagged. mergeBaselineIntoWriter
+// calls this up front to refuse the run instead of letting rowAfterOrdered
+// NULL-fill the column row by row (#843).
+//
+// Both row_after AND row_before are scanned (#843 follow-up): a window whose
+// last event for a post-drop-touched PK is a DELETE carries no row_after, but
+// its row_before is itself a post-drop image (the row as it existed just
+// before deletion) and so still lacks the dropped column — checking only
+// row_after let such a window sail through with zero warning, silently
+// re-emitting the stale pre-drop value on every untouched pass-through row.
+// Per-PK this collapsed-map scan is sufficient: changes is keyed by PK with
+// the chronologically LAST event surviving, and time only moves forward, so
+// if any event for a PK is post-drop then so is that surviving one — there is
+// no ordering in which an earlier post-drop event's signal is lost behind a
+// later pre-drop entry for the same key. A PK with zero post-drop activity in
+// the window remains undetectable by construction (no image ever samples the
+// post-drop schema for it) and is out of scope here.
 func droppedBaselineColumns(changes map[string]*query.ResultRow, colNames []string) []string {
 	dropped := make(map[string]struct{})
-	for _, ev := range changes {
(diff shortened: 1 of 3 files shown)


nethalo:
Applied a should_fix finding from the adversarial review: droppedBaselineColumns only inspected row_after, so it skipped DELETE events entirely (ev.EventType == event.EventDelete -> continue, ev.RowAfter == nil -> continue). A window where a post-drop-touched PK's LAST event is a DELETE bypassed the guard completely - the DELETE's own row is correctly excluded from output (DeletesSkipped), but the missed detection let mergeBaselineIntoWriter proceed and silently re-emit the stale pre-drop value on every OTHER, never-touched pass-through row. Reproduced empirically before the fix: baseline columns {id,status}, one DELETE event with RowBefore:{id:2} and no status key, mergeBaselineIntoWriter returned nil error.

Fix: also scan row_before for both UPDATE and DELETE events. Under binlog_row_image=FULL a DELETE's row_before is itself a post-drop image (the row as it existed just before deletion), so it carries the same detection signal as row_after.

Scoped down one item from the finding: I did not add a raw in.Events-slice scan (the sibling #782 PK-change guard's pattern). Verified the changes map is built by folding the ascending-sorted event slice with last-write-wins (fulltable.go "Build the change map" step), so per-PK the collapsed map is already sufficient - if any event for a PK is post-drop, the chronologically-last surviving one is too, so there's no ordering where an earlier post-drop signal hides behind a later pre-drop entry for the same key. Confirmed via a stash-based before/after: the new regression test fails on the pre-fix code and passes with the fix.

Also updated the existing TestDroppedBaselineColumns case that pinned "DELETE events ignored (no row_after)" as wanting nil - it was pinning the blind spot as intended behavior - and added TestReconstruct843_columnDroppedDeleteOnlyFailsLoud, an end-to-end regression through mergeBaselineIntoWriter with a DELETE-only post-drop event.

Tests: go build ./..., go vet ./internal/reconstruct/..., go test ./internal/reconstruct/... -count=1, go test -tags integration ./internal/reconstruct/... -count=1 all green.

Issue #415, opened 2026-06-07
parser integration tests over-count events under parallel package runs: cross-package DDL leaks into the shared binlog window


nethalo:
## Summary

`TestParseFile_realBinlog` and `TestParseFiles_multiple` (`internal/parser/parser_integration_test.go`) over-count events and fail when the integration suite runs with full package parallelism (`go test -tags integration ./...`). They pass in isolation. This is a pre-existing test-isolation weakness — reproduced on clean `main` — surfaced while working on #414 / PR #413.

## Mechanism

1. These tests bracket a binlog window with `FLUSH BINARY LOGS`, do their own DML, seal, `docker cp` the file, and parse it expecting an exact event count.
2. Go runs test *packages* in parallel, all against the single shared `bintrail-test-mysql` container — one global binlog stream.
3. Concurrent packages' setup DDL (`CREATE TABLE index_state`, `stream_state`, `archive_state`, `table_flags`, `profiles`, ... via `testutil.InitIndexTables`) lands inside the window.
4. **Row events are schema-filtered, but DDL events bypass the schema filter by design** (`parseDDL` emits unconditionally for audit/auto-snapshot purposes) — so foreign `EventDDL`s inflate the expected-total assertions: `expected 4 events, got 7`, `expected 2 events, got 6`.

## Repro (on clean main)

```
go test -tags integration ./internal/parser/ ./internal/indexer/ ./internal/query/ ./internal/console/ -count=1
```
→ `TestParseFiles_multiple` fails with leaked `type=4` (EventDDL) events for foreign internal tables.

## Proposed fix

Count schema-filtered DML events only (INSERT/UPDATE/DELETE), as `TestParseFile_compressedTransactions` already does since PR #413 — the assertions keep full strength because row events ARE schema-filtered and each test owns its schema. Alternatives considered: serializing integration packages (`-p 1`, slower for everyone) or schema-filtering DDL emission (a behavior change just for tests — no).

PR #932, opened 2026-07-10
metadata: coerce SET bitmask to uint64; take snapshot_id allocation under FOR UPDATE

### LINKED ISSUE #844: capture: snapshot_id assigned MAX+1 without FOR UPDATE — concurrent writers merge rows under one id and the resolver skips events
`TakeSnapshot` and `WritePGSnapshot` assign `snapshot_id` with `SELECT COALESCE(MAX(snapshot_id),0)+1` inside the transaction but WITHOUT `FOR UPDATE`; the comment says the collision is "acceptable" because CLI runs are serial — but that's no longer true: under the `watch` daemon, the synchronous DDL hook (auto-snapshot), a manual `bintrail snapshot`, and the console baseline trigger can run concurrently against the same index.

Location: `internal/metadata/metadata.go:729`

Scenario: the DDL hook and a manual snapshot collide → two full sets of rows under one `snapshot_id` → `NewResolver` loads both copies and each table ends with double the columns → "column count mismatch — skipping" for ALL events of those tables until the next snapshot, with the checkpoint advancing (same silent-loss class as the DDL-hook finding).

Fix: `SELECT ... FOR UPDATE` on the assignment (both writes are already transactional), or a counter table with AUTO_INCREMENT.

---

### LINKED ISSUE #846: capture: SET with 64 members and member 64 active is indexed as a negative int64 (same class as the BIT #497 fix)
go-mysql decodes SET with `littleDecodeBit` → int64; with 64 members and the high bit set the bitmask goes negative. `coerceUnsigned` reinterprets BIT (`metadata.go:453`) but SET enters neither the bit special-case nor the 'unsigned' gate (a SET's ColumnType doesn't contain 'unsigned') → the index stores e.g. -9223372036854775808 instead of 9223372036854775808.

Location: `internal/metadata/metadata.go:447`

Scenario: recover emits that negative literal in an INSERT/UPDATE of the SET → MySQL rejects it (out of range) — at least fail-loud on apply — and reconstruct/shim's ordinal→labels mapping can't resolve a negative bitmask. Rare (SETs of exactly 64 members) but the same class already fixed for BIT(64) in #497.

Fix: reinterpret DataType 'set' to uint64 like 'bit'.

---

nethalo:
## What

Two capture-layer fixes in `internal/metadata`:

**SET decoded negative (#846).** go-mysql decodes SET as int64 (`littleDecodeBit`); with exactly 64 members and member 64 active the bitmask has the high bit set and indexes as e.g. `-9223372036854775808`. `coerceUnsigned` already reinterprets BIT (#497) but SET entered neither the BIT special-case nor the `unsigned` gate (a SET's ColumnType never contains "unsigned"). `recover` then emits the negative literal (MySQL rejects it as out of range) and ordinal-to-label mapping cannot resolve a negative bitmask. Fix: reinterpret DataType `set` to uint64 exactly like `bit` — identity for smaller bitmasks, NULL passthrough unchanged.

**snapshot_id merge under concurrency (#844).** `TakeSnapshot` and `WritePGSnapshot` allocated `snapshot_id` with `SELECT COALESCE(MAX(snapshot_id),0)+1` inside the transaction but WITHOUT `FOR UPDATE`; the old comment called the collision "acceptable" because CLI runs are serial — no longer true: under the `watch` daemon the synchronous DDL hook auto-snapshot, a manual `bintrail snapshot`, and the console baseline trigger can run concurrently against one index. Two writers reading the same MAX merge both row sets under one `snapshot_id`; `NewResolver` then loads doubled columns for every table and skips ALL their events ("column count mismatch") while the checkpoint advances — silent loss until the next snapshot. Fix: `FOR UPDATE` on the allocation read at both sites (both writes were already transactional); the second allocator blocks until the first commits, then reads the committed rows and takes the next id.

## Tests

- `TestCoerceUnsigned`: 4 new SET cases (member-64 active, all-64 active, small bitmask identity, NULL passthrough). Fail-before verified: with the `set` branch reverted, `= -9223372036854775808 (int64); want 0x8000000000000000 (uint64)`.
- `TestWritePGSnapshot_concurrentAllocatorSerialized` / `TestTakeSnapshot_concurrentAllocatorSerialized` (integration): hold an uncommitted simulated snapshot writer, assert the second allocator blocks behind it (300ms window), then commit and assert a distinct id with zero rows merged into the first writer's snapshot. Fail-before verified: with `FOR UPDATE` reverted both fail with "completed while a concurrent snapshot writer was uncommitted (gotID=1, otherID=1)".

```
go build ./...                                               ok
go vet ./internal/metadata/ (+ -tags integration)            ok
go test ./internal/metadata/... -count=1                     ok (0.24s)
go test -tags integration ./internal/metadata/... -count=1   ok (6.8s, MySQL 13306)
```

Closes #846
Closes #844

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -16,6 +16,7 @@ import (
 	mysql "github.com/go-sql-driver/mysql"
 
 	"github.com/dbtrail/dbtrail/internal/event"
+	"github.com/dbtrail/dbtrail/internal/metadata"
 )
 
 // Indexer consumes event.Events from a channel and batch-inserts them into
@@ -579,6 +580,25 @@ func EnsureSchema(db *sql.DB) error {
 	if !hasConnCache {
 		if _, err := db.Exec(ddlConnectionCache); err != nil {
 			return fmt.Errorf("failed to create connection_cache: %w", err)
+		}
+	}
+	// snapshot_id_seq post-dates the original schema (#844): a dedicated
+	// AUTO_INCREMENT counter table that lets metadata.TakeSnapshot/
+	// WritePGSnapshot allocate snapshot_id values without the deadlock-prone
+	// `MAX(snapshot_id)+1 FOR UPDATE` pattern it replaces (metadata.
+	// DDLSnapshotIDSeq has the full rationale). metadata.go's own callers
+	// also self-heal this table lazily on first use — the standalone
+	// `bintrail snapshot` command does not go through EnsureSchema — but
+	// creating it eagerly here means the daemon-driven writers named in
+	// #844 (the DDL hook, the console baseline trigger) never hit that lazy
+	// path under load.
+	hasSnapshotIDSeq, err := tableExists(db, "snapshot_id_seq")
+	if err != nil {
+		return err
+	}
+	if !hasSnapshotIDSeq {
+		if _, err := db.Exec(metadata.DDLSnapshotIDSeq); err != nil {
+			return fmt.Errorf("failed to create snapshot_id_seq: %w", err)
 		}
 	}
 	return nil
--- a/internal/indexer/schema.go
+++ b/internal/indexer/schema.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dbtrail/dbtrail/internal/metadata"
 	"github.com/dbtrail/dbtrail/internal/serverid"
 )
 
@@ -58,6 +59,7 @@ func CreateIndexTables(ctx context.Context, db *sql.DB, partitions int, encrypt 
 		{"schema_changes", ddlSchemaChanges},
 		{"fk_constraints", ddlFKConstraints},
 		{"connection_cache", ddlConnectionCache},
+		{"snapshot_id_seq", metadata.DDLSnapshotIDSeq},
 	}
 	for _, t := range ddls {
 		if _, err := db.Exec(t.ddl); err != nil {
--- a/internal/metadata/metadata.go
+++ b/internal/metadata/metadata.go
@@ -671,6 +671,10 @@ func WritePGSnapshot(db *sql.DB, rel *PGRelationSchema) (int, error) {
 		return 0, fmt.Errorf("metadata: WritePGSnapshot requires a relation with at least one column")
 	}
 
+	if err := ensureSnapshotIDSeqTable(db); err != nil {
+		return 0, err
+	}
+
 	tx, err := db.Begin()
 	if err != nil {
 		return 0, fmt.Errorf("metadata: WritePGSnapshot begin: %w", err)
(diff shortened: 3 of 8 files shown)


nethalo:
Redesigned the #844 fix — `FOR UPDATE` on `SELECT COALESCE(MAX(snapshot_id),0)+1 FROM schema_snapshots` was reverted.

**Why**: that lock correctly serializes concurrent snapshot writers in principle, but reliably deadlocks (MySQL `Error 1213`) under 3+ concurrent allocate-then-insert transactions — exactly the concurrency #844 names (the watch daemon's DDL-hook auto-snapshot, a manual `bintrail snapshot`, and the console baseline trigger racing on the same index). Neither `WritePGSnapshot`'s nor `TakeSnapshot`'s caller retries on a transient deadlock, so the fix as merged would crash the ingestion daemon under exactly the load it was meant to handle — a more severe failure than the silent-merge bug it fixed. Reproduced empirically against real MySQL before touching any code.

**Fix**: a dedicated `snapshot_id_seq` AUTO_INCREMENT counter table (`internal/metadata.DDLSnapshotIDSeq`). InnoDB allocates `AUTO_INCREMENT` values under a lightweight, statement-duration lock rather than a row/gap lock held for the whole transaction, so concurrent allocators serialize without ever deadlocking. `allocateSnapshotID` inserts a row into it and reads the id back from the `INSERT`'s own OK packet (`res.LastInsertId()`), inside the same transaction `WritePGSnapshot`/`TakeSnapshot` already use for the snapshot rows.

Schema wiring follows the existing `connection_cache` precedent in this codebase: provisioned in `indexer.CreateIndexTables` (fresh installs) and `indexer.EnsureSchema` (existing installs, at daemon startup), plus a lazy self-heal directly in `metadata.go` since the standalone `bintrail snapshot` CLI command calls `TakeSnapshot` without going through `EnsureSchema`. Also added to `migrations/001_create_tables.sql` and `internal/testutil.InitIndexTables` for parity with the other post-hoc tables.

`TestWritePGSnapshot_concurrentAllocatorSerialized` / `TestTakeSnapshot_concurrentAllocatorSerialized` were rewritten: the old assertion (a second writer blocks behind an uncommitted `FOR UPDATE` holder) is the wrong invariant for this design — `AUTO_INCREMENT` allocation isn't meant to block. The rewritten tests run 8 truly concurrent writers and assert zero errors (no deadlock) and zero `snapshot_id` collisions.

**Verification against real MySQL** (127.0.0.1:13306):
- A standalone stress harness (20 concurrent writers × 10 rounds, repeated 5×) against the new allocator: zero errors, zero collisions, zero deadlocks every round.
- The rewritten integration tests (8 concurrent writers each, covering both `WritePGSnapshot` and `TakeSnapshot`), repeated 5×: all green.
- Full test suite green: `go build ./...`, `go vet ./...`, `go test ./...`, `go test -tags integration ./...` (the one `internal/parser` failure in the full-parallel integration run is the pre-existing, previously-documented `#415` shared-binlog-window flake — passes in isolation, unrelated to this change).

Pushed to `fix/846-844-metadata` (`f5c19bb`).

Issue #784, opened 2026-07-05
recover: per-event generation errors demoted to a comment with exit 0 — partial script applies clean (silent incomplete reversal)


nethalo:
A per-event error in `buildStatement` (e.g. `RowBefore` nil because `UnmarshalRowImage` returns nil on a malformed/truncated JSON blob — best-effort by design, `query.go:595`) is demoted to a comment `-- ERROR ...` and generation CONTINUES: `GenerateSQLFromRows` returns `(written, nil)`, the CLI reports success and exit 0.

Location: `internal/recovery/recovery.go:257`

The code comment and `docs/query-and-recovery.md:352` justify it with "the transaction will roll back on first error anyway" — FALSE: a SQL comment does not fail on apply; the partial script commits clean, leaving an incomplete reversal — exactly the "silently-wrong recovery" that #654/#601/#592 refuse loud. Inconsistent posture within the same file.

Fix: return an error (or exit != 0 + refusal) when `written < len(rows)`, aligned with the drift refusal; and correct the docs claim.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

Issue #601, opened 2026-06-23
recover: reversal SQL references columns dropped/renamed after the event -> fails to apply against the current table


nethalo:
## Summary

`bintrail recover` builds reversal SQL from the stored before/after row images, which are **self-describing** (the column names captured at event time). If a column has been **dropped or renamed** between the event and now, the generated `INSERT`/`UPDATE` references a column the **current** table no longer has → the SQL **fails to apply**.

Distinct from the time-travel display bug (#600): that one is about which columns are *shown*; this one is about generated recovery SQL that won't *execute* against today's schema. Arguably worse — it's the recovery path itself, the reason dbtrail exists.

## Mechanism

- `internal/recovery` generates SQL from `row_before`/`row_after` JSON (keys = column names as they were at capture time).
- A `DROP COLUMN` / `RENAME COLUMN` after the event leaves the captured image referencing the old column.
- The reversal SQL (e.g. `INSERT INTO orders (id, coupon_code, total) VALUES (...)`) is rejected by MySQL when applied to the post-DDL table (`Unknown column 'coupon_code'`).

## Confidence

Mechanism is clear from the self-describing-image design; **repro pending** (not yet run). Should be repro'd (real DDL + an actual `recover` run, then attempt to apply) before fixing — repro > cita.

## Design options (decide on triage)

1. **Detect + warn/annotate:** compare the image's columns against the current schema and emit a loud warning / SQL comment naming the drifted columns, so the operator reconciles manually.
2. **Schema-aware emission:** target the schema in effect at the event time and document that the operator must apply against a compatible table.
3. **Guarded SQL / skip-with-note:** emit column-existence guards, or skip drifted columns with a recorded note.

The "fail loud — never silently emit SQL that errors, or worse, *partially* applies" stance matches the project's no-silent-loss posture.

## Relationship

Named sibling of #600 (same schema-drift family). Different surface, different mechanism: **apply-time**, not display shaping. The third sibling is `reconstruct --output-format mydumper` (baseline-projection drop).

nethalo:
**PostgreSQL impact:** `recover` (delta-only) is the **supported PG recovery surface for beta** (per #593), and the reversal SQL is generated from the same self-describing before/after images — so this apply-time mismatch applies to **PG today, not gated**. Under PG `REPLICA IDENTITY FULL` the before-image carries every column, so a reversal referencing a since-dropped/renamed column will fail to apply against the current PG table. Worth a PG repro alongside the MySQL one.

PR #933, opened 2026-07-10
cascade: buffer generator output in EmitSQL so refusals leave no dangling FK-disable

### LINKED ISSUE #835: cascade: a victim whose statement fails is demoted to a SQL comment with no caveat and exit 0
`EmitSQL` delegates to `gen.GenerateSQLFromRows`, whose loop demotes any `buildStatement` error to a comment `-- ERROR generating reversal for event 0` (synthetic victims carry `EventID=0`, which also makes the comment untraceable) and continues. The returned statement count is never reconciled against `len(rows)+len(setNullStmts)`, so an un-renderable victim-row disappears from the script with exit 0 and Complete=true — exactly the contract the rest of the subsystem (Incomplete caveats, pre-write refusals for drift/budget/TOAST) works to avoid.

Location: `internal/cascaderecover/emit.go:136`

Scenario: a child with a value `formatValue` doesn't support (e.g. a residual deferred marker from the #652 class) is lost from the recovery with no signal beyond a comment buried in the SQL.

Fix: in the cascade path, count expected statements and turn the deficit into a caveat/error, or pass a strict mode to the generator.

---

nethalo:
## What

Issue #835 reported that `EmitSQL` let a `buildStatement` failure demote to a `-- ERROR` comment with exit 0. The generator half of that was fixed by #784 (b699c27): `GenerateSQLFromRows` now accumulates per-event failures and refuses the whole script. This PR fixes the residual in the cascade path and adds the count reconciliation #835 asked for:

- `EmitSQL` wrote its preamble — including `SET FOREIGN_KEY_CHECKS=0;` — to `w` **before** calling the generator, so the render-time refusals (#784 generation failures, #601 schema drift) fired with the preamble already on the writer. In the CLI text path that partial script (an FK-disable with no re-enable, zero statements) was flushed to `--output` despite the non-zero exit. The reversal statements are now rendered into a buffer before any byte reaches `w`, making `EmitSQL` all-or-nothing for every generator refusal — the same contract its hoisted budget (#654) and TOAST (#592) checks already enforce. Peak memory stays bounded by the script budget enforced up front.
- The generated statement count is reconciled against `len(rows)`: a deficit (a synthesized cascade victim vanishing from the script without an error) is now a refusal by construction, so the silent-incomplete hole cannot reopen even if the generator regresses. (The SET NULL statements are pre-built all-or-nothing already, so `len(setNullStmts)` cannot desync.)

Byte output on the success path is unchanged (golden tests pass untouched); error precedence (budget → TOAST → SET NULL build → generator) is preserved.

## Tests

- `TestEmitSQL_generationFailureRefusesCleanly` — a renderable parent DELETE plus an un-renderable synthesized victim (nil before image, `EventID=0`): loud error carrying the #784 refusal, `n=0`, **zero bytes written**. Fails on the pre-fix code: it wrote 706 bytes ending in a dangling `SET FOREIGN_KEY_CHECKS=0;` (verified via stash).
- `TestEmitSQL_rowsRenderBetweenPreambleAndFooter` — success path guard for the buffering reorder: preamble → INSERT → FK re-enable order and statement count pinned.

```
go build ./...                                                            ok
go vet ./internal/cascaderecover/                                         ok
go test ./internal/cascaderecover/... ./internal/cli/... ./internal/console/... ./internal/recovery/... -count=1   ok
go test -tags integration ./internal/cascaderecover/... ./internal/cli/... ./internal/console/... -count=1        ok (MySQL 13306)
```

Closes #835


nethalo:
Applied a review finding on top of the #835 fix.

**Gap**: the fail-loud refusal for a failing synthesized cascade victim (`partialGenerationError`, via `EmitSQL` → `GenerateSQLFromRows`) rendered as `event 0: row_before is nil for DELETE event (event_id=0)` for *every* failing victim, because synthesized cascade victims never carry a real `event_id` (all `EventID==0`, per this PR's own comment on the row literal in `emit_test.go`). With multiple failing victims the refusal entries were byte-identical, so there was no way to tell which child row was actually unrecoverable — and the shared remediation text ("narrow `--pk/--pks`") only helps narrow the recovery *root*, not the victim.

**Fix**: `genFailure` (`internal/recovery/recovery.go`) now also carries `SchemaName`/`TableName`/`PKValues`, taken from the `query.ResultRow` that failed (already populated for cascade victims in `internal/cascade/cascade.go`). `partialGenerationError` renders `schema.table pk=...: <reason>` for `EventID==0` failures instead of the untraceable `event 0` form; normal (non-cascade, non-zero `EventID`) failures are unchanged.

Strengthened `TestEmitSQL_generationFailureRefusesCleanly` (`internal/cascaderecover/emit_test.go`) to assert the traceable `schema.table pk=...` form appears and the old `event 0:` form does not.

Tests: `go build ./...`, `go vet ./internal/recovery/... ./internal/cascaderecover/...`, `go test ./internal/recovery/... ./internal/cascaderecover/... -count=1` (pass), `go test -tags integration ./internal/recovery/... -count=1` (pass, MySQL 13306; `cascaderecover` has no integration tests).

PR #934, opened 2026-07-10
verify: default mode reports never-baselined snapshot tables as inconclusive instead of going green

### LINKED ISSUE #770: verify: default mode goes green on tables it never looked at (enumerates only from baselines, not the schema snapshot)
`runVerifyBaselinePair` derives the table universe solely from `FindBaselinePair` (baseline files). A table that exists in the source and in `schema_snapshots` but was NEVER baselined (baseline job scoped with `--database app1` or `--tables`; table created after the job was configured) produces NO row in the report and the run exits 0 with "N match" — exactly the false assurance `printVerifyReport` claims to prevent, and precisely the table that is also irrecoverable via reconstruct (no baseline = no never-touched rows).

Location: `internal/cli/verify.go:133`

Also: `docs/verify.md` (flags table) states the default of `--tables` is "all tables in the latest schema snapshot", which is only true in live mode.

Fix: cross the latest `schema_snapshot`'s tables against the baseline pair and report each never-baselined table as inconclusive with an explicit detail ("never baselined; unrecoverable via reconstruct"), just as `prevOnly` is already handled.

---

nethalo:
## What

`runVerifyBaselinePair` (baseline-anchored default mode) derived its table universe solely from `verify.FindBaselinePair`, i.e. from the baseline files on disk/S3. A table present in the source and in the latest `schema_snapshots` but NEVER baselined — baseline job scoped with `--database`/`--tables`, or the table created after the job was configured — produced **no row** in the report and the run exited 0 with "N match": exactly the false assurance `printVerifyReport` exists to prevent, over precisely the table that is also irrecoverable via `reconstruct` (no baseline = no never-touched rows).

## Fix

- New `metadata.Resolver.AllTables()` enumerates the loaded snapshot's full table set (sorted schema, table). `runVerifyBaselinePair` crosses it against the `pairs`/`unpaired`/`prevOnly` sets and reports each never-baselined table as **inconclusive** with the detail `never baselined; unrecoverable via reconstruct (extend the baseline job to cover this table)` — mirroring the existing `prevOnly` handling: visible, but not by itself a failure (a default run with real matches still exits 0; an all-inconclusive run still fails via the existing "nothing proven" gate). The resolver is the same latest-snapshot resolver `runVerify` already loads — no extra index query.
- A `--tables` request naming a never-baselined snapshot table now lands on that inconclusive row; the absent-request `StatusError` path now fires only for tables present in **neither** the baselines **nor** the snapshot (its detail says so).
- `docs/verify.md`: the `--tables` flags-table default claim ("all tables in the latest schema snapshot") was only true in live mode; with this change it holds in baseline-anchored mode too, and the row now states the never-baselined-inconclusive behavior explicitly (the "What it does not prove" section already promised this reporting — the code now matches it).

## Tests

- `TestRunVerifyBaselinePair_NeverBaselined` (unit): snapshot has `orders`+`payments`, baselines only `orders`; `payments` must report inconclusive "never baselined", not the `--tables`-absent error. **Fails before the fix** (old output: `mydb.payments error ... requested via --tables but not present in the latest baseline pair`).
- `TestRunVerifyBaselinePair_EndToEnd_NeverBaselined` (integration): DEFAULT no-`--tables` path against real MySQL — report shows `1 match` (orders verified), `payments` inconclusive "never baselined", `1 inconclusive`, exit 0. **Fails before the fix** (old output: `1 match, 0 mismatch, 0 inconclusive, 0 error` with `payments` absent — the exact false green from the issue).
- `TestAllTables` (unit): full, deterministically sorted enumeration; empty resolver.
- `TestRunVerifyBaselinePair_TablesAbsent` updated to pass an (empty) in-memory resolver; its fail-loud contract is unchanged.

```
go build ./...                                                          ok
go vet ./internal/cli/ ./internal/metadata/                             ok
go test ./internal/cli/... ./internal/metadata/... -count=1             ok
go test -tags integration ./internal/cli/... ./internal/metadata/... -count=1  ok (MySQL 13306)
```

Closes #770

### DIFF
--- a/docs/verify.md
+++ b/docs/verify.md
@@ -120,7 +120,7 @@ diff tool involved.
 | `--source-dsn` | *(empty)* | Live source DSN. Pass it for **live-source** mode; omit for **baseline-anchored** mode |
 | `--baseline-dir` | *(empty)* | Local directory of baseline Parquet snapshots |
 | `--baseline-s3` | *(empty)* | S3 URL prefix of baseline snapshots (e.g. `s3://bucket/baselines/`) |
-| `--tables` | *(all)* | Comma-separated `schema.table` list (default: all tables in the latest schema snapshot) |
+| `--tables` | *(all)* | Comma-separated `schema.table` list (default: all tables in the latest schema snapshot; in baseline-anchored mode, snapshot tables with no baseline report `inconclusive` — "never baselined") |
 | `--no-archive` | `false` | Query live MySQL partitions only; skip Parquet archive discovery |
 | `--explain` | `false` | On a baseline-anchored mismatch, print a per-row drill-down |
 
--- a/internal/cli/verify.go
+++ b/internal/cli/verify.go
@@ -219,15 +219,37 @@ func runVerifyBaselinePair(cmd *cobra.Command, indexDB *sql.DB, resolver *metada
 	for _, d := range prevOnly {
 		covered[d.Schema+"."+d.Table] = true
 	}
+	var uncovered []*metadata.TableMeta
 	for _, tm := range resolver.AllTables() {
 		key := tm.Schema + "." + tm.Table
 		if covered[key] || (want != nil && !want[key]) {
 			continue
 		}
-		results = append(results, verify.TableResult{
-			Schema: tm.Schema, Table: tm.Table, Status: verify.StatusInconclusive,
-			Detail: "never baselined; unrecoverable via reconstruct (extend the baseline job to cover this table)",
-		})
+		uncovered = append(uncovered, tm)
+	}
+	if len(uncovered) > 0 {
+		// A table can be absent from the two most recent snapshots yet still
+		// have an older one on disk/S3 — reconstruct.FindBaseline's
+		// stale-fallback path will still find and use it (with a
+		// StaleWarning), so it's recoverable, just not verifiable against
+		// the current top-2 window. Distinguish that from a table with zero
+		// baselines ever, which reconstruct genuinely cannot serve, so the
+		// message doesn't send an operator into unnecessary re-baselining
+		// panic for a table that's actually fine (just stale).
+		everBaselined, err := verify.EverBaselinedTables(cmd.Context(), baselineSrc)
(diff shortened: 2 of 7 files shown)


nethalo:
Applied a confirmed must_fix from adversarial review: the never-baselined report loop conflated two different situations.

A table absent from the two most recent baseline snapshots (the window `FindBaselinePair` uses) was unconditionally reported as "never baselined; unrecoverable via reconstruct" -- but `reconstruct.FindBaseline` (the function `bintrail reconstruct`/the shim's `_snapshot` actually use) has a documented stale-fallback path: it will still find and use an OLDER baseline snapshot for that table (returning a `StaleWarning`). So a table baselined at, say, the 3rd-most-recent snapshot but not the top two is recoverable via reconstruct -- just from a stale snapshot -- while the old message told operators the opposite, risking unnecessary urgent re-baselining.

Fix: added `verify.EverBaselinedTables` (reuses `reconstruct.ListBaselines`, the same source of truth `FindBaselinePair` already scans, no reimplemented directory walking) to check for a baseline at any snapshot time, not just the top two. The report now splits:
- zero baselines ever -> unchanged "never baselined; unrecoverable via reconstruct"
- an older baseline exists -> new "not covered by the two most recent baselines; reconstruct will fall back to an older snapshot (stale)"

Both stay classified Inconclusive (verify still cannot confirm a match/mismatch for the current window either way) -- only the detail message changed to be accurate.

Empirically verified against MySQL 13306 with baselines at t1(-3h, orders+payments), t2(-2h, orders only), t3(-1h, orders only): `payments` now gets the stale-but-recoverable message; a table with zero baselines anywhere still gets the original message. New integration test `TestRunVerifyBaselinePair_EndToEnd_StaleButRecoverable` locks this in.

`go build ./...`, `go vet ./internal/cli/... ./internal/verify/...`, `go test ./internal/cli/... ./internal/verify/... -count=1`, and `go test -tags integration ./internal/cli/... ./internal/verify/... -count=1` all pass.

Issue #759, opened 2026-07-05
capture: crash-replay silently duplicates events (at-least-once, no dedup); docs wrongly claim natural dedup


nethalo:
The pipeline is at-least-once but the index dedups **nothing**: `binlog_events` has only an auto-increment PK (`schema.go:126`), `insertBatch` is a plain INSERT, and there is no cleanup of post-checkpoint events on startup (zero `DELETE FROM binlog_events` outside tests).

Location: `internal/indexer/indexer.go:187`

Scenario: `streamLoop` flushes batches when `BatchSize` fills (independent of the checkpoint ticker); `kill -9` between a flush and the next `saveCheckpoint` → resume from the durable checkpoint → MySQL re-sends the same events → inserted **again** with new `event_id`. On a busy source this duplicates up to `checkpointInterval` (10s) of events on every hard crash. The duplicates reach `recover` (reverse SQL emitted twice), `reconstruct` (event applied twice), and `verify`.

Worse: `docs/streaming.md:97` states these are "deduplicated naturally because the same GTID/position is just re-received from MySQL" — **false**, no layer dedups (query dedups only live-vs-Parquet by `event_id`, which differs here).

Fix: on resume in position/gtid mode, delete events with `(binlog_file,start_pos) > checkpoint` before streaming (single-writer makes this safe), or a unique key `(schema,table,pk_hash,binlog_file,start_pos,event_type)` + `INSERT IGNORE`; and correct the docs claim.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #937, opened 2026-07-10
query: deterministic snapshot LIMIT subset; rollover-safe binlog_file comparison in UntilPos

### LINKED ISSUE #839: query: FetchSnapshot applies LIMIT without ORDER BY — non-deterministic subset of baseline rows
`q += " LIMIT %d"` over `parquet_scan` without ORDER BY lets DuckDB pick N arbitrary rows (dependent on row-groups and parallelism).

Location: `internal/query/snapshot.go:94`

Scenario: `bintrail query --include-snapshot --limit 100` over a 1M-row baseline returns 100 different baseline rows between runs, mixed with events, with no indication the snapshot was arbitrarily sampled — non-reproducible results for a forensic tool. All share `EventTimestamp=baseline ts`, so `MergeAndTrim` doesn't stabilize the selection either (tie-breaks by synthetic EventID = scan order).

Fix: deterministic ORDER BY (e.g. by the archive columns or row index) before the LIMIT, or document/reject `--limit` with `--include-snapshot`.

---

### LINKED ISSUE #840: query: lexicographic binlog_file comparison in UntilPos breaks after the 7-digit rollover (.999999 → .1000000)
`buildQuery` (`query.go:347`) and the parquet mirror (`parquetquery.go:873`) use lexicographic `binlog_file < ?`. MySQL after `mysql-bin.999999` continues with `mysql-bin.1000000`, and '1000000' < '999999' lexicographically: a baseline anchor in the post-rollover file would misclassify ALL events of `.1xxxxx` files as before the anchor (and vice versa), corrupting reconstruct-to-anchor's exact cut with no error.

Location: `internal/query/query.go`

The `BinlogPos` doc (`query.go:110-112`) declares the "zero-padded names" assumption but nothing verifies it and it's not in user docs. Low probability (~1M binlog rotations) but long-lived servers with a small `max_binlog_size` can reach it.

Fix: compare by parsed numeric suffix.

---

nethalo:
## What

Two `internal/query` correctness fixes from the 2026-07-04 audit, sharing the query subsystem:

**#839 — `FetchSnapshot` LIMIT without ORDER BY.** `query --include-snapshot --limit N` pushed `LIMIT N` into `parquet_scan` with no ORDER BY, so DuckDB returned an arbitrary subset of baseline rows that differed between runs (row-group layout / scan parallelism dependent) — and `MergeAndTrim` couldn't stabilize it because all snapshot rows share the baseline timestamp and tie-break on the synthetic EventID, which was assigned in scan order. The limited scan now carries `ORDER BY ALL`: ties are byte-identical rows, so both the selected set and the EventID assignment are deterministic.

**#840 — lexicographic `binlog_file` comparison breaks at the `.999999 → .1000000` rollover.** `buildQuery` (live MySQL) and the `parquetquery` mirror bounded `Options.UntilPos` with `binlog_file < ?`, and `'1000000' < '999999'` lexicographically — an anchor near the rollover misclassifies whole files on the wrong side of the exact cut (verify's baseline-anchored reconstruction) with no error. Both mirrors now compare files by length first, then lexicographically: for one server's sequence (constant basename, suffix zero-padded to a minimum width) that equals numeric-suffix order across the rollover, and equal-length names — the universal case — keep the plain string comparison as the fast path. The `BinlogPos` doc now states the actual ordering contract.

Both mirrors emit the same predicate shape with the same bind args (`CHAR_LENGTH` on MySQL, character-counting `length()` on DuckDB).

## Tests

Fail-before verified by stashing the source fix and re-running the new tests:

- `TestFetchSnapshot_limitDeterministic` (real DuckDB, ids written in descending file order): pre-fix returned `{9, 8}` for `Limit=2`; post-fix `{1, 2}` with deterministic EventIDs across runs.
- `TestFetch_untilPosRollover` in `parquetquery` (real DuckDB over an `archive.BinlogEventColumns` parquet): pre-fix the anchor `mysql-bin.1000000@300` dropped the `mysql-bin.999999` event (`got map[2:true]`); post-fix keeps exactly `{1, 2}`.
- `TestFetch_untilPosRollover` in `internal/query` (integration, real MySQL): same scenario through `Engine.Fetch`, pins the `CHAR_LENGTH` predicate end-to-end.
- `TestBuildQuery_untilPos` / `TestBuildQueryFromFiles_untilPos` updated to the new predicate shape (file ×4 + pos ×1 args) in both mirrors.

```
go build ./...                                                     ok
go vet ./internal/query/ ./internal/parquetquery/                  ok
go test ./internal/query/... ./internal/parquetquery/... -count=1  ok
go test -tags integration ./internal/query/... ./internal/parquetquery/... -count=1   ok (MySQL 13306)
go test ./internal/verify/... ./internal/cli/... -count=1          ok (downstream consumers)
```

Closes #839
Closes #840


nethalo:
Found and fixed the same bug class at a third call site while reviewing this PR: internal/streamrun/streamrun.go's deleteEventsSinceCheckpoint (the resume-time dedup DELETE for #759, called on every `bintrail stream` resume in both position- and GTID-mode — deleteEventsSinceCheckpointGTID delegates to it).

Its WHERE clause used the same plain-lexicographic `binlog_file > ?` comparison that #840 fixed on the read paths (query.go/parquetquery.go). Past the mysql-bin.999999 -> mysql-bin.1000000 suffix rollover this inverts, so the DELETE wrongly matches the pre-rollover file and destroys already-indexed rows that the resume never touched — permanent data loss on a DELETE, not a read-path symptom.

Reproduced empirically against real MySQL (13306): rows at mysql-bin.999999:100 (pre-rollover), mysql-bin.1000000:100, and mysql-bin.1000000:300 (the real straggler). Calling deleteEventsSinceCheckpoint(db, "mysql-bin.1000000", 300) deleted 2 rows instead of 1 — the pre-rollover row was wrongly destroyed alongside the straggler.

Fix applies the same length-then-lexicographic compare this PR already introduced, mirrored for the `>` (later file) direction this call site needs (vs. buildQuery's `<` cut). Added two integration tests against real MySQL (TestDeleteEventsSinceCheckpoint_rolloverSafe, TestDeleteEventsSinceCheckpointGTID_rolloverSafe) that fail with "got 2" on the unfixed comparison and pass with the fix; updated the three existing sqlmock unit tests for the new placeholder count.

Pushed to this branch: 5080371

Generated with Claude Code

PR #938, opened 2026-07-10
verify: normalize TIME fsp padding and FLOAT/DOUBLE text rendering false-MISMATCHes

### LINKED ISSUE #794: verify: TIME(fsp>0) with an integer-second value renders "12:00:00" vs mydumper's "12:00:00.000" — false-MISMATCH not normalized
go-mysql v1.13.0 `timeFormat` returns `%02d:%02d:%02d` WITHOUT a fractional suffix when `secPart==0`, regardless of the declared fsp; MySQL's text protocol and mydumper always pad to the fsp ("09:00:00.000" for TIME(3)). `renderCell` passes raw strings and `normalizeRenderedBytes` only covers zero-date and JSON, and TIME is not in `isDeferredType`.

Location: `internal/verify/render.go:368`

Scenario: `appointments(id PK, start TIME(3))` with an UPDATE in the window and `start='09:00:00'` → "09:00:00" vs "09:00:00.000" → conclusive false MISMATCH in both modes.

Fix: in `normalizeRenderedBytes`/`renderCellNormalized`, for dataType "time" re-format the fraction to the declared fsp (`ColumnType`) symmetrically on both sides — same recipe already used for zero-date/JSON.

---

### LINKED ISSUE #795: verify: FLOAT/DOUBLE in live mode — MySQL text-render (my_gcvt, "1e16") differs from Go (strconv, "1e+16") → conclusive false-MISMATCH on intact columns
The recon side renders the Parquet `float64/float32` with `strconv.FormatFloat('g',-1)` and the source side gives MySQL's text; they differ in exponent format and thresholds (DOUBLE 1e16 intact from the baseline → "1e+16" vs "1e16" → permanent MISMATCH of the table in live mode, with no corruption).

Location: `internal/verify/render.go` (documented decision at `render.go:34-39`)

`render.go:34-39` documents the choice: floats deliberately out of the deferred set so a float divergence is never masked — correct fail-safe direction, but the cost is that live mode is unusable on tables with floats outside the coinciding-render range, and `docs/verify.md` doesn't list it among the causes (its inconclusive list doesn't even apply: this comes out as mismatch). Baseline-anchored mode is unaffected (both sides render with Go).

Fix: symmetric parse+reformat normalization by dataType float/double (implementation, not fundamental).

---

nethalo:
## What

Two conclusive false-MISMATCH classes in `bintrail verify` were pure representation gaps — same stored value, different text renderer on each side of the comparison:

- **#794 — TIME(fsp>0) with integer seconds.** go-mysql v1.13.0's `timeFormat` omits the fractional suffix entirely when the microsecond part is zero (and pads it to the declared fsp otherwise), while MySQL's text protocol and mydumper always pad: `"09:00:00"` vs `"09:00:00.000"` for TIME(3). False MISMATCH in both verify modes; TIME is not in the deferred set, so it was conclusive.
- **#795 — FLOAT/DOUBLE in live mode.** MySQL's my_gcvt text (`"1e16"`, `"0.00001"`) vs Go's strconv (`"1e+16"`, `"1e-05"`) for the same intact value → permanent conclusive MISMATCH on any table with floats outside the coinciding-render range.

## Fix

Both normalized in `normalizeRenderedBytes` (`internal/verify/render.go`) — the seam already shared symmetrically by `renderCellNormalized` (recon side) and the `ConsistentTableChecksumNormalized` hook (live-scan side), same recipe shape as the existing zero-date/JSON normalizations. Keyed by `information_schema` DATA_TYPE:

- **`time`**: trim trailing fractional zeros (and a then-empty `.`) to the minimal value-preserving form. Trim rather than pad-to-fsp because the live-scan hook carries only DATA_TYPE (no COLUMN_TYPE/fsp), and trimming also covers pre-#212 snapshots whose ColumnType is empty. Since go-mysql pads any non-zero fraction to the fsp, the omitted all-zero suffix is the only divergence; removing trailing fractional zeros can never collapse two different values.
- **`float`/`double`**: parse + reformat through Go's shortest-round-trip formatter at the column's own width (32-bit FLOAT / 64-bit DOUBLE). Distinct storable values keep distinct shortest renderings, so only representation collapses. **FLOAT/DOUBLE stay OUT of the deferred-representation set** — the documented no-masking decision in the `renderCell` contract holds; a surviving divergence is a genuine value difference and still reports as a (safe) mismatch. Unparsable bytes return unchanged and fall through to the raw byte comparison.

## Tests

Unit (`render_test.go`): normalization matrices for both types incl. anti-masking pins (distinct TIME fractions and one-ulp-apart doubles stay distinct; non-time/float dataTypes untouched; unparsable/whitespace floats untouched), plus cross-origin agreement (baseline float64 / event `json.Number` / MySQL raw text). Integration (`verify_integration_test.go`): `TestVerifyTable_TimeFspAndFloatText_IsAMatch` runs the real live-source scan — TIME(3) integer-second event image + baseline-origin DOUBLE 1e16 must MATCH, and a tampered `1.0000000000000002e16` must still MISMATCH.

Verified fail-before: with `render.go` reverted to HEAD, the new unit tests and the integration test all fail (`"1e16" != "1e+16"`, `"09:00:00" != "09:00:00.000"`, `VerifyTable` = mismatch).

```
go build ./...                                                        ok
go vet ./internal/verify/                                             ok
go test ./internal/verify/... ./internal/consistency/... -count=1    ok
go test -tags integration ./internal/verify/... ./internal/consistency/... -count=1   ok (MySQL 13306)
```

Closes #794
Closes #795

### DIFF
--- a/internal/consistency/checksum.go
+++ b/internal/consistency/checksum.go
@@ -181,10 +181,14 @@ func consistentTableChecksum(ctx context.Context, db *sql.DB, schema, table stri
 	}
 
 	// Scan every row, hashing as we stream — no full-table buffering.
+	//
+	// promoteFloat is scoped to the normalize != nil (cross-renderer) callers
+	// only — see selectExpr's FLOAT case for why.
+	promoteFloat := normalize != nil
 	selectList := make([]string, len(cols))
 	res.Columns = make([]string, len(cols))
 	for i, c := range cols {
-		selectList[i] = selectExpr(c)
+		selectList[i] = selectExpr(c, promoteFloat)
 		res.Columns[i] = c.name
 	}
 	query := fmt.Sprintf("SELECT %s FROM %s.%s",
@@ -349,11 +353,35 @@ func tableColumns(ctx context.Context, conn *sql.Conn, schema, table string) ([]
 // which is NOT what mydumper dumps or the binlog carries — so the digest would
 // not match a baseline. The CAST makes the canonical form parseTime-independent.
 // Other types (including TIME and YEAR, which the driver never parses) are read
-// as-is.
-func selectExpr(c column) string {
-	switch c.dataType {
-	case "date", "datetime", "timestamp":
+// as-is, EXCEPT FLOAT when promoteFloat is set (see below).
+//
+// promoteFloat controls a FLOAT-only widening: bare `SELECT f` renders through
+// MySQL's my_gcvt truncated to ~6 significant digits (FLT_DIG) — lossy
+// relative to the underlying binary float32 for any value that needs more
+// precision to round-trip (#795). `f+0e0` promotes the read to DOUBLE
+// arithmetic, so my_gcvt renders the value's full double-precision decimal
+// expansion (enough digits to identify the exact float32), which
+// canonicalFloatText(_, 32) then folds back to the same shortest float32
+// text a Go-side renderer produces for the identical stored value — closing
+// the gap without touching the recon side (see internal/verify/render.go's
+// canonicalFloatText doc).
+//
+// Scoped to promoteFloat (wired to normalize != nil, i.e. only
+// ConsistentTableChecksumNormalized's live-source verify caller) rather than
+// applied unconditionally: ConsistentTableChecksum's un-normalized digest must
+// stay byte-identical to mydumper's own FLOAT text (both MySQL's ordinary
+// my_gcvt truncation, matching today) so the baseline-dump-time fidelity
+// check (internal/baseline, #633) keeps comparing like renderings — widening
+// unconditionally would turn THAT comparison into a false MISMATCH instead
+// of fixing this one. `f+0e0` (not CAST(f AS DOUBLE), which needs MySQL
+// 8.0.17+) is plain arithmetic promotion, safe on the whole MySQL 8.0+ floor
+// and on MariaDB.
+func selectExpr(c column, promoteFloat bool) string {
+	switch {
+	case c.dataType == "date" || c.dataType == "datetime" || c.dataType == "timestamp":
 		return "CAST(" + quoteIdent(c.name) + " AS CHAR)"
+	case c.dataType == "float" && promoteFloat:
+		return quoteIdent(c.name) + "+0e0"
 	default:
 		return quoteIdent(c.name)
 	}
--- a/internal/consistency/checksum_test.go
+++ b/internal/consistency/checksum_test.go
@@ -120,6 +120,31 @@ func TestDigestVersionOf(t *testing.T) {
 	}
 }
 
+// TestSelectExpr_FloatPromotion pins selectExpr's scoping of the #795 FLOAT
+// widening: unconditional promotion would break the byte-identity
+// ConsistentTableChecksum (nil normalize, baseline-dump-time fidelity check)
+// promises against mydumper's own unpromoted FLOAT text, so promoteFloat must
+// gate strictly on the cross-renderer (normalize != nil) caller.
+func TestSelectExpr_FloatPromotion(t *testing.T) {
+	f := column{name: "f", dataType: "float"}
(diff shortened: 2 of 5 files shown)


nethalo:
Adversarial review caught a real gap and I've pushed a fix (`d292835`).

**Confirmed:** `canonicalFloatText` alone does not close #795 for FLOAT columns. On the project's MySQL 8.0.46 test instance (127.0.0.1:13306), a bare `SELECT f` for `f FLOAT` (the exact codepath `internal/consistency.selectExpr` used — FLOAT is never CAST-wrapped) truncates to ~6 significant digits via my_gcvt:

```
INSERT INTO ft VALUES (1.234567) -> SELECT f -> "1.23457"   (7th digit lost)
```

That's lossy, not just differently formatted — no amount of parse+reformat after the fact can recover a digit that was never read. DOUBLE doesn't have this problem (my_gcvt renders DOUBLE losslessly via the text protocol), which is why the original fix converged for DOUBLE and not FLOAT. Confirmed `canonicalFloatText("1.23457", 32)` != `canonicalFloatText("1.234567", 32)` — exactly the false MISMATCH class #795 was filed to close, just narrowed to FLOAT.

Also confirmed the test gap: the only FLOAT case in `render_test.go` ("0.30000001192092896" -> "0.3") needs just 1 significant digit and never exercises the truncation path, and the integration test only added a DOUBLE column.

**Fix:** the missing precision has to come from the SQL text itself, not from reformatting after the read. `internal/consistency`'s `selectExpr` now promotes FLOAT reads to DOUBLE arithmetic (`` `f`+0e0 ``) so my_gcvt renders the full double-precision expansion — enough digits to identify the exact float32 — which `canonicalFloatText(_, 32)` then folds to the same shortest float32 text the Go-side (event image / baseline) renderer produces.

This promotion is scoped to the cross-renderer caller only (`normalize != nil`, i.e. `ConsistentTableChecksumNormalized`, which `VerifyTable` uses). It deliberately does **not** apply to the plain `ConsistentTableChecksum` (nil normalize) used by `internal/baseline`'s dump-time fidelity check — that check's contract is "byte-identical to mydumper's own text," and mydumper dumps FLOAT with the same unpromoted my_gcvt truncation, so widening it there would turn *that* comparison into a false MISMATCH instead. Verified `internal/baseline`'s integration tests (including the digest-fidelity keystone) still pass.

Used `` f+0e0 `` rather than `CAST(f AS DOUBLE)` since the latter needs MySQL 8.0.17+ and the documented BYO floor is MySQL 8.0+.

**Tests added:**
- `internal/consistency/checksum_test.go`: unit test pinning `selectExpr`'s promoteFloat scoping (FLOAT gets `+0e0` only when promoteFloat=true; DOUBLE/DATE/TIME unaffected).
- `internal/verify/render_test.go`: a `canonicalFloatText` case exercising the promoted (17-digit) source text, plus an end-to-end pin (`TestRenderCellNormalized_FloatEventAndPromotedMySQLTextAgree`) showing convergence with the recon-side float32 rendering, and an anti-regression check that the *old* 6-sig-fig text does NOT converge (so a revert of the SQL-side fix is caught here too).
- `internal/verify/verify_integration_test.go`: extended `TestVerifyTable_TimeFspAndFloatText_IsAMatch` with a FLOAT column (`gauge`) touched by an in-window UPDATE, needing 7 significant digits. Confirmed empirically that this test fails without the checksum.go fix (`mismatch (content digest differs at equal row count)`) and passes with it.

**Test results** (MySQL 127.0.0.1:13306):
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./... -count=1` — all packages pass
- `go test -tags integration ./internal/verify/... -count=1` — pass
- `go test -tags integration ./internal/consistency/... -count=1` — pass
- `go test -tags integration ./internal/baseline/... -count=1` — pass (dump-time fidelity check unaffected)

PR #939, opened 2026-07-10
cascade: per-root visited/emitted graph walk; FK snapshot selected at delete time

### LINKED ISSUE #831: cascade: global visited/emitted across roots — the second DELETE of the same parent PK is silently skipped
`visited` is keyed by `schema.table|pkvalues` without distinguishing the root event.

Location: `internal/cascade/cascade.go:249`

Scenario: parent pk=1 deleted at 10:00 (cascades children A), re-created with children B at 11:00, deleted again at 12:00 (cascades children B); `recover-cascade --since 09:00 --until 13:00` fetches both DELETE events, but the second is discarded in `visited[pkey]` — children B (events in (T1,T2], outside the root T1 window) are not reconstructed and NO `Incomplete` is added: exit 0 "complete". Sibling variant: a child cascade-deleted by root A, re-inserted, and cascade-deleted by another root B stays in `emitted` with the OLD image (roots processed ASC), restoring stale data with no signal.

Fix: key `visited`/`emitted` by `(root, table, pk)` or `(table, pk, rootTS)`, and dedup at the end keeping the newest-timestamp image.

---

### LINKED ISSUE #834: cascade: FK graph used is the latest snapshot, not the one in effect at delete time
`LoadCascadeFKs` takes `MAX(snapshot_id)` (LIMITATION comment at `cascade.go:496`). If an ON DELETE CASCADE FK is DROPPED after the delete and before the recover, the synthesis doesn't reconstruct those children and there's no caveat (silent under-recovery). If an FK is ADDED, victims are synthesized that were never deleted — that at least collides by PK on apply (loud).

Location: `internal/cascade/cascade.go:496`

Buildable: `fk_constraints` has historical snapshots and events carry `schema_version`; the snapshot `≤ T` of the root delete could be chosen. Documented only as a code comment; `docs/` doesn't mention it.

Fix: select the FK snapshot `≤ T` of the root delete.

---

nethalo:
## What

Two cascade-recovery correctness fixes in `internal/cascade`:

**1. Per-root `visited`/`emitted` keys (#831).** `SynthesizeVictims` keyed its `visited`/`emitted` (and the SET NULL dedup) globally by `schema.table|pk`, so when the same parent PK was deleted, re-created, and deleted again inside the recovery window, the second root DELETE was silently discarded: children created between the two deletes were never reconstructed, a re-deleted child kept its stale first-root image, and no `Incomplete` caveat was added (exit 0 "complete"). Keys now carry the originating root timestamp so each root delete is walked with its own `[since, T]` window; cross-root duplicates collapse after the walk keeping the newest-timestamp image (`dedupVictimsNewest` for victims, a newest-wins slot for SET NULL restores), so recovery still never double-INSERTs a PK and always restores the child's last known state.

**2. FK snapshot in effect at delete time (#834).** `LoadCascadeFKs`/`LoadCascadeFKsForParent` selected the FK graph with `MAX(snapshot_id)`. An `ON DELETE CASCADE` FK dropped after the delete and re-snapshotted before the recover made synthesis skip those children with no caveat (silent under-recovery); an FK added after the delete synthesized victims that were never deleted. `fkSnapshotIDAt` now resolves the newest FK-bearing snapshot whose `schema_snapshots.snapshot_time` is at or before the root delete (2s slack for second-truncated binlog timestamps vs rounded snapshot DATETIMEs). Callers anchor on the earliest parent delete in the batch (`FKGraphAnchor`) so the graph predates every root. When no FK snapshot predates the delete (e.g. a backlog re-index), the earliest recorded graph is used and a caveat is surfaced alongside `Result.Incomplete` — never silently.

`LoadCascadeFKsForParent` gains an `at time.Time` parameter and a caveat return; the CLI (`recover-cascade`) and console cascade paths pass the anchor and append the caveat. The CLI integration fixture now also writes the `schema_snapshots` sibling row for its `fk_constraints` seed, matching the production invariant (both are written in one transaction).

## Tests

New:
- `TestSynthesizeVictims_sameParentPKDeletedTwice` (integration) — fails on pre-fix code exactly as #831 describes (`childc:20` missing, `childc:10` stale `a0` image, SET NULL restore built from stale `n0`), passes after.
- `TestLoadCascadeFKs_snapshotInEffectAtDelete` (integration) — delete between two FK snapshots must use the older graph (pre-fix selects the latest), delete after the newest uses it, pre-history delete falls back to the earliest graph WITH a caveat.
- `TestDedupVictimsNewest` (unit) — newest-image collapse, first-seen order, distinct keys untouched.

Results (all `-count=1`):
- `go build ./...` OK, `go vet ./internal/cascade/ ./internal/cli/ ./internal/console/` OK
- `go test ./internal/cascade/... ./internal/cli/... ./internal/console/...` — ok / ok / ok
- `go test -tags integration ./internal/cascade/... ./internal/cli/... ./internal/console/...` — ok (24.9s) / ok (24.1s) / ok (31.9s)

Closes #831
Closes #834

### DIFF
--- a/internal/cascade/cascade.go
+++ b/internal/cascade/cascade.go
@@ -33,6 +33,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -300,20 +301,39 @@ func SynthesizeVictims(
 	// after its parent's last event, and skewing the re-parented check). Each
 	// layer item therefore carries the originating root T down the recursion;
 	// distinct root deletes keep their own T.
+	//
+	// rootID (the root's own EventID, the binlog_events auto-increment PK) is
+	// carried alongside rootTS and combines with it to KEY visited/emitted:
+	// event_timestamp is a whole-second DATETIME (no fractional part), so two
+	// genuinely distinct root deletes of the same parent PK landing in the
+	// same wall-clock second would collide on rootTS alone — the second
+	// root's subtree would be silently skipped at the visited[] check,
+	// reproducing #831 at sub-second granularity. EventID (the DB's
+	// auto-increment PK) never collides across real distinct rows, so pairing
+	// it with rootTS closes that gap.
 	type layerItem struct {
 		ev     query.ResultRow
 		rootTS time.Time
+		rootID uint64
 	}
 	layer := make([]layerItem, 0, len(parentDeletes))
 	for _, pd := range parentDeletes {
-		layer = append(layer, layerItem{ev: pd, rootTS: pd.EventTimestamp})
+		layer = append(layer, layerItem{ev: pd, rootTS: pd.EventTimestamp, rootID: pd.EventID})
 	}
 
 	for depth := 0; depth < opts.MaxDepth && len(layer) > 0; depth++ {
 		var next []layerItem
 		for _, item := range layer {
 			pev := item.ev
-			rootKey := strconv.FormatInt(item.rootTS.UnixNano(), 10)
+			// Keyed by BOTH rootID and rootTS (not either alone): rootID
+			// disambiguates two real, distinct rows that share a stored
+			// second (EventID is the DB's auto-increment PK, always unique
+			// for real rows); rootTS keeps distinguishing roots built
+			// without a real EventID (e.g. synthetic fixtures in tests,
+			// where EventID defaults to its zero value) as long as their
+			// timestamps differ, preserving pre-existing behavior for that
+			// case. Two roots collide only if BOTH match.
+			rootKey := strconv.FormatUint(item.rootID, 10) + "@" + strconv.FormatInt(item.rootTS.UnixNano(), 10)
 			pkey := pev.SchemaName + "." + pev.TableName + "|" + pev.PKValues + "|" + rootKey
 			if visited[pkey] {
 				continue
@@ -519,7 +539,7 @@ func SynthesizeVictims(
 					}
 					victims = append(victims, victim)
 					// May itself be a parent (grandchildren); keep the SAME root T.
-					next = append(next, layerItem{ev: victim, rootTS: item.rootTS})
+					next = append(next, layerItem{ev: victim, rootTS: item.rootTS, rootID: item.rootID})
 				}
 
 				// Phase-2 augmentation: add the baseline children that referenced
@@ -578,7 +598,7 @@ func SynthesizeVictims(
 								RowBefore:      br.Row,
 							}
 							victims = append(victims, victim)
-							next = append(next, layerItem{ev: victim, rootTS: item.rootTS})
+							next = append(next, layerItem{ev: victim, rootTS: item.rootTS, rootID: item.rootID})
 						}
 					}
 				}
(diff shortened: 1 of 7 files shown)


nethalo:
Applied fixes for 3 review findings on this PR (all verified empirically before fixing).

**#1 and #3 (same underlying bug, both must_fix) — CONFIRMED and fixed.** `FKGraphAnchor` picked ONE FK-topology snapshot for the whole batch, anchored on the earliest root delete. A later root in a multi-root recover-cascade (`--pks`/`--since`/`--until`) whose FK topology changed mid-window was silently recovered against the wrong graph — dropping real cascade victims or fabricating ones that never existed, with no caveat (exit 0 "complete"). Verified with a throwaway before/after test: pre-fix, a batch of [root A @09:10 under RESTRICT, root B @10:00 under CASCADE with a real child victim] returned `Complete=true Victims=[]` — root B's genuine cascade victim silently vanished.

Fix: `GroupParentDeletesByFKGraph` now resolves the FK graph independently for each root's own timestamp and buckets consecutive same-graph roots into one `SynthesizeVictims` call (so the common no-topology-change case still makes just one call); `MergeResults` combines the per-group results with the same cross-root newest-wins semantics `SynthesizeVictims` already applies within a single call. CLI and console `recover-cascade` now loop over groups instead of a single anchored call. `LoadCascadeFKsForParent` additionally returns the resolved snapshot ID so callers can group without comparing FK slices.

New regression test: `TestGroupParentDeletesByFKGraph_multiRootTopologyChange`.

**#2 (must_fix) — CONFIRMED and fixed, narrower than originally described.** `visited`/`emitted` keyed a root purely by `event_timestamp.UnixNano()`; since `event_timestamp` is a whole-second `DATETIME` with no fractional part, two distinct root deletes landing in the same wall-clock second collide on an identical key, silently skipping the second root's cascade walk — reproducing #831 at sub-second granularity.

One nuance from verification: at exact-same-second collision the two roots' `[since, until]` windows are byte-identical (both derived purely from the shared timestamp), so in the common case skipping the duplicate walk is redundant rather than lossy — the `emitted`/`vkey` map also embeds the same collided key, so even a from-scratch second walk would dedup to the same result. The scenario where it's actually lossy is when the FK's `ReferencedColumn` is a non-PK unique column whose value differs between two incarnations of the same PK (delete → recreate with a different unique value → delete again, both landing in the same second) — then each root's cascade genuinely matches different children, and the second is dropped. The regression test (`TestSynthesizeVictims_sameSecondRootsNotCollapsed`) is built on exactly that to get a deterministic (non-flaky, no real-time sub-second race) repro.

Fix: roots are now keyed by `(EventID, event_timestamp)` together — `EventID` is the `binlog_events` auto-increment PK and never collides between real rows; keeping `event_timestamp` in the composite preserves existing behavior for hand-built test fixtures that don't set an `EventID` (confirmed necessary — an `EventID`-only key broke the existing `TestSynthesizeVictims_sameParentPKDeletedTwice`, whose fixture rows don't come through `eng.Fetch` and so default to `EventID=0`).

All cascade/cli/console unit + integration tests pass (MySQL on :13306), including the full pre-existing cascade suite (multi-root, phase-2 baseline, self-ref, composite PK, etc.) — no regressions.

Issue #793, opened 2026-07-05
verify: GEOMETRY/VECTOR in neither isDeferredType nor base64StoredKind — touching (or carrying) a spatial column gives a conclusive false-MISMATCH


nethalo:
go-mysql delivers GEOMETRY (and VECTOR) as `[]byte` (`indexer.go:342`), `marshalRow` base64-encodes it into the event JSON, and `DecodeEventBinaries` never decodes it because `base64StoredKind` (`reconstruct/fulltable.go:994`) only covers blob/text/json. The source/baseline side renders the raw internal binary (SRID+WKB). Since "geometry"/"point"/"linestring"/"polygon"/… are also not in `isDeferredType`, `classify` doesn't degrade: it reports a conclusive MISMATCH.

Location: `internal/verify/verify.go:295`

Scenario: `stores(id INT PK, location POINT)` with any UPDATE in the window (FULL row image carries `location` even if unchanged) → verify permanently red, the operator investigates nonexistent corruption; in cron the gate stays red until the table is manually excluded.

Fix (minimal/safe): add the geometry/vector family to `isDeferredType` (honest Inconclusive); full fix: include it in `base64StoredKind` as binary and decode it like a BLOB.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #940, opened 2026-07-10
verify: map ENUM/SET labels + BIT bytes on the event side; narrow the deferred-Inconclusive gate to unresolved values

### LINKED ISSUE #769: verify: false-MISMATCH in default (baseline-anchored) mode — deferredReprChanged ignores ENUM/SET/BIT carried-but-unchanged in FULL row images
With `row_image=FULL` (a hard requirement), the `row_after` of an UPDATE carries ALL columns, including ENUM/SET/BIT/VARBINARY that did not change — and the event side renders them differently than the baseline: ENUM/SET as ordinal `json.Number` (go-mysql gives int64; verify NEVER calls `MapEventEnumLabels`, unlike reconstruct/shim/console — verified: zero call sites in `internal/verify`), BIT as decimal `uint64` ("5") vs the baseline/text-protocol raw bytes. `deferredReprChanged` returns true only if the deferred column is in `ev.ChangedColumns`, so the common case escapes.

Location: `internal/verify/verify_baseline.go:191`

Scenario: `orders(id INT PK, state ENUM('active','inactive'), amount DECIMAL)`; between two baselines runs `UPDATE orders SET amount=amount+1 WHERE id=1` → recon renders `state="1"`, new baseline renders `"active"` → equal row count, different digest, `deferredReprChanged=false` → `StatusMismatch "content digest differs at equal row count"` → cron/CI red with a false positive that reads as real corruption, in the DEFAULT mode. JSON and BLOB/TEXT are already covered (canonicalizeJSONContainer #692 + DecodeEventBinaries #672); the hole is ENUM/SET/BIT/BINARY/VARBINARY.

Fix: call `MapEventEnumLabels` like every other reconstruction surface (closes ENUM/SET at the root with genuine match) and/or change the gate to "deferred column PRESENT in the row image" (equivalent to the live gate), not "changed".

---

### LINKED ISSUE #791: verify: live-mode deferred gate is too broad — masks real divergence as Inconclusive
`VerifyTable` uses `deferredRepr = hasDeferredRepr(orderedCols) && len(changes) > 0`: it's enough for the table to CONTAIN an ENUM/SET/JSON/binary/BIT column and for ANY event to exist in the window for a content difference at equal row-count to be degraded to Inconclusive.

Location: `internal/verify/verify.go:188`

Scenario: a `wp_posts`-like table with a JSON column; someone corrupts an INT column in-place via `sql_log_bin=0` (the drift class live mode exists to catch); the window has any UPDATE → verify reports Inconclusive, and since inconclusive doesn't fail the run as long as ≥1 match exists elsewhere, CI exits 0 with real divergence present.

Fix (do NOT port `deferredReprChanged` as-is — that imports the baseline-mode false-MISMATCH): close the event-side representation gaps (MapEventEnumLabels, BIT→bytes) and then narrow the gate to "deferred present in some change's row image".

---

nethalo:
## What

Fixes two complementary defects in `internal/verify`'s deferred-representation handling, both rooted in the same fact: with `row_image=FULL`, an UPDATE's `row_after` carries **every** column, not just the changed ones.

**#769 (baseline-anchored / default mode — false MISMATCH):** the event side rendered carried-but-unchanged ENUM/SET columns as ordinals (`json.Number`) and BIT as decimal text, while both baselines carry the label string and the raw `ceil(M/8)` bytes. `internal/verify` never called `reconstruct.MapEventEnumLabels` (every other reconstruction surface does), and the old `deferredReprChanged` gate keyed on `ChangedColumns`, so `UPDATE orders SET amount=amount+1` on a table with an ENUM produced a conclusive `mismatch` — cron/CI red on a faithful recovery.

**#791 (live mode — masked real divergence):** the gate `hasDeferredRepr(orderedCols) && len(changes) > 0` degraded every equal-row-count content difference to Inconclusive as soon as the table merely *contained* a deferred column and any event existed in the window — so an in-place corruption of a plain INT/VARCHAR column (the `sql_log_bin=0` drift class live mode exists to catch) exited 0.

## How

Close the representation gaps at the root, then narrow the gate — per the issues' own fix direction:

- Both verify modes (and `ExplainBaselinePairMismatch`, which must stream the same rows the digest hashed) now run `reconstruct.MapEventEnumLabels` before building the change map.
- `renderCell` renders a numeric BIT value as the raw big-endian `ceil(M/8)`-byte string MySQL's text protocol / mydumper / the baseline Parquet carry (width parsed from `ColumnType` `bit(M)`; wider-than-declared values are never truncated, so genuine divergence still differs).
- `reconstruct.DecodeEventBinaries` now reports whether epoch typing was available for every event (a degraded decode silently leaves stored base64 in place — a representation gap, not a divergence). Existing best-effort callers are unaffected (return value ignored).
- New shared gate `deferredReprUnresolved`: fires only when an INSERT/UPDATE row image carries a deferred-typed value the normalization passes provably could **not** resolve — unmapped/out-of-range ENUM ordinal, non-ASCII label (charset-ambiguous), width-less BIT (pre-#212 snapshot), JSON that fails canonicalization or contains number literals, binary families without epoch typing, spatial/VECTOR (no decode exists, #793). Inconclusive still beats a false MISMATCH; a conclusive digest difference on non-deferred columns is no longer masked. Row-count differences remain conclusive and are never downgraded (unchanged).
- Baseline-anchored mode computes the gate **before** the merge: `SnapshotFullTableImages` drains the changes map, which had silently made the old gate unreachable there (it always saw an empty map). Found by the new unmappable-ordinal test.

JSON **number** literals deliberately stay unresolved: go-mysql renders a JSONB double `1.0` as `"1"` at capture (the parser does not set `UseFloatWithTrailingZero`), and after the round-trip the integer form is indistinguishable from a genuine integer — treating it as resolved would import a new false-MISMATCH class into the default mode.

## Tests

- `go build ./...` — ok
- `go vet ./internal/verify/... ./internal/reconstruct/...` (plus `-tags integration`) — ok
- `go test ./internal/verify/... ./internal/reconstruct/... -count=1` — ok (new: `TestDeferredReprUnresolved` 15-case matrix, `TestRenderCell_BitRendersSourceBytes` incl. the BIT(12) zero-padding case, `TestJSONRenderConclusive`)
- `go test -tags integration ./internal/verify/... ./internal/reconstruct/... -count=1` — ok against MySQL 13306, including three new tests:
  - `TestVerifyBaselinePair_EnumBitCarriedUnchanged_IsAMatch` — the #769 repro (fails as `mismatch` before the fix, `match` after)
  - `TestVerifyBaselinePair_UnmappableEnumOrdinal_Inconclusive` — residual safety: an out-of-range ordinal degrades to Inconclusive, never a false MISMATCH (this test caught the drained-map flaw)
  - `TestVerifyTable_DeferredColumnsResolved_RealDivergenceStaysMismatch` — the #791 repro against the real MySQL text protocol: phase 1 genuine MATCH with ENUM/BIT(12)/JSON columns in the window, phase 2 conclusive MISMATCH after corrupting a non-deferred column directly on the source (was Inconclusive/exit-0 before)

Closes #769
Closes #791


nethalo:
Applied the review finding after verification.

**Applied**: `deferredValueUnresolved`'s `"json"` branch went straight to `jsonRenderConclusive` without checking `binariesTyped`, unlike the binary-family branch. Confirmed the gap is real: `base64StoredKind` (`internal/reconstruct/fulltable.go`) explicitly includes `"json"` as a text-decode target `DecodeEventBinaries` degrades, and `binariesTyped` reports whether *any* base64-stored column — TEXT/JSON included, not just binary — failed to decode at its epoch. `jsonRenderConclusive` happens to reject most undecoded base64 strings because base64's alphabet rarely forms valid bare JSON, but a base64 string equal to the literal true/false/null (all within the base64 alphabet, and each a non-container, number-free JSON literal) would slip through as "resolved" while still undecoded. Added the same `binariesTyped` guard the binary-family branch already has, plus a regression test case (`json with typing unavailable`, value `"true"`, `binTyped=false`) pinning the fix.

Commit: b58480e

Tests: `go build ./...` clean; `go vet ./...` clean; `go test ./internal/verify/...` and `go test -tags integration ./internal/verify/...` pass (including the new case); `go test ./internal/reconstruct/...` and `-tags integration` also pass (unaffected, sanity-checked since the fix reasons about `DecodeEventBinaries`'s contract).

PR #941, opened 2026-07-10
shim: client-disconnect cancellation, --query-timeout, and connection/full-table concurrency caps

### LINKED ISSUE #823: shim: queries without timeout or client-disconnect cancellation, and connections without a cap — pile-up of 100k-row merges
`runPointInTime`/`runFullTable`/`runDiff`/`runSnapshot*` use `context.WithCancel(context.Background())` whose cancel only fires on RETURN — a client disconnect doesn't interrupt an in-flight `FetchMerged` (HandleCommand is blocked inside HandleQuery), and there's no deadline. `serveLoop` (`internal/cli/shim.go:353`) accepts connections with no limit.

Location: `internal/shim/handler.go:445`

Scenario: an ORM with a 30s client-timeout runs a heavy full-table `_snapshot` (S3+DuckDB+index merge, >30s), the client cuts and retries in a loop → each attempt stacks another buffered merge of up to 100k rows (~50MB post-merge, multiplied pre-merge by the number of archive sources — `handler.go:48` admits it) + orphaned DuckDB/S3 fetches → shim OOM (kills ALL forensic sessions) or saturation of the index DB shared with the streamer. Same class as #654 but in a long-running daemon.

Fix: wire the ctx to the TCP connection (cancel on Close), a `--query-timeout` flag, and a semaphore on concurrent full-table queries.

---

nethalo:
## What

The shim's five query entry points (`runPointInTime`/`runFullTable`/`runDiff`/`runSnapshot*`) ran `FetchMerged` under `context.WithCancel(context.Background())` — no deadline, cancel only on return — and `serveLoop` accepted connections without limit. The MySQL protocol is strictly request/response, so nothing read the socket while a query resolved: a client that timed out and closed (the classic 30s-ORM-retry loop) went unnoticed until the fetch completed, stacking orphaned 100k-row full-table merges + DuckDB/S3 fetches until the daemon OOMed or saturated the index DB shared with the streamer.

Three bounds, per the issue's proposed direction, all surfaced as typed MySQL wire errors:

1. **Client-disconnect cancellation** — `handleConn` now moves all socket reads to a dedicated pump goroutine (`watchConn`) relaying bytes through a synchronous `io.Pipe`; the protocol layer reads the pipe. The pump is always parked in `Read` on the real socket, so a FIN/RST is observed **mid-query** and cancels the per-connection context that `Handler.BindConnContext` roots every query context in — the in-flight `FetchMerged` aborts immediately. The same context closes the socket on SIGTERM, so graceful shutdown no longer blocks forever on idle clients parked in `HandleCommand`. TLS auth upgrades (`caching_sha2`/`sha256`) layer transparently over the pump (existing wire tests pass unchanged); go-mysql never arms read deadlines unless a readTimeout option is set, which the shim never does.
2. **`--query-timeout`** (default `5m`, `0` disables; env `BINTRAIL_SHIM_QUERY_TIMEOUT`) — per-query deadline covering the index fetch, archive fetch, and the full-table slot wait. Expiry maps to **1317** `ER_QUERY_INTERRUPTED` naming the flag. `wrapFetchError` coalesces with the query context's own error because drivers differ in their cancellation sentinels (go-sql-driver wraps `ctx.Err()`; sqlmock/DuckDB return their own).
3. **Concurrency caps** — `--max-connections` (default `100`, `0` unlimited; env `BINTRAIL_SHIM_MAX_CONNECTIONS`) refuses over-cap connections with a pre-handshake **1040** "Too many connections" ERR packet (the same first-packet ERR a real mysqld sends). `--max-fulltable-queries` (default `4`, `0` unlimited; env `BINTRAIL_SHIM_MAX_FULLTABLE_QUERIES`) gates concurrent full-table reconstructions — the heaviest path, up to `FullTableRowCap` buffered rows each — on one process-wide semaphore (`shim.Gate`); waiters die at the query deadline with **1203** `ER_TOO_MANY_USER_CONNECTIONS` naming the cap and flag. The `_snapshot` full-table path acquires the gate **after** its binlog-only fallback branch (which delegates to `runFullTable`, which acquires it itself) so a cap-1 gate cannot deadlock; point-lookups and `_diff` are never gated.

Defaults bound the worst case (gate 4 × ~50MB post-merge, orphan reaping ≤ timeout) while leaving normal interactive use untouched; every limit has a documented `0` opt-out (`docs/time-travel-sql.md` Resource limits section + error-code table).

## Tests

New, with fail-before verified:

- `internal/shim/limits_test.go` — `Gate` semantics (nil-unlimited, saturation returns ctx error, release cycle); `queryContext` backward-compat pin (no deadline by default); `--query-timeout` reaps a 10s-delayed fetch in ~100ms with 1317 (pre-fix: blocks the full delay); `BindConnContext` cancel aborts an in-flight fetch with 1317 "client disconnected"; gate saturation → 1203 with cap+flag in the message, then run-and-release once the slot frees.
- `internal/cli/shim_limits_test.go` — `watchConn` unit tests (bytes pass through; client close / parent cancel → context canceled); wire-level `--max-connections` (raw conn holds the slot, real client gets 1040, closing the conn readmits); wire-level `--query-timeout` (real MySQL client gets 1317 in ~300ms against a 30s-delayed index fetch); wire-level disconnect (client abandons a query → the server-side index connection is released within ms, observed via `db.Stats().InUse`). **Fail-before verified** by temporarily neutering the pump: both disconnect tests went red (`server-side fetch still holds 1 index connection(s) 5s after the client disconnected`).

```
go build ./...                                                                 ok
go vet ./internal/shim/ ./internal/cli/ ./cliapp/                              ok
go test ./internal/shim/ ./internal/cli/ ./cliapp/ -count=1                    ok
go test -tags integration ./internal/shim/ ./internal/cli/ ./cliapp/ -count=1  ok (MySQL 13306; 13.5s/16.6s/15.5s)
```

Closes #823


nethalo:
Applied both review findings from the adversarial pass.

**1. [must_fix] discarded fetch error — applied.** Confirmed empirically: `wrapFetchError` overwrote `err` with `ctx.Err()` whenever the query context was already done, and nothing logged the original error anywhere — reproduced with a throwaway test showing an `AccessDenied` S3 error vanish entirely in favor of a bare "query canceled" message. Pre-PR, `wrapFetchError` had no `ctx` param and always wrapped the real error; the `ctx`-override in this PR is what introduced the loss. Fixed by threading `*slog.Logger` into `wrapFetchError` and logging the discarded error (`Warn`, with both `ctx_err` and `underlying_err`) before the override, at all 5 call sites (`runPointInTime`, `runFullTable`, `runDiff`, `runSnapshot` x2). Verified live in the updated `TestShim_QueryTimeoutWireError` / `TestShim_ClientDisconnectCancelsInFlightFetch` output — the log now shows e.g. `underlying_err="query failed: canceling query due to user request"` instead of silence.

**2. [should_fix] missing shutdown-during-active-query test — applied.** Confirmed: `TestServeLoopExitsOnContextCancel` explicitly notes `handleConn` is unreachable on its path (Accept never returns a connection), and `TestWatchConnCancelsOnParentCancel` only checks `watchConn`'s returned context, never running a live query through `handleConn`. `TestShim_ClientDisconnectCancelsInFlightFetch` covers client-initiated close, not parent-ctx (daemon shutdown) cancellation. Added `TestShim_ShutdownCancelsInFlightFetch` in `internal/cli/shim_limits_test.go`: drives `serveLoop` directly (bypassing `startTestShimFull`'s cleanup-time-only cancel), issues a query with a 30s-delayed fetch, cancels the parent ctx mid-flight (simulating SIGTERM) while the client's own context stays live, and asserts both that `serveLoop`'s `done` channel closes promptly (proving `wg.Wait` unblocks) and that the abandoned index fetch actually aborts (`db.Stats().InUse` returns to 0) — not just that the client eventually notices.

Verification: `go build ./...`, `go vet ./...`, and the full unit suite (`go test ./...`) all green across every package. Integration tests for the touched packages (`go test -tags integration ./internal/shim/... ./internal/cli/...`) against MySQL 13306 also pass.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Issue #996, opened 2026-07-10
console: embed a MySQL-protocol time-travel shim in `watch`, bound to the selected server


nethalo:
## Problem

Time-travel (`_flashback` / `_snapshot` / `_diff`) is only reachable today through the standalone `bintrail shim`, which is single-`--index-dsn`. But `bintrail-console watch` is multi-source: each monitored server streams into its **own** per-source index database (`bintrail_idx_<entry-id>`, `cmd/bintrail-console/monitor.go:29`). The console navigates by server and resolves the per-source bundle automatically via `connManager`; the shim does not.

The result is a bad operator experience. To time-travel a source you already monitor in the console you must:

1. Discover the per-source index DB name by hand (`SHOW DATABASES LIKE 'bintrail_idx_%'`, then match the one with a populated `schema_snapshots`).
2. Hand-build an `INDEX_DSN` pointing at it.
3. Bring up a **separate** container behind `--profile flashback` (or run `bintrail shim` manually), pinned to that one index.
4. Repeat all of the above for every other monitored source, since one shim serves exactly one index.

Everything in steps 1–3 is information the `watch` daemon already has resolved in memory.

## Proposal (Option B): `watch` embeds the shim, bound to the selected server

`bintrail-console watch` already hosts:
- the `reconstruct` / flashback engine in-process (used by the console's Time-travel tab and `/api/reconstruct`), and
- the `connManager` that maps every registered server → its per-source `bundle` (index DB, resolver, baseline source).

Have `watch` optionally expose a **MySQL-protocol time-travel port** wired to that same `connManager`, so the `_flashback` / `_snapshot` / `_diff` virtual schemas resolve against **whichever monitored server is selected** — no separate container, no `--profile flashback`, no hand-built DSN. Server selection reuses the console's existing per-request server-selection mechanism (e.g. the `X-Bintrail-Server` header analogue: a tenant-per-server mapping, or a `USE <server>.<schema>` convention on the wire).

### Sketch
- New opt-in flag on `watch`, e.g. `--flashback-listen 127.0.0.1:3308` (+ env `BINTRAIL_CONSOLE_FLASHBACK_LISTEN`), default off.
- Reuse `internal/shim` `Handler`, but source its per-tenant `bundle` (index DB + resolver + baseline dir) from the `connManager` instead of a one-shot `shim.yaml`. Each registered server becomes a shim tenant automatically; baseline-dir per server comes from the same registry entry the console already reads.
[...]

Issue #822, opened 2026-07-05
shim: _snapshot full-table degrades to binlog-only with only a server-side Warn — the client receives a partial table indistinguishable from a complete one


nethalo:
When the operator DID configure `--baseline-dir`/`--baseline-s3` (i.e. explicitly asked for full-table completeness) but `FindBaseline` returns `ErrNoBaseline` for this table/AsOf (table not yet baselined, or created after the last baseline), or the PK can't be resolved from the snapshot (`snapshot.go:91`), `runSnapshotFullTable` falls back to `runFullTable` and returns ONLY the rows with binlog activity in the window. The Warn is server-side; the wire-client has no indicator.

Location: `internal/shim/snapshot.go:117`

A DBA using the resultset of `SELECT * FROM _snapshot.t AS OF T` to decide a restore concludes the table had N rows when it had N+M. Contrasts with the shim's own AllowGaps criterion (strict default: "silent partial results are worse than a loud failure", `cli/shim.go:66`) and the 1104 cap that does fail loud.

Fix: with a baseline configured, `ErrNoBaseline` in the full-table shape should be a wire error (e.g. 1526 with an actionable message) or at least set the OK/EOF packet's warning-count so `SHOW WARNINGS` shows it.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #1007, opened 2026-07-11
shim: fold PostgreSQL baselines in single-row _snapshot (#1006)

### LINKED ISSUE #1006: shim: single-row _snapshot silently degrades to binlog-only for all PostgreSQL sources (baseline fold never applied)
## Summary

The shim's single-row `_snapshot` (`SELECT ... FROM _snapshot.<table> AS OF '<ts>' WHERE <pk> = <v>`) **silently degrades to the binlog-only `_flashback` path for every PostgreSQL source**, so the baseline fold is never applied. A row that existed at the AS OF instant but was not touched within the retained binlog window resolves to **zero rows** (as if it never existed) instead of its baseline image.

This contradicts `docs/postgres.md`, which states the shim's single-row `_snapshot` works against a PG baseline.

## Root cause

`internal/shim/snapshot.go` (`runSnapshotPointInTime`, ~line 437):

```go
if !baselinePKStringMatchable(dataType) {
    h.logger.Warn("shim: _snapshot PK type not safe for baseline lookup; using binlog-only path", ...)
    return h.runPointInTime(q)
}
```

`dataType` comes from `schema_snapshots.data_type`, which is the **empty string `""` for PostgreSQL sources** (`internal/metadata` populates `pg_type_oid`/`pg_type_mod` for PG, not the MySQL `data_type` token). `baselinePKStringMatchable("")` returns `false`, so the gate always trips and the handler falls back to binlog-only.

The gate never consults the source flavor. The offline `reconstruct` path was given a PostgreSQL bypass in #593 (slice D) — PG baselines store every column as raw pgoutput text (`baseline.Column.RawText`), so the `pkColumn = ?` string-bound match in `ReadBaselineRow` is a byte-for-byte identity join for **any** PK type — but that bypass never reached this shim gate.

## Scope

- **Affected:** `bintrail-pg shim` / `bintrail shim` single-row `_snapshot`, and the embedded flashback port in `bintrail-console watch` (#996 routes through the same `internal/shim/snapshot.go`).
- **Not affected:** CLI `reconstruct` and the console HTTP `/api/reconstruct` endpoint (they use `reconstruct.ApplyAt`, not this gate — verified working for PG).
- **Full-table `_snapshot`** already fails *loud* for PG (`SupportedPKType("")` → wire error telling the user to use `_flashback`), which is acceptable since full-table PITR is out of GA scope for PostgreSQL. Only the single-row path degrades *silently*.

Distinct from #921 (missing pre-pin rendering-GUC warn on the shim/console fold) — that is about a warning, this is the fold not running at all.

## Evidence (live, RDS PostgreSQL 16.14)

Against a real RDS source with a `bintrail-pg baseline` snapshot and a witness row (`id=7`) updated after the baseline:

- `reconstruct --pk 7 --at <pre-update>` → **baseline image** (correct).
- `_snapshot.orders AS OF '<pre-update>' WHERE id=7` → **empty** (wrong; should equal the baseline image).
- `_snapshot.orders AS OF '<post-update>' WHERE id=7` → the delta row (only because a delta exists — indistinguishable from `_flashback`).

Shim log confirms: `WARN shim: _snapshot PK type not safe for baseline lookup; using binlog-only path ... pk_column=id pk_type=""`.

## Fix

Bypass the `baselinePKStringMatchable` gate in `runSnapshotPointInTime` when `query.SourceFlavor(h.indexDB) == "postgres"` (short-circuit so the MySQL hot path never pays the extra read), mirroring the reconstruct-path bypass. Add a PostgreSQL integration test asserting the never-touched-since-baseline row resolves under `_snapshot` while `_flashback` returns nothing.

nethalo:
Closes #1006.

## Problem
The shim's single-row `_snapshot` silently degraded to binlog-only for every PostgreSQL source: `schema_snapshots.data_type` is `""` for PG (it records `pg_type_oid`, not the MySQL type token), so `baselinePKStringMatchable("")` was always false and the handler fell back to `runPointInTime`. A row present in the baseline but never touched in the binlog window resolved to zero rows — the baseline fold `docs/postgres.md` promises never ran. Also affects the embedded flashback port in `bintrail-console watch` (#996, same `internal/shim/snapshot.go` path).

## Fix
Bypass the PK-type gate in `runSnapshotPointInTime` when `query.SourceFlavor(h.indexDB) == "postgres"`. PG baselines store every column as raw pgoutput text (`baseline.Column.RawText`, #593), so `ReadBaselineRow`'s string-bound `pkColumn = ?` match is a byte-for-byte identity join for any PK type — the same reason the offline `reconstruct` path treats all PG PK columns as string-matchable (#593 slice D). The flavor read is short-circuited so the MySQL hot path (a matchable type) never pays for it.

Full-table `_snapshot` already fails *loud* for PG (out of GA scope); only the single-row path degraded silently. CLI `reconstruct` and the console HTTP `/api/reconstruct` use `reconstruct.ApplyAt` and were unaffected.

## Test
`TestSnapshotBaseline_PostgresPKTypeBypass` (integration): PG-flavored `stream_state`, `schema_snapshots` with empty `data_type`, a baseline row never touched afterwards — asserts `_snapshot` folds the baseline (id resolves) while `_flashback` returns nothing. Fails before the fix (0 rows), passes after.

## Validated live on RDS PostgreSQL 16.14
Built this branch, ran the shim against a real RDS source with a `bintrail-pg baseline`, witness row `id=7` updated after the baseline:

| query (`WHERE id=7`) | before fix | after fix |
|---|---|---|
| `_snapshot AS OF <pre-update>`  | (empty) | **77.77 / 7.0000007 / new** (baseline) |
| `_snapshot AS OF <post-update>` | 314.15  | 314.15 / 2.718281828 / paid (delta) |
| `_flashback AS OF <pre-update>` | (empty) | (empty) — binlog-only, unchanged |

### DIFF
--- a/internal/shim/snapshot.go
+++ b/internal/shim/snapshot.go
@@ -496,7 +496,9 @@ func (h *Handler) fullTableTextCell(schema, table, column string, v any) any {
 // canonicalization in the offline merge), so it is kept as a separate
 // matcher: a future change to one path's type support must not silently
 // move the other. Types reconstruct rejects outright (FLOAT, BLOB, BIT,
-// JSON, …) are not listed and therefore fall back to the binlog-only path.
+// JSON, …) are not listed and therefore fall back to the binlog-only path —
+// for MySQL-family sources. A PostgreSQL source bypasses this matcher entirely
+// (its data_type is "" and its baseline is raw text — see runSnapshotPointInTime).
 func baselinePKStringMatchable(dataType string) bool {
 	switch strings.ToLower(strings.TrimSpace(dataType)) {
 	case "int", "integer", "smallint", "tinyint", "mediumint", "bigint",
@@ -520,8 +522,10 @@ func baselinePKStringMatchable(dataType string) bool {
 // It falls back to the binlog-only point-lookup (runPointInTime) — with
 // a log line so the degradation is visible — in four cases:
 //   - no baseline source configured (the pre-#355 default),
-//   - the PK column's type isn't one the baseline matcher supports
-//     (a typed WHERE would silently miss the row),
+//   - the PK column's type isn't one the baseline matcher supports AND the
+//     source isn't PostgreSQL (a PG source carries an empty data_type but its
+//     raw-text baseline makes every PK a string-identity match — see the flavor
+//     bypass below; a typed WHERE would otherwise silently miss the row),
 //   - the schema resolver is unavailable (can't determine the PK type),
 //   - no baseline exists for this table at-or-before AsOf.
 //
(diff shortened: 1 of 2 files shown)


nethalo:
Multi-agent review (code / silent-failure / tests / comments) + rebased on `main` (0.31.0). Applied:

- **Rebased onto current `main`** — the branch was authored before #1005 rewrote `runSnapshotPointInTime`; the gate change and #1005's tail change don't overlap, rebase is clean, `go build` + `go vet -tags integration ./internal/shim/` green.
- **Fix the degrade log line** (silent-failure): for a PG-shaped row (`data_type=""`) whose flavor couldn't be confirmed, the old message blamed the (irrelevant, empty) PK type. It now names the real cause — an unconfirmed postgres flavor (transient `stream_state` read failure / legacy index) — and warns that a never-touched row will read as absent. MySQL non-matchable types keep the original message.
- **Soften the "any PK type" comment** to "string-identity join that can only recover the right row or find nothing — never a wrong row" (code-reviewer verified: PG baselines are `RawText`, so the worst case is a safe-empty degrade, never a wrong row).
- **Docstring accuracy** (comment-analyzer): scoped the `runSnapshotPointInTime` "four cases" list case 2 to non-PostgreSQL, and added a PG caveat to `baselinePKStringMatchable`'s doc.
- **Negative-control test** `TestSnapshotBaseline_NonPostgresPKTypeStaysGated` (test-analyzer, highest-value gap): a MySQL source with a JSON PK must STILL degrade to binlog-only (0 rows) — pins the flavor term as load-bearing, so a future refactor that widened the gate for every source is caught.
- **PG delta-fold coverage**: the PG test now also folds a baseline row updated after the snapshot (`id=8` → post-snapshot UPDATE wins), proving deltas merge on the PG path, not just that the baseline is read.

Design note (kept as-is, per review): the single-row path degrades log-only (established #355 contract) rather than failing loud like the full-table path (#822) — the log fix makes any degrade debuggable. `SourceFlavor` stays an uncached per-query read (only on the non-matchable path); code-reviewer judged it negligible vs the DuckDB read already on that path.

Core behavior was validated live on RDS PostgreSQL 16.14 (in the PR description); the gate logic is unchanged by these edits.

PR #1029, opened 2026-07-16
ext/cliapp: extension seams for embedding distributions (commands, read-plane API, doctor checks, agent commands, source jobs)

nethalo:
Adds the extension seams an embedding distribution — a build that imports `cliapp`/`ext` and wraps the core — uses to inject behavior without patching the core. **Every seam defaults to a no-op: the stock binary's behavior is unchanged.** Follows the existing seam convention (`ext.SetAuditSink`, #730): package-level registration, set once from `main()` before command dispatch, startup-only concurrency contract stated on every setter.

### 1. `indexquery` — exported read-plane facade (new top-level package)

Read-only programmatic access to the index (live MySQL partitions + Parquet archives merged) for tooling and embedding distributions:

- Type aliases to the internal query types so external modules can construct them: `Options`, `ResultRow`, `Engine`, `QueryPlan`, `FetchMergedOptions`, `ArchiveFetcher`, `GapError`, `SourceEmptyError`, `Tuning`.
- One-line wrappers: `New`, `FetchMerged`, `FormatGapWarning`, `Connect`, `ParseSourceDSN`, `EnsureSchema`, `WrapSchemaMigrationErr`, `AddDuckDBTuningFlags`, `DuckDBTuningFromFlags`, `TunedArchiveFetcher`, `OutputJSON`.
- The index schema stays owned by the core — `EnsureSchema` is the only schema-mutating entry point exposed.

### 2. `cliapp.AddCommands(cmds ...*cobra.Command)`

Registers additional top-level commands on the root command; call from `main()` before `cliapp.Main`.

### 3. `ext.RegisterSourceJob` / `ext.RunSourceJobs`

Daemon-lifecycle jobs scoped to a capture source (`SourceJobInfo{SourceDSN, IndexDSN, Flavor}`). Wired at `up`'s phase-3 start next to the rotation loop; jobs are secondary and must never be fatal to the stream.

### 4. `ext.RegisterDoctorCheck` / `ext.RunDoctorChecks`

Extra preflight checks appended to the report on both surfaces — `bintrail doctor` and `bintrail up` phase 1 — with the same FAIL-blocks-boot semantics as built-in checks. Status strings must match `doctor.CheckStatus` (`pass`/`fail`/`warn`/`skip`); an unknown status degrades to a counted WARN with a note instead of an uncounted malformed entry.

### 5. `ext.RegisterAgentCommand` / `ext.LookupAgentCommand`

Registry for agent WebSocket command types. `dispatch` consults it only when no builtin case matches — builtin command types always win, and with nothing registered an unknown type fails exactly as before. Duplicate registration is last-wins (documented). The agent runner builds the handler deps (`ext.AgentDeps{IndexDB, SourceDB, SourceDSN, SourceHost}`) once and hands them to the dispatch loop via the new `Channel.ExtDeps` field; payload unmarshalling is the handler's job (raw `json.RawMessage` in, `any` out into the response data, error into the response error).

### Tests

Unit tests per seam: registration→effect and default→no-op (save/restore of the package registries, `ext_test.go` style); dispatch-level end-to-end routing for the agent registry including builtin-precedence and unregistered-type pins; compile-level signature pins plus no-DB-work paths for `indexquery`.

### Verification

- `go build ./...`
- `go test ./... -count=1` (unit tier, all green)
- `go vet -tags integration ./...` (clean — integration test call sites updated for the extended `dispatch` signature)
- `gofmt -l` clean on all touched files

### DIFF
--- a/cliapp/addcommands_test.go
+++ b/cliapp/addcommands_test.go
@@ -7,12 +7,27 @@ import (
 )
 
 func TestAddCommandsRegistersOnRoot(t *testing.T) {
-	cmd := &cobra.Command{Use: "ext-test-cmd", Run: func(*cobra.Command, []string) {}}
+	ran := false
+	cmd := &cobra.Command{Use: "ext-test-cmd", Run: func(*cobra.Command, []string) { ran = true }}
 	AddCommands(cmd)
-	t.Cleanup(func() { rootCmd.RemoveCommand(cmd) })
+	t.Cleanup(func() {
+		rootCmd.RemoveCommand(cmd)
+		rootCmd.SetArgs(nil) // restore cobra's default os.Args handling
+	})
 
 	found, _, err := rootCmd.Find([]string{"ext-test-cmd"})
 	if err != nil || found != cmd {
 		t.Fatalf("rootCmd.Find(ext-test-cmd) = (%v, %v), want the injected command", found, err)
+	}
+
+	// Execute through the real root so the injected command runs under the
+	// actual PersistentPreRun chain (log setup) — AddCommands must dispatch,
+	// not merely register.
+	rootCmd.SetArgs([]string{"ext-test-cmd"})
+	if err := rootCmd.Execute(); err != nil {
+		t.Fatalf("rootCmd.Execute() = %v", err)
+	}
+	if !ran {
+		t.Fatal("injected command did not run through rootCmd.Execute")
 	}
 }
--- a/cliapp/addcommands_test.go
+++ b/cliapp/addcommands_test.go
@@ -1,0 +1,18 @@
+package cliapp
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+func TestAddCommandsRegistersOnRoot(t *testing.T) {
+	cmd := &cobra.Command{Use: "ext-test-cmd", Run: func(*cobra.Command, []string) {}}
+	AddCommands(cmd)
+	t.Cleanup(func() { rootCmd.RemoveCommand(cmd) })
+
+	found, _, err := rootCmd.Find([]string{"ext-test-cmd"})
+	if err != nil || found != cmd {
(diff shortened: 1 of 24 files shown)


nethalo:
Applied consolidated review hardening in 097411c:

- **Agent dispatch (registry branch)**: panic containment — a panicking extension handler now yields a per-command error response instead of killing the agent process; handler results are pre-marshalled at dispatch so an unmarshalable value (NaN/channel/cycle) becomes a per-command error rather than a `writeJSON` failure that tears down the connection into a reconnect loop. `(nil, nil)` documented as a valid empty success (data field omitted on the wire).
- **`ext.RunSourceJobs`**: each job runs on its own goroutine with a recover — the core enforces the "secondary, never fatal" contract; docs now name the actual wiring point (`bintrail up` startup) instead of promising every daemon.
- **Doctor**: registered extension checks run under the same 30s budget as `doctor.Build`; a panicking check converts to a FAIL entry so the report still renders; status strings are trimmed/lowercased before matching, and a truly unknown status keeps the WARN coercion plus a greppable `slog.Warn`.
- **`up`**: source jobs receive the flavor the stream actually runs with (`strmFlavor`, i.e. default or `BINTRAIL_SOURCE_FLAVOR`) instead of hardcoded `"mysql"`.
- **indexquery**: FetchMerged docs now carry the two non-obvious caveats (redaction requires `NoArchive=true`; `*QueryPlan` may be nil); exported the persisted event-type vocabulary (`EventType`, `EventInsert`…`EventSnapshot`, `EventTypeName`).
- **`ext.Register*`**: panic on nil handlers so misuse fails at startup; added `ext.ResetForTest` for cross-package test cleanup.

Tests: dispatch panic/unmarshalable/empty-success pins, a WS-level round-trip proving `Channel.ExtDeps` reaches registry handlers through the real `listenLoop`, doctor text+json rendering of a registered check via `runDoctorTo`, `AddCommands` executed through the real root command, and source-job panic containment. `go build ./...`, `go test ./... -count=1`, and `go vet -tags integration ./...` all green.

PR #1032, opened 2026-07-16
consoleapp: export the console entrypoint for embedding builds

nethalo:
## What

Moves the `bintrail-console` command layer out of `cmd/bintrail-console` (package `main`) into a new importable top-level **`consoleapp`** package — the console sibling of `cliapp`:

- **`consoleapp/root.go`**: the cobra root command, its persistent flags (`--log-level`/`--log-format`), and an exported `Main(version, commitSHA, buildDate string) int` that mirrors `cliapp.Main` — sets the build-metadata vars, formats `rootCmd.Version` identically to before, executes, prints the error to stderr itself, and returns the exit code. Same startup-only concurrency contract as `cliapp` / the `ext` seams.
- All command files moved via `git mv` (`serve.go`, `watch.go`, `monitor.go`, `monitor_replica.go`, `baseline.go`, `verify.go`, `flashback.go`, `user.go`, `env.go`) together with every `*_test.go`, including the `//go:build integration` ones. Subcommand `init()` registration on `rootCmd` is unchanged.
- **`cmd/bintrail-console/main.go`** is now a thin wrapper: the three `-ldflags`-injected build vars — names stay `main.Version`/`main.CommitSHA`/`main.BuildDate`, so the Makefile's `BINTRAIL_LDFLAGS` and the `.goreleaser.yaml` ldflags block keep working unchanged, as does the `./cmd/bintrail-console` build path — plus `os.Exit(consoleapp.Main(Version, CommitSHA, BuildDate))`.
- **Guard test** `consoleapp/thinwrapper_test.go` (`TestConsoleBinaryIsThinWrapper`, in the style of `cliapp`'s `TestCoreBinaryIsUIFree`): asserts via `go list -json` that `cmd/bintrail-console` compiles exactly one Go file (`main.go`) and imports `github.com/dbtrail/dbtrail/consoleapp`, pinning the thin-wrapper contract against regression.

## Why

Embedding distributions — builds that import `cliapp` and wrap the OSS core — could already wrap the core CLI (`cliapp.Main`) but not the console binary, whose behavior was locked inside `package main`. Exporting the entrypoint gives the console the same seam.

This is the **console slice of #943**; the mcp/pg entrypoints are left for a follow-up, so the issue stays open. Refs #943.

## Behavior-identical

`--version`, `--help`, and the unknown-command error path were verified **byte-for-byte identical** against a build of the unmodified tree; an ldflags-injected build prints `bintrail-console version 9.9.9-test (commit abc1234, built 2026-07-16)` as before. Forced edits beyond the package split (the complete list):

- `.github/workflows/ci.yml`: the `console` path filter gains `consoleapp/**` (console code no longer lives only under `cmd/bintrail-console/`).
- `consoleapp/flashback.go`: one file-local comment now says "It lives in consoleapp" instead of "in the binary".
- `consoleapp/root.go`: the `SilenceErrors` comment now says `Main()` instead of `main()`; `rootCmd.Version` formatting moved from `init()` to `Main` (same string, mirroring `cliapp`).

No logic edits otherwise.

## Test evidence

- `gofmt -l consoleapp cmd/bintrail-console` — clean
- `go build ./...` — OK
- `go test ./... -count=1` — 42 packages ok, 0 failures
- `go vet -tags integration ./...` — clean (integration-tagged files are never compiled by `go build`)
- `go test -tags integration ./consoleapp/ -count=1` — ok (Docker MySQL on 13306)
- `go run ./cmd/bintrail-console --version` / `--help` — diffed against the pre-change tree: identical (commands listed: `serve`, `watch`, `user`, `completion`, `help`)

Refs #943

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
               - '.github/workflows/ci.yml'
             console:
               - 'internal/console/**'
+              - 'consoleapp/**'
               - 'cmd/bintrail-console/**'
 
   unit:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Exported `consoleapp` package**: the web console's command layer (root command plus the `serve`/`watch`/`user` subcommands) moved from `cmd/bintrail-console` into the importable top-level `consoleapp` package, with `consoleapp.Main(version, commitSHA, buildDate)` as the entrypoint — the console sibling of `cliapp.Main`, for embedding distributions that import the OSS core and wrap it. `cmd/bintrail-console` is now a thin wrapper (the `-ldflags`-injected `main.*` build vars plus `os.Exit(consoleapp.Main(...))`), pinned by a guard test; binary behavior, flags, and the version string are unchanged.
+
 ## [0.34.0] - 2026-07-16
 
 ### Added
--- a/cmd/bintrail-console/main.go
+++ b/cmd/bintrail-console/main.go
@@ -12,15 +12,17 @@
 // ~/.config/bintrail/config.env) file is loaded on startup and the
 // BINTRAIL_INDEX_DSN / BINTRAIL_CONSOLE_* env vars are honored with
 // flag > env > default precedence.
+//
+// All command behavior lives in the importable consoleapp package; this
+// main exists only to receive the -ldflags-injected build metadata (which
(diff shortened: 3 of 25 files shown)


nethalo:
Superseded by #1031. A review finding on #1031 showed the console-auth seam is unreachable unless the console entrypoint is also exported, so #1031's fix pass performed the identical `consoleapp` extraction (same git-mv renames, same `Main` signature, same thin wrapper) alongside the seam. #1031's `thinwrapper_test.go` guard was ported over from this branch, so nothing here is lost. Closing to avoid two branches creating the same top-level package. #943 stays open for the mcp/pg entrypoint exports.

Issue #1061, opened 2026-07-20
CI trust tests: import boundary, field allowlist, no-credential, single-hook, MCP exclusion


nethalo:
## Summary

CI trust tests — the load-bearing artifacts. The privacy guarantees must be things a skeptical reviewer can verify in CI output in 30 seconds, not promises in docs.

## Tests

1. **Import-boundary (P0)**: assert via `go list -deps` that `internal/telemetry` imports **none of** `internal/serverid` (or wherever bintrail_id resolution lives), `internal/config`, `internal/byos`, `internal/parser`, `internal/indexer`, `internal/query`, `internal/recovery`. Structural guarantee: a package that cannot import `config`/`parser` cannot read a DSN or a row; one that cannot import server-identity code cannot read `bintrail_id`. This survives careless future contributors in a way a field allowlist alone does not (an allowlist is blind to value *provenance*).
2. **Allowlist + no-credential**: reflect over the event struct and fail on any field outside `allowedFields`; assert the request builder attaches **no `Authorization` header, no cookie, no credential, no account identifier**.
3. **Single-hook**: walk `rootCmd.Commands()` recursively for **all three binaries** (`cliapp`, `consoleapp`, `cmd/bintrail-pg`) and fail if any subcommand defines its own `PersistentPreRunE`/`PreRunE` — cobra runs only the closest hook, so a future shadow would silently un-instrument a subtree.
4. **MCP exclusion**: `go list -deps ./cmd/bintrail-mcp` must not contain `internal/telemetry` (same mechanism as the existing `uifree_test.go` decouple guard).
5. **Demo no-egress**: covered in the demo-image guard issue; reference it here so the suite is enumerated in one place.

## Acceptance

All tests run in the default CI lane (no build tags needed except where `go list` requires module context), fail loudly with actionable messages, and are referenced from TELEMETRY.md by name.

Part of #1054.

Issue #1059, opened 2026-07-20
wire telemetry into the three binaries (cliapp, consoleapp, bintrail-pg); exclude MCP + console JS


nethalo:
## Summary

Wire `internal/telemetry` into the three binaries. There is no shared bootstrap — each binary has its own cobra root and its own `PersistentPreRunE` (the same deliberate seam `internal/observe` uses), so the hook lands in three places:

- `cliapp/root.go` (`bintrail`)
- `consoleapp/root.go` (`bintrail-console`)
- `cmd/bintrail-pg/main.go` (`bintrail-pg`)

## Pattern (per binary)

```go
PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
    observe.Setup(os.Stderr, logFormat, logLevel)
    tel = telemetry.Init(resolveTelemetryConfig())  // also drains spool, async
    telSpan = tel.RecordCommand(cmd.Name())
    return nil
},
```

After `Execute()` returns in `Main()`: classify the error (`telemetry.ClassifyError`), `telSpan.Finish()` (local append only — never network). A deferred `recover()` around `Execute()` captures panics as `error_class="internal"` (**never** the panic message — it can contain DSNs/paths), best-effort appends, then re-panics so the user-visible crash is unchanged. Existing `os.Exit` paths are fine — `Finish()` already happened and delivery never depends on this process (spool model).

- Version fields come from each binary's ldflags vars (`cliapp.Version` / `consoleapp` version var / `bintrail-pg` `Version`), minor-truncated inside the telemetry package.
- `cmd.Name()` is the leaf command name only.

## Exclusions (explicit, in this issue)

- **MCP server (`cmd/bintrail-mcp`)**: no telemetry at all — it is invoked by an agent inside an IDE/LLM session; no human consent surface exists.
- **Console web UI**: no JS beacon, ever (the frontend stays zero-dependency); anything the console binary reports goes through the same Go-side hook as any other command.

New command = zero telemetry code (the hook covers it). Optional enrichment at interesting call sites: `telSpan.SetError(telemetry.ClassifyError(err))`.

## Acceptance

- All three binaries record command_run/command_error through the shared package (verified with an `httptest` endpoint in integration-style unit tests).
- Panic path test: induced panic produces `error_class=internal` and the original panic still surfaces.
- MCP binary provably clean: no `internal/telemetry` in `go list -deps ./cmd/bintrail-mcp`.

Part of #1054.

Issue #1054, opened 2026-07-20
Epic: opt-out usage telemetry (metadata-only, spool-based)


nethalo:
## Summary

Add opt-out usage telemetry to the bintrail binaries: which commands run, whether they succeed, coarse error classes, version/OS/arch mix. Purpose: roadmap prioritization (command mix) and reliability signal (error-class rates). Nothing else.

Telemetry is **enabled by default** with a prominent first-run notice and multiple easy, layered kill switches (`bintrail telemetry off`, `DO_NOT_TRACK=1`, `BINTRAIL_TELEMETRY=off`, `--no-telemetry`). This is a deliberate, documented tradeoff; the design constraints below exist to keep the payload strictly metadata-only and source-verifiable.

## Non-negotiable design constraints

1. **Closed field allowlist** (the entire wire payload): `schema_version`, `event_type` (`command_run|command_error|daemon_beacon`), `command` (leaf name only, never args/flag values), `outcome`, `error_class` (bounded enum, never `err.Error()`), `duration_bucket` (tail collapsed at `>10m`), `version` (minor-truncated, e.g. `0.40`), `is_release`, `os`, `arch` (`amd64|arm64|other`), `is_ci`, `is_interactive`, `run_id` (ephemeral UUIDv4, in-memory, per-process; **never** present in daemon beacons).
2. **NEVER-list** (enforced by CI, verbatim in TELEMETRY.md): DSNs/connection strings; hostnames; client IPs; schema/table/column names; PK values; row data; query text; file paths; server-id; server-uuid/bintrail_id; GTIDs; binlog file names/positions; flag values; positional args; verbatim error strings/panic messages; row counts; credentials.
3. **No persistent install ID.** No ID is ever written to disk. Cost accepted: no DAU/WAU/retention — only daily approximate volumes.
4. **No credentials on the wire, ever.** Dedicated ~20-line HTTP client in `internal/telemetry`; no `Authorization` header; separate host (`telemetry.dbtrail.io`), never a path under the authenticated API. Must NOT reuse `byos.MetadataClient`.
5. **No network in any command's request path.** Local NDJSON spool, drained at startup of a later invocation by a detached goroutine with a hard 2s deadline (Go-toolchain model).
6. **Build-time kill switch**: endpoint injected via `-ldflags -X`; empty default means plain `go build` / `go test` produces a telemetry-inert binary by construction (also the distro-packager guarantee and the test-suite guarantee).
[...]

Issue #1060, opened 2026-07-20
daemon telemetry beacons (stream, agent, console watch): ≤1/UTC day, off-thread, startup notice line


nethalo:
## Summary

Daemon telemetry for the long-running processes: `bintrail stream`, `bintrail agent`, `bintrail-console watch` (and `bintrail-pg stream`). Daemons follow the same opt-out default as interactive commands, with strict cadence and visibility rules.

## Rules

1. `Beacon()` **spawns its own goroutine per emission** — never a network call (or spool fsync) on the event-loop tick, checkpoint ticker, or WS heartbeat. A synchronous POST on the stream tick means replication lag; on the agent heartbeat it means reconnect storms.
2. **At most one beacon per UTC day** per daemon process — beacon cadence must not reconstruct uptime/maintenance windows at finer granularity. Plus one final off-thread beacon on `ctx.Done()` bounded by a short deadline.
3. `daemon_beacon` events carry **no `run_id`** — a months-lived process's run_id would be a longitudinal key.
4. **Startup visibility line** (stderr, once, when telemetry resolves to enabled):
   `telemetry: usage reporting is ON (metadata-only); disable with BINTRAIL_TELEMETRY=off or DO_NOT_TRACK=1 — if this host was cloned from an image, see TELEMETRY.md`
   The golden-image clause is load-bearing: one baked AMI/container layer otherwise silently applies one human's setting to a fleet.
5. Daemons never print the interactive first-run notice (non-TTY) — the startup line above is their disclosure surface.

## Acceptance

- Unit test: beacon rate-limiter (≤1/UTC day, day rollover), no run_id in beacon payload, shutdown beacon respects deadline.
- No beacon logic executes inside the stream/heartbeat tick paths (goroutine spawn only).
- Startup line asserted in daemon startup tests when enabled; absent when disabled.

Part of #1054.

nethalo:
Design constraint surfaced while reviewing #1064 — this issue must handle it, or daemon beacons silently never arrive:

`drain` runs **only from `telemetry.Init`**. A daemon calls `Init` once at startup and then lives for months, while `Beacon` merely appends to the local spool. On a host that runs only the daemon (the common production shape), beacons would spool and then age out after 7 days without ever being sent — the liveness signal, which is the entire point of the metric, would never work.

[...]

PR #1064, opened 2026-07-21
telemetry: metadata-only spool-based usage reporting (core + consent + command)

### LINKED ISSUE #1056: internal/telemetry core: event schema, error taxonomy, spool, sender, build-time kill switch
## Summary

New package `internal/telemetry`: event schema, error taxonomy, local spool, drain/sender, and the build-time kill switch. No wiring into commands yet (that's a follow-up) — this issue is the engine.

## Public API

```go
func Init(cfg Config) *Client          // resolves consent (see consent issue); drains spool async; no-op when disabled or endpoint==""
func (c *Client) RecordCommand(command string) *Span
func (s *Span) SetError(class string)  // outcome=error + error_class
func (s *Span) Finish()                // duration bucket + local spool append; never network
func (c *Client) Beacon(ctx context.Context) // daemons; own goroutine; ≤1/UTC day
func ClassifyError(err error) string
```

## Event schema

One flat struct, closed allowlist (see epic): schema_version, event_type, command, outcome, error_class, duration_bucket (`<100ms / 100ms-1s / 1s-10s / 10s-1m / 1m-10m / >10m`), version (minor-truncated), is_release, os, arch (`amd64|arm64|other`), is_ci, is_interactive, run_id (ephemeral UUIDv4; never in `daemon_beacon`).

## Error taxonomy (`errclass.go`)

`errors.Is`/`errors.As` mapping → `db_connection | db_permission | binlog_parse | binlog_not_found | schema_mismatch | config_invalid | flag_invalid | storage_io | network | not_found | internal | unknown`. Default `unknown`. **Never** the raw error message (error strings contain DSNs, hostnames, schema/table names, paths).

## Spool (never network in the request path)

- `Span.Finish()` appends one NDJSON line to `~/.config/bintrail/telemetry-spool/<date>.ndjson` — plain `O_APPEND|O_CREATE`, 0600, **no fsync** (hot path).
- Drain happens at **startup** of a later invocation inside `Init()`, detached goroutine, hard 2s deadline, concurrent with the command body. Sub-100ms commands deliver on the next run — accepted (Go toolchain model).
- **Claim-by-rename** against concurrent processes sharing $HOME (cron + manual): drainer atomically `os.Rename`s the active spool file to a unique temp name, drains that immutable snapshot, deletes it. No flock across network I/O, no read-then-truncate race, no double-send.
- Spool cap ~5MB / 7-day rotation — a permanently-offline box never grows unbounded.
- `os.UserHomeDir()` error or read-only config dir → silently disable everything (mirrors the env-loader skip-on-error pattern).
- Every entry point wraps `defer recover()` — a telemetry panic must never alter a command's behavior, output, or exit code.

## Sender

Dedicated ~20-line HTTP client: POST NDJSON/JSON to the endpoint, short timeout, drop on failure (events stay in the claimed file only until the drain attempt finishes; failed sends are dropped, not retried forever). **No `Authorization` header, no credential, no account identifier, ever.** Do not reuse `internal/byos.MetadataClient` (it is Bearer-authenticated by design — the exact pattern telemetry must not copy).

## Build-time kill switch

```go
// injected via -ldflags -X; empty default = telemetry physically inert
var telemetryEndpoint string
```

`Init()` returns a no-op client when empty. GoReleaser will set the real endpoint only when the ingestion side is live (separate decision); plain `go build` / `go test` yields an inert binary by construction. Unit tests inject an `httptest.Server` endpoint via `Config`.

## Import boundary

`internal/telemetry` must not import `serverid`, `config`, `byos`, `parser`, `indexer`, `query`, or `recovery` — a package that cannot import them cannot read a DSN, a row, or `bintrail_id`. Enforced by a CI test (tracked in the CI-trust-tests issue), but respect it from the first commit.

## Acceptance

- Unit tests: spool append/rotate/cap, claim-by-rename under concurrent drainers, duration bucketing, error classification, no-op behavior with empty endpoint, recover() safety.
- `go build ./...` produces an inert binary (no endpoint).

Part of #1054.

### LINKED ISSUE #1057: telemetry consent resolution, controls, and first-run notice (default ON)
## Summary

Consent/controls resolution and the first-run notice for opt-out telemetry.

## Resolution chain (highest wins)

1. `DO_NOT_TRACK=1` → hard off: no notice, no spool, no send. Checked first, before any I/O ([consoledonottrack.com](https://consoledonottrack.com) convention).
2. `--telemetry=on|off` / `--no-telemetry` (persistent root flag on each binary).
3. `BINTRAIL_TELEMETRY=on|off` — one new entry in `internal/cli/env.go` `EnvBindings` + the `config init` template (`cliapp/config.go`), inheriting flag > env > default for free.
4. State file `~/.config/bintrail/telemetry.json`: `{"enabled":..., "notice_shown":..., "decided_at":...}` — written only by `bintrail telemetry on|off` with the atomic temp+fsync+rename+0600 recipe (same as `internal/console/registry.go` `save()`; there is no exported helper — replicate it). The file carries **no identifier** and never goes on the wire.
5. **Default: ON.**

CI detection (`CI`, `GITHUB_ACTIONS`, `TF_BUILD`, `TRAVIS`, `CIRCLECI`, `JENKINS_URL`, `BUILDKITE`, `GITLAB_CI`) suppresses sending as defense-in-depth. Non-TTY suppresses the notice (never print the notice into piped output).

## First-run notice (stderr, interactive runs only, shown once, sentinel `notice_shown`)

```
bintrail sends metadata-only usage stats — command names, version,
OS/arch, and success/error class. No identifier is stored or sent, and
NEVER your data, schemas, tables, DSNs, hostnames, IPs, or file paths.

Telemetry is currently ON.
  bintrail telemetry off     turn it off
  bintrail telemetry show    see exactly what is sent (sends nothing)
  DO_NOT_TRACK=1             disable telemetry entirely

Details: https://github.com/dbtrail/dbtrail/blob/main/TELEMETRY.md
```

Honest wording is mandatory: state plainly that it is ON, and how to turn it off, before anything has been sent from this machine (first run only spools; delivery happens on a later run by design).

## Also

- Resolution result must expose *which source decided* the state (for `telemetry status`).
- `Init()` consumes this resolution; disabled means no spool write and no drain.

## Acceptance

Unit tests for the full precedence matrix (each level overriding the ones below), CI-env suppression, notice-once behavior, atomic state-file write, corrupt/absent state file → default ON without error.

Part of #1054.

### LINKED ISSUE #1058: bintrail telemetry subcommand: status | show | on | off
## Summary

`bintrail telemetry` subcommand — the user-facing transparency and control surface. Also registered on `bintrail-pg` and `bintrail-console` (shared implementation, `internal/cli`-style so all three binaries get it).

## Subcommands

- `telemetry status` — resolved state (on/off) **and which source decided it** (DO_NOT_TRACK / flag / env / state file / default), plus spool location and whether the binary has an endpoint compiled in at all.
- `telemetry show` — prints the **exact JSON bytes** that would be sent for a sample invocation; sends nothing (GitHub CLI `GH_TELEMETRY=log` precedent). This is the artifact docs and skeptical users point at.
- `telemetry on` / `telemetry off` — atomic write of `~/.config/bintrail/telemetry.json` (temp+fsync+rename, 0600).

## Notes

- `--format json` support consistent with the rest of the CLI (`IsValidOutputFormat`).
- `off` must confirm in output what is now disabled and that `DO_NOT_TRACK=1` is the stronger machine-wide switch.
- No network calls in any of the four subcommands.

## Acceptance

Unit tests: status source-attribution for each precedence level; `show` output round-trips as valid JSON and contains only allowlisted fields; on/off write and re-read; `show`/`status` work with empty endpoint (inert build).

Part of #1054.

nethalo:
closes #1056
closes #1057
closes #1058

Part of the telemetry epic #1054. Depends on #1063 (demo image guard) landing first.

The three issues are one PR because they are one unit: the core cannot be exercised without consent resolution, and neither is inspectable without the command. Splitting them would ship two PRs of dead code.

## What this is

`internal/telemetry` — event schema, error taxonomy, consent resolution, local spool, sender, build-time kill switch — plus the `telemetry` command registered on all three binaries. **No command records anything yet**; wiring the root hook is #1059.

## The guarantees, and how each is enforced

| Guarantee | Mechanism |
|---|---|
| Closed field allowlist | `AllowedFields` + a reflection test **and** a test over the serialized bytes (a custom marshaller could add keys reflection alone would miss) |
| No credential on the wire | `postNDJSON` sets only `Content-Type`; test asserts no `Authorization`/`Cookie`/api-key/server-uuid. Deliberately **not** `byos.MetadataClient` |
| No persistent identifier | Nothing but the consent file is written; `run_id` is per-process and in-memory, and beacons omit it |
| Source builds cannot send | `endpoint` is `-ldflags`-injected, empty by default → `Init` returns an inert client, no spool, no network |
| Telemetry cannot break a command | `recover()` at every entry point; nil `*Client`/`*Span` are safe; a corrupt consent file degrades to the default |
| Nothing on the request path | `Finish` appends one local NDJSON line; delivery happens on a **later** run, detached, 2s deadline |

## Verified against the real binary

```
$ go build ./cmd/bintrail && bintrail telemetry status
Telemetry:    OFF
Consent:      on (decided by: default)
Endpoint:     not compiled in — this build cannot send anything
```

With an endpoint injected it reports `ON` and the URL; `telemetry off` records the choice and `DO_NOT_TRACK=1` overrides everything (`decided by: DO_NOT_TRACK`). `telemetry show` prints the exact 13-field payload and performs no network access.

## Two design decisions worth a look

**Claimed spool files are removed only after the send attempt.** The first cut removed before sending, which loses data in the *common* case, not an exotic one: a sub-100ms command's process exits while the detached drain goroutine is still in its HTTP call. Now such a file survives as a claim and a later run adopts it (`TestDrainReclaimsAbandonedClaim`), while a claim younger than 5 minutes is left alone so a live drainer is never robbed (`TestDrainLeavesFreshClaimAlone`).

**`Beacon` takes no context and spawns no goroutine.** The design sketch had it spawn one so a POST could never land on a daemon's event loop. The spool model already guarantees that — a beacon is a local append — so the goroutine was ceremony.

## Test plan

- [x] `go test ./... -count=1` — full suite passes
- [x] `go vet` clean, `gofmt` clean
- [x] Package tests: allowlist (×2), no-credential, precedence matrix (6 cases), CI suppression, corrupt/absent state file, 0600 permissions on both files, spool cap, claim-by-rename under 4 concurrent drainers (exactly-once), abandoned-claim reclamation, age expiry, beacon ≤1/UTC day with day rollover, notice-once, nil safety, end-to-end HTTP delivery
- [x] Manual CLI exercise of status/show/on/off with a temp HOME, with and without a compiled-in endpoint
- [ ] Structural CI trust tests (import boundary via `go list -deps`, single-hook walk, MCP exclusion) — those are #1061

### DIFF
--- a/cliapp/root.go
+++ b/cliapp/root.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dbtrail/dbtrail/internal/agent"
 	"github.com/dbtrail/dbtrail/internal/cli"
 	"github.com/dbtrail/dbtrail/internal/observe"
+	"github.com/dbtrail/dbtrail/internal/telemetry"
 )
 
 var (
@@ -54,6 +55,10 @@ func init() {
 	// package and self-registered via init(); they moved to internal/cli so both
 	// binaries expose them.
 	cli.AddMaintenanceCommands(rootCmd)
+	// Usage telemetry control surface (status/show/on/off). Registered on every
+	// binary that can report, so `telemetry off` works from whichever one the
+	// operator has on PATH.
+	cli.AddTelemetryCommand(rootCmd)
 }
 
 // AddCommands registers additional top-level commands on the bintrail root
@@ -72,6 +77,7 @@ func AddCommands(cmds ...*cobra.Command) {
 func Main(version, commitSHA, buildDate string) int {
 	Version, CommitSHA, BuildDate = version, commitSHA, buildDate
 	rootCmd.Version = fmt.Sprintf("%s (commit %s, built %s)", Version, CommitSHA, BuildDate)
+	telemetry.SetVersion(version)
 
 	err := rootCmd.Execute()
 	if err == nil {
--- a/cliapp/root.go
+++ b/cliapp/root.go
@@ -47,6 +47,7 @@ binlog files still existing on disk.`,
 func init() {
 	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "Log level: debug, info, warn, error")
 	rootCmd.PersistentFlags().StringVar(&logFormat, "log-format", "text", "Log format: text or json")
+	rootCmd.PersistentFlags().String("telemetry", "", "Usage telemetry: on or off (overrides BINTRAIL_TELEMETRY; DO_NOT_TRACK=1 overrides everything)")
 	// Register the source-agnostic read commands that have moved to internal/cli
 	// (#529) so a future bintrail-pg can register the same set. Today: status.
 	cli.AddReadCommands(rootCmd)
--- a/cmd/bintrail-pg/main.go
+++ b/cmd/bintrail-pg/main.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/dbtrail/dbtrail/internal/cli"
 	"github.com/dbtrail/dbtrail/internal/observe"
+	"github.com/dbtrail/dbtrail/internal/telemetry"
 )
 
 var (
@@ -72,6 +73,9 @@ func init() {
 	// the same index DSN (#951). `bintrail-pg stream` additionally runs the
 	// built-in rotation loop for safe-by-default retention.
 	cli.AddMaintenanceCommands(rootCmd)
+	// Usage telemetry control surface, same set as the core binary.
+	cli.AddTelemetryCommand(rootCmd)
+	telemetry.SetVersion(Version)
 }
 
 func main() {
--- a/cmd/bintrail-pg/main.go
+++ b/cmd/bintrail-pg/main.go
@@ -63,6 +63,7 @@ func init() {
 	rootCmd.Version = fmt.Sprintf("%s (commit %s, built %s)", Version, CommitSHA, BuildDate)
 	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "Log level: debug, info, warn, error")
 	rootCmd.PersistentFlags().StringVar(&logFormat, "log-format", "text", "Log format: text or json")
+	rootCmd.PersistentFlags().String("telemetry", "", "Usage telemetry: on or off (overrides BINTRAIL_TELEMETRY; DO_NOT_TRACK=1 overrides everything)")
 	// The source-agnostic read plane (status/query/recover/reconstruct/shim),
 	// shared verbatim with the core binary. The PostgreSQL capture command
 	// (stream) is registered in stream.go's init().
--- a/consoleapp/root.go
+++ b/consoleapp/root.go
@@ -65,6 +65,7 @@ yourself.`,
 func init() {
 	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "Log level: debug, info, warn, error")
 	rootCmd.PersistentFlags().StringVar(&logFormat, "log-format", "text", "Log format: text or json")
+	rootCmd.PersistentFlags().String("telemetry", "", "Usage telemetry: on or off (overrides BINTRAIL_TELEMETRY; DO_NOT_TRACK=1 overrides everything)")
 	// Usage telemetry control surface, same set as the core binary.
 	cli.AddTelemetryCommand(rootCmd)
 }
--- a/consoleapp/root.go
+++ b/consoleapp/root.go
@@ -24,6 +24,9 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/dbtrail/dbtrail/internal/cli"
+	"github.com/dbtrail/dbtrail/internal/telemetry"
+
 	"github.com/spf13/cobra"
 
 	"github.com/dbtrail/dbtrail/internal/observe"
@@ -62,6 +65,8 @@ yourself.`,
 func init() {
 	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "Log level: debug, info, warn, error")
 	rootCmd.PersistentFlags().StringVar(&logFormat, "log-format", "text", "Log format: text or json")
+	// Usage telemetry control surface, same set as the core binary.
+	cli.AddTelemetryCommand(rootCmd)
 }
 
 // Main configures build metadata and runs the bintrail-console root command,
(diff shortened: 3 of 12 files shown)


nethalo:
Two independent reviews (general + silent-failure). No privacy or wire defects found — the allowlist, the no-credential path, `normalizeClass`/`sanitizeCommand` coercion, and the suppress-only nature of the CI/no-home/no-endpoint gates all traced clean. One real correctness bug, now fixed.

**`drain` double-sent batches.** `os.Rename` preserves mtime, so a freshly claimed spool file still carries the time of its last *append*. For every prior-day file — the normal drain target — that is hours old, so the 5-minute abandoned-claim guard never engaged: a second drainer would adopt a claim still in flight and re-send it. The adoption branch also worked in place rather than re-claiming, so it wasn't exclusive either.

Claim age now comes from a stamp in the claim's own filename, and adoption re-renames. Measured: against the previous implementation `TestConcurrentDrainSendsEachBatchOnce` delivers **15 lines where 5 were written**; with the fix, exactly 5.

**The tests covering this passed for the wrong reason** — both built claims with a fresh mtime, which a real claim can never have. Rewritten to construct claims the way `drain` does; verified they fail against the old implementation and pass against the new one.

**`telemetry off` stranded events.** `drain` only runs while telemetry is *enabled*, and the age-out sweep lives inside it — so anything spooled before the operator opted out sat on disk indefinitely while the command said "nothing will be sent". `off` now discards them, and reports a failure to do so instead of swallowing it.

Also addressed: `BINTRAIL_TELEMETRY_DEBUG` (an on-by-default subsystem that swallows every error otherwise gives nobody a way to tell healthy from silently broken); registered the `--telemetry` flag the help text already promised; corrected the package doc, which claimed no identifier is ever written to disk while `run_id` is in fact spooled — the substantive guarantee (no *persistent install* identifier) holds, but an inaccurate comment is a defect when the comments are the audit surface; JSON `status` no longer emits a bare relative spool path when there is no home directory.

Deferred to #1060 with rationale, not dropped: daemons call `Init` once and live for months, so beacons would spool and age out unsent — daemon wiring needs its own periodic drain.

PR #1066, opened 2026-07-21
telemetry: record every command from the three root hooks

### LINKED ISSUE #1059: wire telemetry into the three binaries (cliapp, consoleapp, bintrail-pg); exclude MCP + console JS

nethalo:
closes #1059

Part of the telemetry epic #1054.

> **Stacked on #1064** — merge that first. Until it lands, this PR's diff also shows #1064's commits; it shrinks to just the wiring once #1064 is in. Branched off #1064's branch but targeting `main`, so merging #1064 with branch deletion will not close this PR.

## What this does

Wires the engine into `cliapp`, `consoleapp` and `cmd/bintrail-pg`. Cobra funnels every subcommand through the root's `PersistentPreRunE`, so this is the **only** instrumentation point — a new command needs no telemetry code at all.

`TelemetryHook` holds the shared plumbing so the recover/classify logic exists once instead of three times. Each binary gains two lines: `tel.Start(cmd)` in its hook, and `tel.Execute(rootCmd)` in place of `rootCmd.Execute()`.

- **Panics** are recorded as `error_class=internal` and re-raised, so the user-visible crash is byte-for-byte unchanged. The panic *value* is never recorded — panic messages routinely carry DSNs, table names and paths.
- **Command names** become hyphenated paths (`archive reconcile` → `archive-reconcile`), built from cobra command names, which are compile-time constants and can never carry an argument or flag value.
- **The `telemetry` subtree is exempt.** Initialising there would drain and deliver the spooled backlog moments before `telemetry off` discards it — an operator opting out would watch the tool phone home on its way out.
- **MCP server untouched**: `cmd/bintrail-mcp` has no root hook and is not wired. The structural guard for that is #1061.

## A defect this surfaced

Delivery only becomes reachable once something calls `Init`, and measured against a local endpoint it did not work: a `status` that failed in **2 ms delivered nothing** across repeated runs. The process exits microseconds after the drain goroutine starts, killing it mid-POST and leaving a claim the next fast run cannot adopt either. Telemetry would have been permanently dead for anyone whose commands are quick — while appearing healthy.

Fix: `Client.Shutdown`, a bounded 250 ms wait for an in-flight drain, plus claim reclamation dropped from 5 minutes to 1 (still 30× the drain deadline).

Re-measured with a local receiver:

| | result |
|---|---|
| run 1 (2 ms command) | event spooled, **nothing sent** — no network on the request path |
| run 2 | delivers run 1's event, total **28 ms** (~5 ms of wait) |
| run 3 (no backlog) | **23 ms** — a drain over an empty spool returns immediately, so no cost |

Received headers on the wire were `accept-encoding, content-length, content-type, host, user-agent` — the no-credential guarantee confirmed against a real listener, not just a unit test.

## Test plan

- [x] `go test ./... -count=1` — full suite passes; `go vet` and `gofmt` clean
- [x] New tests: `commandPath` including nested subcommands, the telemetry-subtree exemption, error passthrough, and that a panic is re-raised with its value intact (also exercising the nil-span path)
- [x] End-to-end against a local HTTP receiver: spool-then-deliver across runs, timings above, headers inspected
- [x] Real classification verified on a real failure: a refused index connection recorded `error_class=db_connection`

### DIFF
--- a/cliapp/root.go
+++ b/cliapp/root.go
@@ -20,6 +20,10 @@ var (
 	logFormat string
 )
 
+// tel records one usage event per invocation. Wired at the root so every
+// command — including ones added later — is covered without touching them.
+var tel cli.TelemetryHook
+
 // Build metadata, set by Main from the caller's -ldflags-injected values.
 // Package-level because commands (e.g. the agent's hello frame) read them
 // at runtime.
@@ -38,6 +42,7 @@ capabilities. The index is self-contained — recovery does not depend on
 binlog files still existing on disk.`,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 		observe.Setup(os.Stderr, logFormat, logLevel)
+		tel.Start(cmd)
 		return nil
 	},
 	SilenceErrors: true, // we handle error output ourselves in main()
@@ -80,7 +85,7 @@ func Main(version, commitSHA, buildDate string) int {
 	rootCmd.Version = fmt.Sprintf("%s (commit %s, built %s)", Version, CommitSHA, BuildDate)
 	telemetry.SetVersion(version)
 
-	err := rootCmd.Execute()
+	err := tel.Execute(rootCmd)
 	if err == nil {
 		return 0
 	}
--- a/cmd/bintrail-pg/main.go
+++ b/cmd/bintrail-pg/main.go
@@ -29,6 +29,10 @@ var (
 	logFormat string
 )
 
+// tel records one usage event per invocation, wired at the root like the core
+// binary's.
+var tel cli.TelemetryHook
+
 // Build-time variables injected via -ldflags. These are the SAME names the core
 // bintrail binary uses (main.Version/CommitSHA/BuildDate), so the Makefile's
 // BINTRAIL_LDFLAGS applies to this binary verbatim — exactly as bintrail-console
@@ -53,6 +57,7 @@ before-image (and de-TOASTed unchanged values) is present in the WAL. See
 'bintrail-pg stream --help'.`,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 		observe.Setup(os.Stderr, logFormat, logLevel)
+		tel.Start(cmd)
 		return nil
 	},
 	SilenceErrors: true, // we handle error output ourselves in main()
@@ -80,7 +85,7 @@ func init() {
 }
 
 func main() {
-	err := rootCmd.Execute()
+	err := tel.Execute(rootCmd)
 	if err == nil {
 		return
 	}
--- a/consoleapp/root.go
+++ b/consoleapp/root.go
@@ -43,6 +43,10 @@ var (
 // the running build (rootCmd.Version carries the human-formatted variant).
 var appVersion = "dev"
 
+// tel records one usage event per invocation, wired at the root like the core
+// binary's.
+var tel cli.TelemetryHook
+
 var rootCmd = &cobra.Command{
 	Use:   "bintrail-console",
 	Short: "Read-only web console over the Bintrail binlog index",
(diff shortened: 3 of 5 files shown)


nethalo:
Reviewed. No behavioural drift in any of the three binaries — the hook returns `root.Execute()`'s error unmodified, so cliapp's agent exit codes (64/65), the JSON error shaping, and bintrail-pg's `os.Exit(1)` all still see exactly what they saw before. `Shutdown` was checked against nil/disabled/recovered clients and double-calls; `claimReclaimAfter = 1min` was confirmed safe against the 2s drain deadline (one shared context across the whole loop, plus a 2s client timeout).

One real defect found and fixed: **shell completion was instrumented.** Cobra's `__complete` is an ordinary child of root, so the root hook fired for it — every TAB press would spool an event, attempt a POST, and pay up to 250 ms on a keystroke. Worse for the data than for the latency: `sanitizeCommand` rejects underscores, so `__complete` collapses into `"other"`, which would have been the largest bucket in the whole dataset on any developer's machine, indistinguishable from a genuine unknown command.

The exemption now covers `__complete`/`__completeNoDesc`/`completion`/`help` alongside the `telemetry` subtree, and is decided by walking the command chain rather than string-matching the rendered path. Verified on the real binary: two TAB completions plus a `help` record nothing; a real command records exactly one event.

Also corrected two comments that claimed more than the code delivers — the single-hook CI guard does not exist yet (it is #1061, and the comment now says so), and the panic re-raise preserves value/exit status/frames but is not literally byte-for-byte, since Go marks the trace `[recovered, repanicked]`.

PR #1067, opened 2026-07-21
telemetry: keep beacons alive in long-running daemons

### LINKED ISSUE #1060: daemon telemetry beacons (stream, agent, console watch): ≤1/UTC day, off-thread, startup notice line

nethalo:
closes #1060

Part of the telemetry epic #1054. Builds on #1064/#1066 (already merged).

## The problem this fixes

`Init`'s drain runs **exactly once**, at startup. A daemon lives for months, so its beacons would append to a spool that nothing ever delivers, and then age out after `maxSpoolAge`. The liveness signal — the entire point of a beacon — would silently never work, while looking healthy.

## What it does

`Client.RunDaemon(ctx, name)` is the loop: beacon + drain on its own goroutine, never on the replication, heartbeat or reconnect paths. Wired into `stream`, `agent`, `watch` and `bintrail-pg stream`; `up` is covered for free because it delegates to `runStream`. Each site passes `cmd.Name()`, so a beacon's `command` matches what that command's own event would have said.

**The first beacon comes after a full tick, not at startup.** The per-day cap is process-local, so a daemon that beaconed on start would emit one per restart, bounded only by the spool cap. Waiting an hour means a crash loop never beacons at all — and "alive for at least an hour" is the more honest liveness signal anyway.

**Daemons get one disclosure line at startup**, since they are not interactive and never see the first-run notice. It carries the cloned-image warning, because an image-baked opt-in travelling with the AMI is the realistic way a host ends up reporting without anyone there having chosen it.

## Two test-side defects found while writing this

- `spooledLines` treated a never-created spool directory as fatal instead of as zero events. That `Fatalf` then fired cleanup while the daemon goroutine was still reading `daemonTick` — a data race that only `-race` reports. Both fixed.
- The first version of the crash-loop test asserted on the spool, which is wrong: each tick beacons **and** drains, so the file is gone moments later either way. It now counts what an `httptest` endpoint actually receives.

## Test plan

- [x] `go test ./... -race -count=1` — full suite passes (CI's flags); `go vet`, `gofmt` clean
- [x] 10 consecutive `-race` runs of `internal/telemetry`
- [x] **Verified the crash-loop test fails against a beacon-at-startup loop** — `delivered 1 events before the first tick` — so it is a real guard, not decoration
- [x] Covered: returns on ctx cancel (a hanging loop would hang daemon shutdown), no beacon before the first tick, exactly one per UTC day across repeated ticks, delivery actually reaches the endpoint, beacons carry no `run_id`, and the loop is inert (returns at once) on a disabled or nil client

### DIFF
--- a/cliapp/agent.go
+++ b/cliapp/agent.go
@@ -289,6 +289,10 @@ func runAgent(cmd *cobra.Command, args []string) error {
 	ctx, cancel := context.WithCancel(cmd.Context())
 	defer cancel()
 
+	// See runStream: a long-lived process needs its own drain, off the
+	// heartbeat and reconnect paths.
+	go tel.Client().RunDaemon(ctx, cmd.Name())
+
 	var flushState *flushPipelineState
 
 	if byosMode {
--- a/cliapp/stream.go
+++ b/cliapp/stream.go
@@ -145,6 +145,12 @@ func runStream(cmd *cobra.Command, args []string) error {
 	ctx, cancel := context.WithCancel(cmd.Context())
 	defer cancel()
 
+	// Usage telemetry for a process that may live for months: Init's drain runs
+	// once, at startup, so without this loop a daemon's beacons would spool and
+	// age out undelivered. Own goroutine — never on the replication path. Also
+	// covers `up`, which delegates here.
+	go tel.Client().RunDaemon(ctx, cmd.Name())
+
 	// Graceful shutdown on SIGINT/SIGTERM: cancel the context so streamrun.One
 	// flushes its batch and writes a final checkpoint. (Installed before
 	// startup rather than after StartSync as it historically was — a signal
--- a/cmd/bintrail-pg/stream.go
+++ b/cmd/bintrail-pg/stream.go
@@ -199,6 +199,11 @@ func runPGStream(cmd *cobra.Command, args []string) error {
 	ctx, cancel := context.WithCancel(cmd.Context())
 	defer cancel()
 
+	// Usage telemetry for a months-lived process: Init's drain runs once, so
+	// without this loop a daemon's beacons would never be delivered. Own
+	// goroutine — never on the replication path.
+	go tel.Client().RunDaemon(ctx, cmd.Name())
+
 	// Graceful shutdown on SIGINT/SIGTERM: cancel the context so pgstreamrun.One
 	// flushes its batch and writes a final checkpoint before returning.
 	sigCh := make(chan os.Signal, 1)
--- a/consoleapp/watch.go
+++ b/consoleapp/watch.go
@@ -390,6 +390,10 @@ func runUpConsoleOnly(cmd *cobra.Command) error {
 	ctx, stop := signal.NotifyContext(cmd.Context(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
+	// Usage telemetry for a months-lived process: Init's drain runs once, so
+	// without this loop a daemon's beacons would spool and age out undelivered.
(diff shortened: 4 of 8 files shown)


nethalo:
Reviewed. The constraints that matter all hold: every one of the five call sites derives a context that is cancelled on return, nothing ever joins the loop goroutine (so it cannot delay shutdown or alter an exit code), `up` reaches the `go` statement exactly once, and no beacon or POST touches the replication, heartbeat or reconnect paths.

One real defect, and it was the kind worth catching: **`TestRunDaemonDoesNotBeaconBeforeFirstTick` was flaky.** It budgeted fixed sleeps against a 400 ms tick, leaving the loop ~280 ms to tick, spool, POST, and have the handler count it — reproduced at ~5 failures in 36 iterations across 6 concurrent test processes. That is exactly the intermittent CI redness that trains people to hit re-run instead of reading.

Split into two tests with the margin on the safe side. The crash-loop guard uses a 30 s tick and observes for 250 ms — contention can only make the loop *late*, never early, so a wide margin cannot fail falsely. The per-day-cap test polls to a deadline for the first delivery rather than budgeting for it, then waits out many ticks, where fewer ticks firing only makes the assertion more true. Re-ran the review's exact contention scenario: **0 failures**, and the guard still fails against a beacon-at-startup loop.

Also took the review's point on the log level: the daemon disclosure moves from INFO to **WARN**. It is the only disclosure a daemon ever makes, and the host that most needs it — one cloned from an image, where nobody present chose this — is also the one most likely running `--log-level warn` across a fleet. An INFO line would be silenced exactly where it matters most. It stays a structured slog record so it cannot corrupt a JSON log stream.

On scope: `bintrail-console serve` and `bintrail shim` are also months-lived and deliberately not included here. They are not broken — their command event still delivers on a later run; they simply do not beacon. Adding `shim` in particular needs a way to reach the client from `internal/cli`, which has no hook of its own, so I would rather do it deliberately than bolt on a package-level singleton at the end of this stack. Noted on #1054.

PR #1080, opened 2026-07-23
stream: record continuity loss when --reset skips binlog history

### LINKED ISSUE #1079: stream: --reset skips binlog history without recording continuity loss (and erases prior gap_lost history)
## Summary

`bintrail stream --reset` discards the saved checkpoint and restarts from the source's current binlog position without leaving any durable record of the binlog range it skipped. The events in that range are permanently lost to the index, yet `stream_state.gap_lost_at` stays NULL, `status` reports continuity `ok`, and `status --fail-on-gap` exits 0. The unfillable-gap auto-advance path stamps the loss (`persistGapAutoAdvance`, #402) precisely so a skip can never masquerade as health — the reset path bypasses all of it.

Worse: because reset is implemented as `DELETE FROM stream_state WHERE id = 1`, it also **destroys any previously recorded `gap_lost_at` / `gap_lost_detail`**, erasing loss history that #402 went out of its way to persist durably.

## Field observation that motivated this

On a long-running position-mode deployment (MySQL 8.0 source, `gtid_mode=OFF`), the index has a clean ~10-hour hole: continuous capture up to `T`, nothing until `T+10h`, then continuous capture resuming at the source's then-current position — mid-file, same binlog file spanning the hole's end. At resume time the binlog containing the window was still available on the source (7-day `binlog_expire_logs_seconds`, no server restart in weeks), i.e. the gap was fillable, so a checkpoint resume or the auto-advance machinery cannot explain the jump — only a checkpoint discard can.

Nothing flagged it: `gap_lost_at` NULL, continuity `ok`, console healthy. The loss surfaced **8 days later** — after the binlog had expired and the window had become unrecoverable — and only because a baseline-anchored `verify` between two snapshots reported row drift (deletes and inserts from inside the hole, missing from the change log). The exact operator action that triggered the discard is no longer provable (container logs rotated away), but the code path below produces this observable signature exactly.

## Code path

`internal/streamrun/streamrun.go`:

- Reset branch (step 6): `DELETE FROM stream_state WHERE id = 1`, `saved = nil`, one `slog.Warn` — the only trace of the discarded position is a log line.
- With `saved == nil`, the gap-detection block (step 6b) is skipped entirely (`if saved != nil`), so neither `detectPositionGap` nor the `persistGapAutoAdvance` stamp-then-advance machinery ever runs.
- `resolveStartWithAutoDiscoverForFlavor` then auto-discovers the source's **current** position and streaming resumes there.

Contrast with the unfillable-gap path a few lines below, which stamps `gap_lost_at`/`gap_lost_detail` durably *before* advancing the checkpoint (ordering guaranteed by #402) and notifies the supervisor via `OnGapAutoAdvance`.

## Reproduction

1. `bintrail stream ...` (position mode) until a checkpoint is saved; stop the stream.
2. Write rows on the source.
3. `bintrail stream --reset ...` (same flags) — resumes at the current position.
4. The rows from step 2 are absent from `binlog_events`; `bintrail status --fail-on-gap` exits 0 and Continuity shows `ok`. No `EVENTS PERMANENTLY LOST` banner ever appears.

## Expected

Any forward jump over an unread binlog range — operator-initiated or not — should leave the same durable trail the auto-advance path leaves:

1. When `--reset` finds an existing checkpoint, stamp `gap_lost_at` + `gap_lost_detail` (old checkpoint file/pos or GTID set → "checkpoint discarded via --reset; restarting from current position") **before** clearing state, mirroring the stamp-then-advance ordering of #402.
2. Carry the stamp into the fresh `stream_state` row (or stop deleting the row and clear only the position columns), so reset can no longer erase prior loss history.
3. `status` then reports `gap_lost` with the reset detail, `--fail-on-gap` exits non-zero, and the console badge reflects the loss — until an operator-visible acknowledgment, same lifecycle as an auto-advance loss.

A one-line docs touch-up on `--reset` ("permanently skips all events between the discarded checkpoint and the current position; the skipped range is recorded as lost") would follow from the fix.

## Notes

- Position mode with source GTIDs disabled is exactly the configuration where nothing else can catch this after the fact — GTID-mode executed-set comparison would at least surface the divergence on later restarts.
- Related: #402 (stamp-before-advance ordering for auto-advance), #649 (continuity verdict in `status`).
- Credit where due: the baseline-anchored `verify` (snapshot→snapshot replay) was the only mechanism that detected the loss, exactly as designed — but by the time a scheduled verify caught it, the source binlogs had expired. Recording the loss at reset time closes that window.

nethalo:
closes #1079

## Summary
- `--reset` no longer `DELETE`s the `stream_state` row up front. The old checkpoint is kept aside; the discard is persisted only once the new start position is resolved, so a startup failure in between leaves the old checkpoint (and any prior `gap_lost` record) intact.
- A reset that actually jumps over binlog history is durably recorded like an unfillable-gap auto-advance: `gap_lost_at`/`gap_lost_detail` stamped **first**, then the fresh checkpoint (`persistGapAutoAdvance`, same #402 stamp-then-advance invariant), plus the `OnGapAutoAdvance` supervisor hook so the console badge reflects the loss. `status` now shows the `EVENTS PERMANENTLY LOST` banner and `--fail-on-gap` exits non-zero after such a reset.
- A no-op reset (new start equals the discarded checkpoint — same file:pos, or structurally equal GTID sets) records nothing and just writes the fresh checkpoint. Cross-mode discards (position↔gtid) are conservatively treated as jumps.
- Reset can no longer erase prior loss history; monitor Stop remains the explicit acknowledgment path.
- `--reset` help text and `docs/streaming.md` now state that the skipped range is permanent and recorded.

## Test plan
- [x] New unit tests for `resetJumpDetail` (position noop/jump, GTID structural-equality noop/jump, cross-mode) — `internal/streamrun/reset_test.go`
- [x] `go test ./... -count=1` passes; `go test -race ./internal/streamrun/` passes
- [x] `go vet -tags integration ./...` clean

### DIFF
--- a/cliapp/stream.go
+++ b/cliapp/stream.go
@@ -36,6 +36,9 @@ Use --reset to clear the saved checkpoint and force a new start position:
   bintrail stream --reset --start-file mysql-bin.000500 ...
 
 Without --reset, the checkpoint always wins (idempotent behavior is preserved).
+Resetting past a saved checkpoint permanently skips every event between the
+discarded checkpoint and the new start position; the skipped range is durably
+recorded as lost (surfaced by 'bintrail status' and its --fail-on-gap flag).
 
 Gap detection: on restart, bintrail checks whether the source still has the
 binlogs needed to resume from the checkpoint. If binlogs have been purged, it
@@ -93,7 +96,7 @@ func init() {
 	streamCmd.Flags().StringVar(&strmSSLCert, "ssl-cert", "", "Path to client certificate file for mutual TLS")
 	streamCmd.Flags().StringVar(&strmSSLKey, "ssl-key", "", "Path to client private key file for mutual TLS")
 	streamCmd.Flags().StringVar(&strmFormat, "format", "text", "Output format: text or json")
-	streamCmd.Flags().BoolVar(&strmReset, "reset", false, "Clear saved checkpoint before starting (forces use of --start-file/--start-gtid)")
+	streamCmd.Flags().BoolVar(&strmReset, "reset", false, "Clear saved checkpoint before starting (forces use of --start-file/--start-gtid); the skipped range is recorded as permanently lost")
 	streamCmd.Flags().BoolVar(&strmNoGapFill, "no-gap-fill", false, "Refuse to start if an unfillable binlog gap is detected (instead of auto-advancing past purged data)")
 	streamCmd.Flags().IntVar(&strmGapTimeout, "gap-timeout", 30, "Timeout in seconds for the one-shot gap-detection queries run at startup (SHOW BINARY LOGS plus @@gtid_purged/@@gtid_executed on MySQL or BINLOG_GTID_POS/@@gtid_binlog_pos on MariaDB); raise on managed servers with many binlog files")
 	streamCmd.Flags().DurationVar(&indexer.WriteTimeout, "write-timeout", indexer.DefaultWriteTimeout, "Deadline for each index write (batch INSERT, checkpoint, digest lookup). A mid-statement network stall surfaces as an error within this window instead of freezing the stream on kernel TCP retransmission (~13-16 min). Raise for very large batches over a slow link")
--- a/cliapp/stream.go
+++ b/cliapp/stream.go
@@ -36,8 +36,9 @@ Use --reset to clear the saved checkpoint and force a new start position:
   bintrail stream --reset --start-file mysql-bin.000500 ...
 
 Without --reset, the checkpoint always wins (idempotent behavior is preserved).
-Resetting past a saved checkpoint permanently skips every event between the
-discarded checkpoint and the new start position; the skipped range is durably
+Resetting to any position other than the saved checkpoint — later or earlier;
+direction is not inferred — permanently skips every event between the discarded
+checkpoint and the new start position, and the skipped range is durably
 recorded as lost (surfaced by 'bintrail status' and its --fail-on-gap flag).
 
 Gap detection: on restart, bintrail checks whether the source still has the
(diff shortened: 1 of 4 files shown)


nethalo:
Multi-agent review round (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer) applied:

**Fixed**
- **Cross-mode reset persisted a stale `mode`** (regression in this PR): `saveCheckpoint`'s ON DUPLICATE KEY UPDATE never updated `mode`; with the row now surviving `--reset`, a position↔gtid switch would silently revert on the next restart. Added `mode = VALUES(mode)` + sqlmock pin.
- **MariaDB false-loss stamping**: `resetJumpDetail` compared GTID sets with the MySQL-only parser, so every MariaDB gtid→gtid reset — including a byte-identical set — was classified as a jump and stamped `gap_lost`. Now flavor-aware (`mariadbGTIDSetsEqual`), with MariaDB noop/jump tests.
- Reset counters restart at zero (the carried value was overwritten by the first checkpoint tick anyway).
- User output + supervisor hook now fire only after the durable record lands; persistence extracted to `persistResetDiscard` with sqlmock tests on both branches.
- Comment/doc corrections: a jump **replaces** a prior unacknowledged loss record's detail; the no-op check is coordinates-only and direction-blind (a rewind also records a loss).

**Noted, deliberately out of scope (pre-existing)**
- `persistGapAutoAdvance`'s stamp is a bare UPDATE with no RowsAffected check (a concurrently-deleted row loses the stamp silently); the PG sibling uses INSERT…ODKU. Pre-dates this PR on the auto-advance path.
- `bintrail-pg reset` still hard-DELETEs `stream_state`, erasing loss records — same class as #1079, PG binary.

Issue #1074, opened 2026-07-23
console: per-session authorization seam — sessions carry an access policy (action tier + data profile) enforced at the API layer


nethalo:
## Problem

Every authenticated console session is equivalent. The password login, the static `--token`, and an identity admitted through an installed `ext.ConsoleAuthProvider` all mint the same in-memory session, and every session can do everything the server allows: read all events, generate recovery SQL, CRUD the server registry, change settings, trigger baselines. There is no way for an auth backend to admit an identity with *less* power than the operator who started the daemon.

`ext.ConsoleSessionIssuer` takes only a display-string identity — the seam has no channel through which a provider could scope the session it is minting.

## Proposal

Sessions optionally carry an **access policy** with two orthogonal dimensions:

1. **Role** — a fixed role resolving to a set of **permission strings**, controlling which endpoints the session may call.
2. **Data profile** — the name of an existing RBAC profile (the `profiles`/`flags`/`access_rules` tables managed by `bintrail flag|profile|access`), scoping what data the session sees. Enforcement of this dimension is #1075; this issue only carries the name on the session and exposes it.

A session with **no policy behaves exactly as today** (full access, no profile). The OSS build never attaches a policy on its own, so OSS behavior is unchanged — this is a seam.

## Permission model

Not an ordered tier ladder — a `permission` is a `verb:noun` string, each role maps to a precomputed permission set (inheritance baked in at definition), and each route requires exactly one permission. This keeps the door open for per-deployment grant/revoke overrides later without redesigning enforcement.

Permissions (initial set): `status:read`, `query:execute`, `recover:execute`, `reconstruct:execute`, `baseline:create`, `servers:read`, `servers:write`, `servers:delete`, `settings:read`, `extview:read`.

Roles (permission sets precomputed, each a superset of the previous):

| Role | Permissions |
|---|---|
| `viewer` | `status:read`, `servers:read` — health/status surface only, **no row data** |
| `analyst` | viewer + `query:execute`, `reconstruct:execute`, `extview:read` — reads data, changes nothing |
| `operator` | analyst + `recover:execute`, `baseline:create` — incident response |
| `admin` | operator + `servers:write`, `servers:delete`, `settings:read` — everything (a policy-less session is equivalent to admin) |

## Scope

- **Session model**: attach an optional policy at mint time; store it with the in-memory session (never persisted, like the session itself).
[...]

Issue #699, opened 2026-07-02
Forensics: capture original SQL statement per row event (binlog_rows_query_log_events) + STATEMENT_DIGEST fingerprinting


nethalo:
## Motivation

The index already answers **who** touched a row (`connection_id`, extracted from the QUERY(BEGIN) event). It cannot answer **what statement** did it. Capturing the original SQL per row event closes that gap and is a natural extension of the forensics surface:

- "which exact query deleted these rows" next to the before-image, in `query`/console/MCP results
- query-shape analytics: normalize the captured statement via `STATEMENT_DIGEST()` to a stable hash, enabling "top query patterns mutating this table" and "a new DELETE pattern appeared" (the funnel into recovery)

Prior art: [avaitla/mysqlbinlog2parquet](https://github.com/avaitla/mysqlbinlog2parquet) ships both (its `query`, `query_fingerprint`, `query_hash` columns) and reports them as its most-used forensic columns.

## How it works

MySQL emits a `RowsQueryEvent` (the literal statement text) immediately before the row events of a statement when `binlog_rows_query_log_events=ON` (default OFF, dynamic, no restart). go-mysql already parses it — `replication/row_event.go` `RowsQueryEvent`, wired in `parser.go` — we currently ignore it.

MariaDB's sibling is `binlog_annotate_row_events` → `MARIADB_ANNOTATE_ROWS_EVENT`, also already parsed by go-mysql (`parser.go`), and the dump flag `BINLOG_SEND_ANNOTATE_ROWS_EVENT` must be set on the syncer for replication mode.

## Sketch

1. **Parser** (`internal/parser/parser.go`): capture the pending statement text when a `RowsQueryEvent` / `MariadbAnnotateRowsEvent` arrives; attach it to the row events that follow, scoped to the statement (cleared on next query/statement boundary) — same pattern as `connection_id` from QUERY(BEGIN), but per-statement not per-transaction.
2. **Schema**: new nullable column on `binlog_events` (e.g. `query_text TEXT DEFAULT NULL` + optional `query_hash CHAR(64)`), added via `indexer.EnsureSchema` (idempotent, like `connection_id`). NULL when the variable is OFF or for pre-existing events.
3. **Digest**: compute at index time against the **index** connection (`SELECT STATEMENT_DIGEST(?)`) — index contract floor is MySQL 8.0+, so no dependency on the source; batch/cache by statement text to avoid a round-trip per event.
[...]

PR #1088, opened 2026-07-23
console: enforce data profiles per session at request time

### LINKED ISSUE #1075: console: enforce data profiles per session at request time (per-server rule loading, redaction, archive and reconstruct gates)
## Problem

The console's RBAC profile support is process-global: `--profile` loads deny/redact rules **once at startup**, from the CLI-supplied index only, applies them to every session identically, and forces `--no-archive` for the whole process (archives don't enforce redaction). With sessions that carry their own access policy (see the session-authorization seam issue), a per-session data profile cannot be honored by the current plumbing at all:

- rules live in **each server's** index (`profiles`/`flags`/`access_rules`), and the selected server changes per request (`X-Bintrail-Server`);
- the no-archive and reconstruct gates live on the per-server `bundle`, not on the session;
- `ext.ConsoleQueryContext.Fetch` applies "whatever RBAC redaction the console was started with" — process-wide.

## Proposal

When a session carries a data-profile name, resolve and enforce that profile **at request time against the selected server's index**:

- **Rule loading**: `query.ProfileExists` + `query.LoadProfileRules` against the selected server, cached per `(server, profile)` with a short TTL. A profile name that does not exist on the selected server → 403 with an actionable message (mirror of the fail-loud startup check — never "enforce nothing" on a typo).
- **Read paths**: apply `DenyTables`/`RedactColumns`/`ProfileActive` on the events endpoint, the recover endpoint, and `ConsoleQueryContext.Fetch` for extension views.
- **Archive gate**: a profiled session queries with archives disabled (per-session `no-archive`), for the same reason the startup flag forces it: Parquet archive reads bypass redaction.
- **Reconstruct/baseline gate**: a profiled session gets 403 on `/api/reconstruct` and baseline listings — baseline reads bypass redaction (same invariant that already gates reconstruct under a startup profile).
- **Composition with the startup `--profile`**: both apply — union of deny tables and redact columns. The startup profile is the floor; a session profile can only narrow further.
- **Exports**: client-side JSON/CSV export already operates on the redacted `eventDTO`s on screen — verify with a test, no change expected.

## Notes

- `query_text`/`query_hash` withholding under `ProfileActive` (#699 semantics) must fire for session profiles exactly as for CLI profiles.
- Sessions with no profile keep today's behavior bit-for-bit; the OSS build attaches no policies, so this is inert until a provider does.

Depends on #1074 (policy carried on the session).

nethalo:
closes #1075

## Summary
A session may carry a data-profile name (`ext.AccessPolicy.Profile`, from #1074). When it does, the console resolves that profile **against the selected server's index at request time** and enforces it. The OSS build attaches no policy, so every session is unprofiled and behavior is unchanged.

- **Redaction on data reads** (`/api/events`, `/api/recover`): the session profile's deny/redact rules are resolved (`query.ProfileExists` + `query.LoadProfileRules`, cached per `(server, profile)` with a 30s TTL) and **unioned** onto the startup floor — the startup `--profile` is the floor, a session profile only narrows further. `ProfileActive` is forced so `query_text`/`query_hash` stay withheld (#699). A profile that does not exist on the selected server is a **403** (never "enforce nothing" on a typo). Archives are forced **off** per request (Parquet reads bypass redaction).
- **Raw/baseline-data surfaces refused (403)** for a profiled session: reconstruct, baseline listings, recover-cascade, and verify — their reads bypass redaction or their synthesis can't honor it. Extension-view data routes are refused by `rbacViewGuard` too; the reconstruct/cascade/verify **capabilities** go false so the SPA hides them.
- **Two per-request helpers** — `rbacActiveFor(r)` / `profileActiveFor(r)` — fold the session profile into the process-global `rbacActive()`/`profileActive` signals; every gate site calls them, so a profiled session refuses/hides exactly as a startup profile does. Inert in OSS.

## Composition with the startup `--profile`
Both apply: the union of deny tables and redact columns. The startup profile stays the floor; a session profile can only add restrictions.

## Out of scope (per the epic)
The console `/mcp` endpoint and the flashback port authenticate by token (no session policy) — untouched.

## OSS behavior preserved
No session ever carries a profile in the OSS build, so `sessionRestricted` is always false and every helper/gate is inert — byte-for-byte unchanged. `TestSessionRestricted`/`TestRbacActiveForFoldsSession` pin the no-profile path.

## Test plan
- [x] `go build ./...` clean; `go vet -tags integration ./internal/console/` clean
- [x] Unit: helpers (restricted/rbacActiveFor/profileActiveFor), `applySessionProfile` no-op without a profile, `writeSessionProfileError` (403/500), recover-cascade 403 for a profiled session, capabilities reconstruct/cascade false
- [x] **Integration (run against real MySQL 8.0 — all pass):** per-session redaction (denied table hidden, `ssn` column nulled, email intact) with a policy-less credential on the SAME server still seeing raw data (proves per-session, not process-global); nonexistent profile → 403; reconstruct + baselines → 403
- [x] Existing `TestIntegrationProfileRBAC` / reconstruct / verify integration tests still pass (no regression)

### DIFF
--- a/internal/console/api.go
+++ b/internal/console/api.go
@@ -201,7 +201,12 @@ func (s *Server) handleEvents(w http.ResponseWriter, r *http.Request) {
 		writeJSONError(w, http.StatusBadRequest, err.Error())
 		return
 	}
-	rows, plan, err := s.fetch(r.Context(), b, opts)
+	opts, err = s.applySessionProfile(r.Context(), r, b, opts)
+	if err != nil {
+		writeSessionProfileError(w, err)
+		return
+	}
+	rows, plan, err := s.fetchRestricted(r.Context(), r, b, opts)
 	if err != nil {
 		writeFetchError(w, err)
 		return
@@ -244,6 +249,11 @@ func (s *Server) handleRecover(w http.ResponseWriter, r *http.Request) {
 		writeJSONError(w, http.StatusBadRequest, err.Error())
 		return
 	}
+	opts, err = s.applySessionProfile(r.Context(), r, b, opts)
+	if err != nil {
+		writeSessionProfileError(w, err)
+		return
+	}
 	// Refuse to generate an undo script for the entire index; a recovery must
 	// be scoped to at least one schema.
 	if opts.Schema == "" {
@@ -267,7 +277,7 @@ func (s *Server) handleRecover(w http.ResponseWriter, r *http.Request) {
 	// Coverage gaps come back in plan.GapHours and are surfaced as warnings
 	// below — the recover UI renders them, so an incomplete-coverage undo is
 	// flagged to the operator rather than silently presented as complete.
-	rows, plan, err := s.fetch(r.Context(), b, opts)
+	rows, plan, err := s.fetchRestricted(r.Context(), r, b, opts)
 	if err != nil {
 		writeFetchError(w, err)
 		return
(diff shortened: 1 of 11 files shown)


nethalo:
Code review found one material item (data-redaction boundary): the per-`(server, profile)` rule cache wasn't invalidated on a server **DSN edit**, so a profiled session could enforce rules resolved against the *old* index against the *new* one for up to the 30s TTL (an under-redaction window). Fixed: `profileRuleCache.invalidate(id)` is now called beside `cm.evict(id)` on both DSN edit and delete (nil-safe; unit-tested). Amended into the branch.

PR #1109, opened 2026-07-25
verify: add --format json (#954)

### LINKED ISSUE #954: verify: no --format json — the one command positioned for cron/CI is the only read command with text-only output
internal/cli/verify.go:73-82 registers no `--format` flag; output is tabwriter text only (:346-370). docs/verify.md has an "Exit codes (for cron/CI)" section and recommends scheduling it. Every sibling read command (status/query/recover/reconstruct/archive reconcile/init) has `--format json`.

A scheduled consumer can only read the exit code — it can't tell which table mismatched or whether the run was all-inconclusive without scraping columns. Also blocks the console verify-lifecycle (#677), which needs a structured shape anyway.

Fix: add `--format text|json` emitting per-table `{schema, table, verdict, matched, total, reason}` + summary counts + continuity, reusing `cliutil.OutputJSON` and the `{"error":...}` stderr convention; design the struct so `/api/verify` (#677) can serialize the same type.

---

nethalo:
## Problem

`bintrail verify` registered no `--format` flag: output was tabwriter text only. `docs/verify.md` recommends scheduling it and documents exit codes, but a cron/CI consumer could read *only* the exit code — it could not tell **which** table mismatched, or whether the run was all-inconclusive, without scraping columns. Every sibling read command (`status`/`query`/`recover`/`reconstruct`/`archive reconcile`) already has `--format json`.

closes #954

## What changed

- **`--format text|json`** on `verify` (default `text`), validated with `cliutil.IsValidOutputFormat` and emitted with `cliutil.OutputJSON`. The `{"error":"…"}`-on-stderr convention comes free once the flag exists — both roots (`cliapp/root.go` and `cmd/bintrail-pg/main.go`, which each register `verifyCmd` via `cli.AddReadCommands`) look up a `format` flag by name in their `wantsJSON` shim.
- **The shape lives in `internal/verify/report.go`**, not the command file, as the issue asked: `verify.Report` / `TableReport` / `Summary` / `ExplainReport`, built by a **pure** `NewReport(mode, results)` — no IO, no flags, no stdout — so a future console `/api/verify` (#677) can serialize the identical type server-side. The CLI owns only the encoding.
- **Exit semantics are unchanged and now un-drift-able.** `Report.ExitError()` is the single source of the exit contract for *both* renderings: mismatch fails, hard error fails, all-inconclusive fails, one-baseline-only exits 0. The old text-path `switch` is gone; `writeVerifyText` renders from the same `Report`.
- **Nothing silently dropped.** The JSON carries the per-table `Anchor` the text table has no column for, the `Detail` (as `reason`), digests, row counts, summary counts, and a run-level `verdict` (`verified` / `mismatch` / `error` / `unproven` / `no_predecessor`). The two prose paths that bypassed the report also get a JSON form: the "only one baseline" message becomes `verdict: "no_predecessor"` + `message` (still exit 0), and a failed `--explain` drill-down becomes `explain[].unavailable` instead of prose that would corrupt the document.
- **`--explain` in JSON** marshals through `MismatchExplanation.ReportEntry()` rather than the internal type directly — `byKind` and `deferredSeen` are unexported, so a direct marshal would have silently dropped the per-kind overflow breakdown and the deferred-type caveat that the text `Write` prints.
- `docs/verify.md`: new "Machine-readable output (`--format json`)" section with a worked example and field semantics, plus the flag table row.

## Deviations from the issue text (deliberate)

- The issue sketched per-table `{schema, table, verdict, matched, total, reason}`. `matched`/`total` map to nothing verify computes per table, so the row is `{schema, table, status, source_rows, reconstruct_rows, source_digest, reconstruct_digest, anchor, reason}` — `status` matches the existing `verify.Status` values and the text `STATUS` column, `reason` keeps the issue's word for `Detail`. `verdict` exists at the **run** level, where it means something.
- The issue asked for "continuity". `verify` never reads `stream_state` and computes no continuity signal — the only "continuity" in its world is the `docs/verify.md` cross-reference to `bintrail status`. Asserting a continuity verdict verify did not check would be exactly the false assurance this command exists to prevent, so the report deliberately carries none; the docs now say so explicitly and point at `status --format json` (`continuity.status`). What the JSON *does* add over the text report is the per-table `anchor` — verify's own chain-anchoring signal, which the text table drops.

## Tests

- `internal/verify/report_test.go`: classification + sorting, unknown/zero status counted as **error** (never inconclusive) with the raw value preserved in `reason`, verdict/exit precedence, no-predecessor exits 0 and serializes `"tables":[]`, and `ReportEntry` carrying the overflow breakdown + deferred-type note.
- `internal/cli/verify_format_test.go`: JSON shape per verdict (match/mismatch/inconclusive, all-inconclusive, no-predecessor, explain entry), text output unchanged, invalid `--format` rejected before any DB connection, and the `format` flag registration the `wantsJSON` shim depends on.
- `internal/cli/verify_test.go`: the existing exit-code table now runs **under both formats**, pinning that `--format` is a rendering choice and never a semantics change.

`go build ./...`, `go test ./... -count=1 -race`, and `go vet -tags integration ./...` all pass.


nethalo:

**No critical findings.** Both headline priorities come out clean, verified path by path:

- **Exit semantics are unchanged on all seven paths** (format validation, zero baselines, one baseline, `len(results)==0`, the report, live mode, explain failures). `verdictOf` + `ExitError` (`internal/verify/report.go:204-231`) reproduce the old `switch` precedence-for-precedence and message-for-message; `normalizeStatus` (`:190-200`) preserves the `default: errored++` rule. Both prose bypasses behave identically in text and JSON.
- **No information loss in either direction.** All nine `TableResult` fields reach `TableReport`; JSON is a strict superset of the text columns. The only text-only artifact ("no per-row content differences found — the mismatch is in row count") is derivable from `total_differing_rows: 0`.
- Excluding continuity is correct — `verify` genuinely never reads `stream_state`.
- No `bintrail-pg` breakage: neither root has a persistent `--format`, `verify` never calls `BindCommandEnv`, and `EnvBindings` has no `format` term.

### Important

1. **`--format JSON` splits the JSON contract in half.** `internal/cli/verify.go:445-447` dispatches with `strings.EqualFold`, but both roots decide the error shape with an exact compare — `cliapp/root.go:148-150` and `cmd/bintrail-pg/main.go:110-112` do `f.Value.String() == "json"`.

   Input: `bintrail verify … --format JSON` on a mismatched table → JSON report on stdout, but `2 table(s) diverged from the source` as **bare text** on stderr instead of `{"error":…}`. It also makes `verify` the only command in the repo where `--format JSON` produces JSON at all — `status`/`query`/`recover`/`rotate`/`recover-cascade`/`telemetry` all compare exactly and silently render text.

   One-line fix restoring parity with the siblings *and* both shims: `return vfyFormat == "json"`.

2. **The design rationale is already falsified — #677 is closed and implemented.** `internal/verify/report.go:36-40` justifies the package placement as "so a future console /api/verify (#677) can serialize the identical shape", but `internal/console/verify_trigger.go:74-114` already defines `VerifyTableResult`/`VerifySummary`/`VerifyStatus`, and `consoleapp/verify.go:404-422` already re-implements the status→bucket classification including the same anti-false-assurance `default → Error` rule `normalizeStatus` now owns.

   The types themselves are clean and CLI-free (pure `NewReport`, no flags/IO/stdout) — the *premise* is stale and nothing here moves the console onto them. In-scope: correct the comment. Follow-up: `normalizeStatus` is unexported, which is the concrete blocker to `consoleapp` reusing it instead of keeping a third copy.

3. **`explain[]` ordering is nondeterministic while `tables[]` is sorted.** `internal/cli/verify.go:317` appends in `pairs` iteration order, and `pairs` is built by ranging a map (`internal/verify/verify_baseline.go:362`) and never sorted — note `unpaired` and `prevOnly` *are* sorted at `:400-401`, so this looks like an oversight in the original. With `--explain --format json` and ≥2 mismatched tables, `explain[0]` and `explain[1]` swap between identical runs. Text has the same nondeterminism, but text is read by a human; JSON is a document.

### Suggestions

- Make the "single exit decision" literally true: `writeVerifyText`'s `VerdictNoPredecessor` branch (`internal/cli/verify.go:466-469`) is unreachable, because `:176-177` prints the prose and returns directly. Correct today only by coincidence of `nil == ExitError()`.
- Text output *did* change for one case, contrary to the PR body: an unrecognized status now prints `error` + `unrecognized verify status "bogus": <detail>` instead of the raw value. An improvement — but `TestVerifyTextFormatUnchanged` never feeds a bogus status, so nothing pins it.
- The two machine-readable surfaces of the same command disagree on field names: CLI JSON emits `reason`, the console emits the same datum as `detail`, and the console `VerifySummary` has no `total`.
- Stale comment references to the deleted `printVerifyReport`: `consoleapp/verify_test.go:92`, `internal/cli/verify_baseline_integration_test.go:113`.
- `NewReport(mode, nil)` returns `VerdictUnproven`; both CLI callers guard `len(results)==0` first so it's unreachable today, but that's exactly the caller this type is being built for.

### Strengths

- Routing both renderings through one `ExitError()` and then running the *existing* exit table under both formats (`verify_test.go:34-52`) is what makes "JSON is a rendering choice, not a semantics change" checkable rather than asserted.
- `ReportEntry` catching the unexported `byKind`/`deferredSeen` avoids a genuine information-loss bug — a naive `json.Marshal(ex)` would have silently dropped the missing-row overflow breakdown from the machine-readable output. Pinned with a real cap overflow.
- `normalizeStatus` promoting unknown status to `error` *while preserving the raw value in `reason`* is better than the code it replaces.
- Hand-writing the decode-side struct in `verify_format_test.go:59-86` instead of round-tripping `verify.Report` is the right instinct — a field rename fails instead of silently passing.

PR #1110, opened 2026-07-25
recover: apply-side codegen — PG session_replication_role and opt-in AUTO_INCREMENT restore

### LINKED ISSUE #1003: recover: apply-side codegen — PG session_replication_role and optional AUTO_INCREMENT restore
## Background

Split out from #787 (closed: the apply-side caveats are now documented in `docs/query-and-recovery.md` under **Restore limitations (MySQL)**, and the script already pins `time_zone`/`sql_mode` via `internal/recovery/recovery.go:317`/`:330`). This issue tracks the remaining deferred **codegen** enhancements — the parts that go beyond documentation.

## Proposed enhancements

1. **PostgreSQL: suppress trigger re-fire on apply.** MySQL has no session toggle to suppress triggers during an apply (fundamental — documented), but PostgreSQL does. For the `bintrail-pg` reversal script, emit `SET LOCAL session_replication_role = replica;` in the preamble so the apply does not re-fire the target's triggers (which would produce new, double-applied side effects). Gate it so it only applies to the PG codegen path.

2. **Optional `AUTO_INCREMENT` restore step (MySQL).** Reverting an `INSERT` deletes the row but does not decrement the counter; reverting a `DELETE` re-inserts with an explicit id, which can bump it further — so the table ends row-identical but the *next* id differs. Optionally emit a commented-or-flagged `ALTER TABLE ... AUTO_INCREMENT = N` step (opt-in, since the correct `N` depends on operator intent) so operators who care can restore the counter deterministically.

## Scope / notes

- Home: `internal/recovery/recovery.go` (the script preamble at `recovery.go:278-296` and the per-flavor branches).
- Both are additive to the generated script; the documented caveats stay accurate either way.
- The MySQL trigger re-fire remains fundamental — this issue does not claim to fix it, only the PG-suppressible and AUTO_INCREMENT-reconstructible parts.

nethalo:
Two opt-in additions to the generated reversal script, both additive and both dialect-gated. The documented caveats stay accurate — **MySQL trigger re-fire remains fundamental and is NOT claimed fixed**.

closes #1003

## 1. `--suppress-triggers` (PostgreSQL dialect only)

Emits `SET LOCAL session_replication_role = replica;` immediately after `BEGIN`, so the apply does not re-fire the target's own triggers and double-apply side effects the original triggers already logged as their own (separately reversed) events.

**Transaction framing** (the issue asked to check this): the script is already `BEGIN`/`COMMIT`-wrapped by `GenerateSQLFromRows`, so `SET LOCAL` is the correct form — transaction-scoped, restored at `COMMIT`/`ROLLBACK`, never leaking into the applying session. It sits alongside the existing `SET LOCAL standard_conforming_strings = on`.

**Deviation from the issue text, deliberate: it is OPT-IN rather than unconditional.**
- *Privilege*: setting `session_replication_role` requires superuser (PostgreSQL ≤ 14) or an explicit `GRANT SET ON PARAMETER session_replication_role TO <role>` (15+). Emitting it unconditionally would break every `bintrail-pg recover` script applied by an ordinary role — a regression on a released default path, hit at the worst possible moment (mid-recovery).
- *Semantics*: `replica` also disables **`FOREIGN KEY` constraint triggers**, so referential integrity is not enforced while the reversal applies. That is a real behavior change the operator should choose, not inherit silently.

Both facts are stated inline in the emitted script (with the remedy), and the script's header `NOTE` switches to accurate text when suppression is on (it otherwise says triggers *will* fire).

## 2. `--restore-auto-increment` (MySQL dialect only)

Appends, per table the reversal writes, a `SELECT IFNULL(MAX(<col>),0)+1` / `ALTER TABLE ... AUTO_INCREMENT = <N>` pair.

- **After `COMMIT`**, not inside it: `ALTER TABLE` is DDL and implicitly commits in MySQL, so inside the transaction it would split the reversal's atomicity — the invariant `GenerateSQLFromRows` buffers the whole body to protect.
- **Commented out**, per the issue's "if it cannot be derived safely, emit it commented-out with the reasoning inline". `N` is not derivable from the index: the schema snapshot records data types, PKs and generated/identity flags but **not** which column is `AUTO_INCREMENT` (`is_generated` comes from `GENERATION_EXPRESSION`, not `EXTRA`), and the correct `N` depends on operator intent (reuse the ids the reversal freed vs. never reuse them). No column name is guessed either — a wrong guess would point `MAX()` at the wrong column and produce a plausible-but-wrong `N`.
- The reasoning, the placement rationale, and the InnoDB clamp behavior (a value at or below the id in use is ignored → a too-low `N` is a no-op, never data loss) are all inline in the block.

## Shared command surface

`internal/cli/recover.go` is shared by `bintrail` and `bintrail-pg`, so both flags appear on both binaries. Each is inert on the other dialect, and `runRecover` `slog.Warn`s when a flag is set against an index it cannot help — so an operator's mental model ("triggers won't fire") is never silently wrong.

## Sample output (MySQL, opted in)

```
COMMIT;

-- ─── AUTO_INCREMENT restore (--restore-auto-increment) ─────────────────────
-- ... reasoning ...
--
-- `db`.`orders`:
--   SELECT IFNULL(MAX(`<auto_increment_column>`), 0) + 1 FROM `db`.`orders`;
--   ALTER TABLE `db`.`orders` AUTO_INCREMENT = <N>;
```

## Tests

`internal/recovery/applycodegen_test.go`:
- PG preamble contains the `SET LOCAL` only when opted in, only on the PG dialect, and positioned **between** `BEGIN` and `COMMIT`; default PG script does not.
- MySQL never emits `session_replication_role`, flag or no flag.
- `AUTO_INCREMENT` block absent by default; present when opted in with one commented-out pair per written table, asserted **after** `COMMIT` by index, and every `ALTER TABLE` line asserted commented out.
- PG + `--restore-auto-increment` emits no block; an empty reversal emits no block.

`go build ./...`, `go test ./... -count=1 -race`, `go vet -tags integration ./...` all pass. The existing PG-preamble assertions (`internal/pgstreamrun`, `internal/console` integration tests) are unaffected because the default output is byte-identical.

## Docs

`docs/query-and-recovery.md`: both flags documented in the reversal-script "Key properties" list (including the opt-in rationale and the FK-trigger caveat), and the two "Restore limitations (MySQL)" bullets amended — the trigger bullet keeps its "fundamental limitation for MySQL" claim intact and now points at the PG flag; the `AUTO_INCREMENT` bullet points at the new step while keeping "the script still never picks `N` for you".

## Left out

- No env-var binding for either flag (`internal/cli/env.go` binds a deliberate subset; `--allow-gaps`/`--no-archive` are unbound too).
- No `is_auto_increment` column added to `schema_snapshots` — that would let the block name the column, but it is a schema migration + capture + `EnsureSchema` change well past this issue's scope.
- `recover-cascade` is untouched: it builds its own Generator and writes its own preamble/epilogue (`SET FOREIGN_KEY_CHECKS=0/1` around the generator's output), so neither switch applies there. It has the same counter-drift problem, but the issue scoped this to `recover`; wiring it through is a separate, trivial follow-up if wanted.
- MySQL trigger re-fire is untouched and still documented as fundamental.


nethalo:

**Nothing here blocks merge.** The dialect gating is genuinely airtight, the transaction framing is exactly as claimed, and the after-`COMMIT` placement is well reasoned. The findings are one real hole (whose worse half is pre-existing) plus accuracy problems in emitted text.

### Important

1. **Comment break-out: a newline in a table name — or a PK value — puts executable SQL inside the "commented-out" block.**

   `internal/recovery/recovery.go:469-471` writes identifiers with `fmt.Fprintf(w, "--   ALTER TABLE %s …")`. `QuoteName` escapes backticks but not newlines, and a `--` comment ends at the first newline. MySQL permits any BMP character except U+0000 in a backtick-quoted identifier, so a table named `orders\nDROP TABLE users; -- ` produces three uncommented `DROP TABLE users;` statements, each with the trailing junk commented away — after `COMMIT`, so they run standalone. The vector is not filtered upstream: `internal/parser/parser.go:603` passes `schema, table` straight from TABLE_MAP into the event.

   This is the one place where the comment prefix is the *entire* safety mechanism — but **this PR does not introduce the class**. `recovery.go:288` already does `fmt.Fprintf(&body, "-- [%d] reverse %s on %s.%s pk=%s at %s%s\n", …)` with raw schema, table **and PK values**, and `event.EscapePKValue` escapes only `\` and `|`. A newline inside a VARCHAR PK value is far more reachable than one in a table name, and it breaks out *inside* the `BEGIN`/`COMMIT`. That ships today on the default path.

   Whichever severity you assign, fix all three sites together with one shared `sanitizeForComment` (replace `\r`/`\n`, or refuse) — fixing only the new one leaves the easier vector open.

2. **The emitted note over-claims what `session_replication_role = replica` does** (`recovery.go:351-353`, `docs/query-and-recovery.md:436`). Two inaccuracies in text an operator reads as a safety guarantee:
   - *"the target's own triggers do NOT fire while it applies"* is false for `ENABLE ALWAYS` triggers (they still fire) and — the surprising direction — `ENABLE REPLICA` triggers do **not** fire under `origin` and **do** fire under `replica`. The flag can *cause* trigger side effects that would not otherwise occur, the opposite of what the note promises.
   - *"referential integrity is not enforced during the apply"* reads as a transient window. With RI constraint triggers disabled there is no deferred re-validation at `COMMIT`: rows written in violation stay violating **permanently** while the FK remains marked valid. That permanence belongs in the emitted comment (`recovery.go:404-405`), not only in the docs.

3. **The flag's stated rationale doesn't hold under a filtered recover — the normal invocation.** The justification everywhere is "double-apply side effects the original triggers already logged as their own (separately reversed) events". True only when those sibling events are in the same script. With `--table` or `--pk`, the trigger's audit/counter rows were never fetched, so suppressing triggers means the derived tables are neither re-fired nor reverted — they silently keep the state the bad change produced. Worth one clause: "only avoids double-application when the reversal's scope also covers the tables the triggers write."

4. **Test gap that would have caught (1).** `applycodegen_test.go` has no odd-identifier case. `TestRestoreAutoIncrement_MySQLOnlyAfterCommit:188-192` asserts every `ALTER TABLE` line is commented, but only with tables named `orders`/`items`. `applyCodegenRow("db", "or\nders")` fails that assertion today.

### Suggestions

- **Dialect misdetection makes the warnings misfire** (`internal/cli/recover.go:378-386`). `DialectForIndex` → `query.SourceFlavor` returns `""` (→ MySQL) on a missing `stream_state` row or a read error. On a PG index in that state, `--suppress-triggers` warns "has no effect on a **MySQL/MariaDB index**" — a false statement about the index — and `--restore-auto-increment` emits the MySQL block on a PG script with no warning, contradicting `docs:437`. Low impact (the block is inert text) and the misdetection is pre-existing.
- **`SET LOCAL` fails soft outside a transaction** (`recovery.go:406`): PostgreSQL emits a warning, not an error, and does nothing — triggers then fire while the operator believes they're suppressed. The script's own `BEGIN` covers `psql -f`; a per-statement-autocommit client would not. One line in `docs:436` closes it.
- The checklist lists tables whose counter cannot have drifted (`recovery.go:328`): `touched` includes reversed `UPDATE`s, which never move the counter.
- The block names two intents ("reuse the ids this reversal freed, or never reuse them") but supplies a recipe for only one. The never-reuse value comes from `information_schema.TABLES.AUTO_INCREMENT`, which must be read *before* applying the reversal — a sequencing detail the block omits.

### Strengths

- **Dialect gating is airtight.** Emission is guarded at generation time inside the `PostgresDialect` (`recovery.go:386-408`) and `MySQLDialect` (`:414`) branches, not at flag-parse time — so the shared `internal/cli` registration across both binaries cannot leak either flag onto the wrong dialect. Every other `GenerateSQLFromRows` caller (`mcptools`, `console/api.go`, `console/recover_cascade.go`, `agent/handler.go`, `cascaderecover/emit.go`) was checked: none sets them, so default output stays byte-identical.
- **Transaction framing is exactly as claimed**, and asserted positionally by index rather than by substring presence.
- `\x00` as the `touched` map separator is correct rather than lucky — MySQL forbids U+0000 in identifiers.
- The reasoning for *not* guessing `N` or the column name (`recovery.go:430-440`) is the right call, documented where a contributor will see it.
- Docs honor the constraint: the "fundamental limitation for MySQL" and "the script still never picks `N` for you" bullets both survive intact.

Issue #1107, opened 2026-07-25
A backup fold after a long capture outage can use all of the host's memory


nethalo:
Kind: Bug. State: gates met (#1112; measured 2026-08-28); first slice is a touched-row budget, the disk merge after. As of v0.81.0.

## What happens today
An operator's scheduled backup (a fold) holds every distinct changed row in memory before merging; paging the events does not bound that. On TPC-C, five-minute windows peak at 8.5 GB; one two-hour window after a capture outage reached 14.3 GB six minutes into the fold (3.5 GB at three), and the host's management agent dropped.

## What should happen
First, a touched-row budget: past it the fold refuses before writing, recorded like a capture gap. Then hash the key into K disk buckets and merge one at a time, with the one existing key function.

## Why it matters
data safety.

Words used: fold (an incremental backup: the previous backup plus the changes recorded since, without re-reading the database).

<details><summary>For the engineer (original text, verbatim)</summary>

**Original title:** reconstruct: hash-partition the full-table merge so peak memory is bounded by a divisor, not by rows-touched

Sibling of #1097 (and, further back, #842 / #654). #1097 bounds the **raw event slice** by paginating the fetch; this issue bounds the **folded change map**, which #1097 explicitly leaves unbounded.

## Why #1097 isn't the end of it

`ReconstructTable`'s merge is a hash join whose build side is a Go map:

- build side: `changes map[string]*query.ResultRow` — one entry per **distinct touched PK**, each holding a full `row_after` image;
- probe side: the baseline Parquet, scanned once by DuckDB in its own order.

The build side must be complete before the probe starts, because the probe visits each baseline row exactly once and needs that PK's final image immediately — there is no "come back to it later". So after #1097 the ceiling moves from *"events in the window"* to *"distinct rows touched in the window"*, and a window touching tens of millions of distinct PKs still exhausts RAM.

## Why not just push the join into DuckDB

The obvious answer — load both sides into DuckDB, `LEFT JOIN` on the PK, let it spill — is blocked by the fact that **the join key does not exist as stored data on both sides**:

- the baseline side carries natively-typed PK columns (DuckDB hands back `time.Time` for a DATETIME/TIMESTAMP PK);
- the event side carries `pk_values`, a pipe-delimited string the indexer built with go-mysql's per-type formatting plus `|`→`\|` / `\`→`\\` escaping.

[...]

Issue #797, opened 2026-07-05
baseline: anchor is a 1-second timestamp, not the GTID/binlog position it already records


nethalo:
All consumers (`fulltable:381`, `shim/snapshot.go:138` and `:415`, `verify/verify.go:136`, cascade) use `Since=snapshotTime` ('Started dump at', inclusive `TO_SECONDS >=`). The exact GTID/file/pos of the snapshot ARE in the Parquet metadata (`MetaKeyGTIDSet`/`BinlogPos`) but are only copied to the output metadata-file; the docs (`dump-and-baseline.md:221`) say the anchor marks "where the deltas begin", which the code doesn't fulfill.

Location: `internal/reconstruct/fulltable.go:381`

The double-apply window is absorbed by last-write-wins merge (idempotent), but the LOSS window is not: row-event headers carry the statement's EXECUTION time, not the commit — a transaction that executed its UPDATE at 09:59:50, with dump 'Started at' 10:00:00 and commit 10:00:05, is neither in the dump's MVCC snapshot NOR in the fetch (ts<Since) → silently stale row in reconstruct/_snapshot/verify (verify would see it as a causeless MISMATCH). Indexed events carry gtid/binlog_file/start_pos, so position-filtering is buildable; pgbaseline already designed the exact contract with `MetaKeyLSN`.

Fix: filter deltas by the baseline's recorded GTID/position instead of the 1s timestamp.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #1112, opened 2026-07-25
reconstruct: stream the full-table event window instead of materializing it (#1097)

### LINKED ISSUE #1097: reconstruct: stream/batch the event window with incremental merge instead of materializing it in RAM
Follow-up to #842, tracking the one fix from that issue's list that is **deliberately deferred** — the other three landed in #1098.

#842 said:

> Fix: batch/stream the event window with incremental merge; pass the container-safe budget to the merge/baseline DuckDB instances; account the global parallelism against the per-table threshold; write a completeness marker.

The last three items (container-safe DuckDB budget for the merge/baseline sessions, parallelism-scaled volume warning, completeness marker) are done. This issue tracks the first and largest one, left out of that PR as a separate, bigger piece of work.

## The problem

`internal/reconstruct/fulltable.go`'s full-table path (`ReconstructTable` → `query.FetchMerged`) fetches **every** matching event for a table's window in one call and holds the whole result in memory:

- Each `query.ResultRow` carries the fully-parsed JSON `row_before`/`row_after` images, not just a pointer or a diff.
- `ReconstructTable` then builds a change map (`map[string]*query.ResultRow`) with one entry per **touched PK**, keyed by `pk_values` — a second RAM cost on top of the raw event slice, proportional to the number of distinct rows touched (not events).
- `maybeWarnEventVolume`/`--warn-event-threshold` (now parallelism-scaled, #842) is **advisory only** — it logs a warning above the threshold and keeps going. There is no hard cap: a table whose window genuinely holds tens of millions of events, or a handful of pathologically hot rows with huge BLOB/TEXT payloads, can still exhaust RAM regardless of the DuckDB session's own budget, because none of that memory is DuckDB's — it's plain Go heap holding decoded events.

Capping the merge/baseline DuckDB sessions' resource budget (the #842 fix that did land) bounds DuckDB's own footprint, but does nothing for this Go-side event/change-map memory, which for a wide window or a BLOB/TEXT-heavy table can dominate.

## What's needed

Batch/stream the event window fetch with an incremental merge, so the process holds a bounded slice of events (and change-map entries) at a time instead of the whole window — the same shape of fix `recover`'s reversal-script bound (`--max-script-bytes`, see `docs/query-and-recovery.md §Memory Footprint`) applies to a different stage of the same command family. This likely means:

- Fetching + merging events in time-ordered chunks against the baseline stream, rather than one `FetchMerged` call materializing the whole `(baseline, --at]` window.
- Deciding how the change map's "last write wins per PK" semantics survive chunking (a PK touched in chunk 1 and again in chunk 3 must still resolve to chunk 3's image).
- Preserving the existing correctness guards that scan the **raw event slice** (`pkChangingUpdateInEvents` for #782, the TOAST-marker check, the added/dropped-column guards for #602/#843) — these currently assume a single complete slice; a chunked design needs an equivalent window-complete check per chunk or a bounded carry-over state.
- Turning `--warn-event-threshold` from warn-only into an actual bound once the fetch is chunked, if that's still desired — or documenting that streaming removes the need for it.

## Scope

This is explicitly a bigger, riskier change than the three #842 fixes that already landed (DuckDB session budget, parallelism-scaled warning, completeness marker) — it touches the core fetch/merge algorithm of the full-table path, not just its resource wiring, so it's tracked separately rather than folded into that PR.

nethalo:
closes #1097

## Summary

Full-table `reconstruct` fetched **every** event in a table's baseline→target window in one `query.FetchMerged` call and held the whole slice — each event carrying both decoded JSON row images — before folding it into the `PK → last event` change map. The window is now walked a page at a time and folded incrementally.

Peak memory moves from *events in the window* to *distinct rows touched*, and `row_before` (which nothing downstream reads) is dropped at fold time rather than retained for the life of the map.

## How

- **Keyset pagination, not OFFSET.** `query.Options.AfterEvent` cuts on the composite `(event_timestamp, event_id)` sort key. Both components are needed: `event_timestamp` has one-second resolution and collides heavily, so a timestamp-only cursor would skip or re-return the boundary second. `event_id` makes the key total.
- **`query.FetchMergedStream`** hoists source discovery, planning and gap enforcement out of the page loop, so a `*GapError` still surfaces once before any data is read rather than once per page.
- **Archive scoping advances with the cursor** (`parquetquery.sinceLowerBoundHint`). Without this the row-level predicate alone would be correct but ruinous: archive files are listed and downloaded per source, so every page would re-read the whole window and discard everything before the cursor.
- **`--fetch-batch-size`** (default `100000`, env `BINTRAIL_RECONSTRUCT_FETCH_BATCH`) exposes the page size.

Both full-table paths use it — the baseline merge and the no-baseline binlog-only fallback, which has no baseline to bound its window and is if anything the more exposed of the two.

## The guard migration is the load-bearing part

The two guards that read an event's **before-image** — [#592](https://github.com/dbtrail/dbtrail/issues/592) (unresolved TOAST marker) and [#782](https://github.com/dbtrail/dbtrail/issues/782) (PK-changing UPDATE) — now run per event in `foldPage`, on the untrimmed event, and their calls are **deleted** from `mergeBaselineIntoWriter` and `writeBinlogOnlyChanges`.

Left in place they would have been worse than absent: `retainEvent` blanks `RowBefore`, so both would have found nothing to inspect, returned clean on every input, and still read as guards at the call site.

Per-event checking is also **strictly stronger** for #782 — the map only ever held the surviving last event per PK, so a PK-changing UPDATE whose old key a later event reused was invisible to a map scan. The map-level helpers stay for the callers whose maps do carry before-images (the shim's `_snapshot`).

## What this does NOT fix

The change map still holds one entry per **distinct touched PK**, and paging does not bound it: the merge scans the baseline once and needs each row's final image at the moment it reaches that row. A window touching tens of millions of distinct rows is still dominated by the map. That half is [#1107](https://github.com/dbtrail/dbtrail/issues/1107), and the docs say so rather than implying the OOM is closed.

## Trade-off, stated plainly

A page costs one index query **plus one scan per archive source**. Cursor-advancing file scoping keeps this near-linear rather than quadratic, but the file at the page boundary is re-read once per page — so a very small page over a mostly-archived window is slower. `docs/query-and-recovery.md` gets a §Streaming the event window covering the knob in both directions.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go vet -tags integration ./...`
- [x] `go test ./... -count=1`
- [x] `go test -race ./internal/... ./cliapp/...`
- [x] New: paging delivers the window exactly once in order and runs the prologue **once** (pinned by sqlmock expectations); short-page termination; a skipped source under `AllowGaps` does **not** end the stream early; a non-advancing cursor fails loudly instead of looping forever; incompatible option combinations refused before any DB work.
- [x] New: the #782 reinsert permutation **split across a page boundary** — the case a whole-slice scan could never get wrong and a paged fold could; last-write-wins across pages; `retainEvent` drops the before-image and does not alias its page.
- [x] New: `sinceLowerBoundHint` advances with the cursor without overriding the [#797](https://github.com/dbtrail/dbtrail/issues/797) `SincePos` margin.

Note: two commits — `7dd87f0` (the fetch layer) was committed by a concurrent session mid-work; content is unchanged and the squash-merge folds them into one.


nethalo:
Pushed `edf5d6a` — three review findings, all addressed.

**1. The keyset SQL was never executed by an engine.** Every test proving keyset correctness ran against a stub that implemented the semantics *as I intended them*; `go vet -tags integration` proved the predicates compile, not that they evaluate. Both are now executed for real, and both fixtures put several events in the **same second** — the case a timestamp-only cursor gets wrong silently in one of two directions:

- `internal/parquetquery`: pages a real Parquet fixture through DuckDB at page size 2 and asserts the result is identical to the unpaged read. Plain unit test — DuckDB is in-process, no Docker.
- `internal/query` (integration): sweeps page sizes 1, 2, 3, 5, 12, 50 against a real MySQL index and asserts every one matches the unpaged fetch. Page size 1 means *every* page ends mid-second. This is also what proves a `time.Time` binds correctly against the DATETIME column and that the hour-aligned `TO_SECONDS` pruning hint never excludes a row it should have kept.

**2. Per-page DB work in the decode passes.** `MapEventEnumLabels` and `DecodeEventBinaries` each rebuilt their epoch-aware state — a `schema_snapshots` query plus a resolver load per epoch touched — on every call. Free when a call meant "the whole window"; a 400-page window paid it 400 times. Both are now backed by an `eventDecoder` built once per table and reused across pages, with its typed-ness accumulating across pages. The exported one-shot functions keep their contract for callers that still fetch in a single call (`verify`, single-row reconstruct, console).

**3. The archive re-read cost was understated.** "The boundary file is read again on each page" reads like a rounding error. It isn't: cursor scoping advances at **hour** granularity, archive partitions are hourly, and on S3 `parquetquery` *downloads* each file rather than scanning in place — so a given hour's file is re-fetched about `events in that hour / --fetch-batch-size` times, and DuckDB row-group statistics prune the scan, not the download. ~1× at the default for an hour holding 100k events, ~10× for one holding 1M. The formula is now in the docs and the flag help instead of a vague warning.

Also recorded in `foldPage` why the TOAST check being *stricter* than the map-level one it replaced is inert today — the marker is a PostgreSQL concept and full-table reconstruct refuses PG sources ([#597](https://github.com/dbtrail/dbtrail/issues/597)) — and what to revisit if that gate is ever lifted.

## Verification after the changes

- `go build ./...`, `go vet ./...`, `go vet -tags integration ./...`
- `go test ./... -count=1` — clean
- `go test -race ./internal/... ./cliapp/...` — clean
- `go test -tags integration ./internal/{reconstruct,query,parquetquery,verify,cli}/ -count=1` — all pass
- `go test -tags integration -run TestEndToEnd .` — passes

nethalo:
Pushed `224e26d` — review round (5 agents: code, tests, silent-failure, comments, type-design). **Two critical findings, both mine, both with the same shape: a guard that stopped guarding while its own test stayed green.**

## Critical 1 — the #843 dropped-column guard went dead

Found independently by four of the five agents; two reproduced it empirically.

`droppedBaselineColumns` reads `ev.RowBefore` and runs on the map `retainEvent` trims. For a PK whose only post-drop evidence is a **DELETE** — no `row_after` at all — it saw nothing, so the run proceeded and emitted a dump carrying the baseline's `CREATE TABLE` still declaring the dropped column, with every never-touched pass-through row re-emitting its stale pre-drop value. For a column dropped for PII reasons, that is data re-exposure through a recovery artifact.

I enumerated the before-image readers in `retainEvent`'s doc comment and missed this one — in a PR containing four separate comments warning about exactly this failure mode. Prose was not the right mechanism.

`TestReconstruct843_columnDroppedDeleteOnlyFailsLoud` stayed green the whole time, because it hand-builds the change map with `RowBefore` populated — a shape production can no longer produce.

**Fix:** `foldResult.observeImages` narrows an intersection of image key-sets during the fold, while the images are intact; `droppedBaselineColumns` now takes that intersection plus `colNames`. The check could not simply move into `foldPage` alongside #592/#782 because `colNames` are not known until the baseline is materialized, well after the fold. Every #843 test now runs through the real fold, so a re-break fails the suite. One deliberate behavior change is documented: seeing every event rather than the surviving last-per-PK makes it stricter for a drop-then-re-add window.

## Critical 2 — the stream could end with a nil error mid-window

`FetchMergedStream` consulted `skipped` on the short-page branch and **not** on the empty-page one. `AllowGaps=true` + a failing archive source + a MySQL side already out of rows returned success having delivered nothing — and the caller then reconstructs from the baseline alone: post-baseline DELETEs resurrected, INSERTs missing, UPDATEs stale, loads cleanly. The triggering topology is the ordinary one, since MySQL holds the recent partitions and archives the old ones.

**Fix:** retire sources that came back **empty** — sound because the cursor only advances, so every later page applies a strictly tighter filter and a source behind it answers the same forever; this also kills a per-page full-prefix S3 stale-registration probe and its warning (#383). Retire sources that **failed** too, but never let one end the walk for the sources still healthy, and emit one end-of-walk warning naming them. The retirement is what guarantees termination.

## Tests that were not testing what they claimed

- `TestWriteBinlogOnlyChanges_noPartialOutputOnRefusal` never called the function; it asserted a freshly created `TempDir` was empty, so it could not fail. Deleted.
- `foldEventWindow` had **no coverage at all**. Added two tests pinning the four things it does around the fold, each of which regresses silently: decode state built **once** (sqlmock arms exactly one `schema_snapshots` read), `First` captures only the first event, `Total` accumulates across pages, the volume warning latches once per table.
- Nothing asserted `foldPage` stores `retainEvent`'s copy — reverting it to `changes[pk] = ev` kept the whole suite green while re-pinning every page in memory.
- `TestSinceLowerBoundHint_cursorNeverOverridesSincePosMargin` compared `05:00` against `05:00`, a tie that passes even if the #797 margin is deleted. Added the discriminating case.
- `--fetch-batch-size` validation had no test.

## Also

`eventDecoder.typed` accumulated across pages and was never read (a dump could carry base64 with only a `Debug` line) → now warns. #592/#782 refusals reached the operator prefixed `fetch events:` → unwrapped. `AfterEvent`+`DESC` enforced only inside `FetchMergedStream` → moved to both engine entry points. **The S3 re-fetch formula I published understates by ~2x** (pruning keeps a file whose hour *ends* at-or-after the bound, so the preceding hour is re-fetched and yields nothing) → corrected in all three places, plus eight stale comments.

## Deferred, recorded on #1107

A purpose-built trimmed change-row type that would make Critical 1 a **compile error** instead of a comment. #1107 rewrites this map anyway, and "what must an entry carry" is the same question its temp-file format has to answer.

## Verification

`build` · `vet` · `vet -tags integration` · `go test ./...` · `go test -race ./internal/...` · `-tags integration` on reconstruct/query/parquetquery/verify/cli · `TestEndToEnd` — all green.

Issue #1102, opened 2026-07-25
console: cascadeBaselineProvider bypasses bundle.findBaseline, losing the #766 local→S3 fallback


nethalo:
## Gap

`internal/console/manager.go` defines `bundle.findBaseline` (around line 257):

```go
func (b *bundle) findBaseline(ctx context.Context, schema, table string, at time.Time) (string, time.Time, reconstruct.StaleWarning, error) {
    ...
    if b.baselineFallbackSrc == "" || !errors.Is(err, reconstruct.ErrNoBaseline) { ... }
    return reconstruct.FindBaseline(ctx, b.baselineFallbackSrc, schema, table, at)
}
```

This wraps `reconstruct.FindBaseline` with a fallback: when the primary `b.baselineSrc` has no baseline for the table (`ErrNoBaseline`), it retries against `b.baselineFallbackSrc` — the durable S3 copy (#766's local→S3 fallback).

`internal/console/recover_cascade.go` constructs the cascade Phase-2 provider directly against the bundle's primary source (around line 341):

```go
baselineProvider = &cascadeBaselineProvider{source: b.baselineSrc, resolver: b.resolver}
```

and `cascadeBaselineProvider.BaselineChildren` (around line 532) calls `reconstruct.FindBaseline(ctx, p.source, schema, table, at)` directly — **not** `b.findBaseline`. So the cascade Phase-2 fallback only ever sees `b.baselineSrc`; if the local baseline directory doesn't have the table (or is missing/pruned) but the S3 fallback copy does, cascade recovery silently gets `ErrNoBaseline` (falls back to Phase-1-only, flagged as `nobaseline:` incomplete) instead of finding the table via `b.baselineFallbackSrc` — the same rescue the rest of the console (reconstruct tab, `_snapshot` shim route, etc.) already gets via `b.findBaseline`.

## Suggested fix

Either:
- have `cascadeBaselineProvider` hold a reference to the `*bundle` (or an equivalent lookup function) and call `b.findBaseline` instead of `reconstruct.FindBaseline` directly, or
- pass a `FindBaseline`-shaped function value into `cascadeBaselineProvider` at construction time so the cascade path composes with whatever fallback the bundle has configured, without a direct `*bundle` dependency.

[...]

Issue #783, opened 2026-07-05
reconstruct: --at cutoff is by 1-second statement timestamp, never a transaction boundary — can produce states that never existed


nethalo:
`Fetch` uses `event_timestamp <= at` (`query.go:341`) and `ApplyAt` cuts on `After(at)`: a Rows_event's timestamp is the statement's EXECUTION time, not the commit time.

Location: `internal/reconstruct/reconstruct.go`

Scenario: a multi-second transaction (statements at T1 and T2, commit at Tc) is split by an `--at` between T1 and T2 (half the transaction is applied), and a transaction that executed before `--at` but committed AFTER is included whole — in both cases the reconstructed state never existed in the source.

This is "implementation, not fundamental": the binlog gives commit order and GTID (the `gtid` column is indexed), so a transaction-boundary cut (verify's `UntilPos` style, or grouping by GTID) is buildable with existing data; sub-second would be irresolvable with the `DATETIME(0)` the index persists today. No doc mentions the cut's non-atomicity.

Fix: cut at the transaction boundary using the indexed GTID/position; document the 1s granularity limit.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #1114, opened 2026-07-25
mcp: add a read-only reconstruct (time-travel) tool

### LINKED ISSUE #953: mcp: add a read-only reconstruct (time-travel) tool — the flagship capability is unreachable from the AI surface
cmd/bintrail-mcp/main.go:97-134 registers exactly four tools (query, recover, status, list_schema_changes); docs/mcp-server.md:13 pitches "four read-only tools." The console ships single-row PITR (internal/console/reconstruct.go -> `reconstruct.FindBaseline -> ApplyAt/BuildHistory`) and the CLI has `reconstruct --at/--history` + `recover-cascade`, but the AI surface has neither — an agent can't answer "what did this row look like at 3pm," and its delta-only `recover` tool emits incomplete SQL when asked to undo an FK CASCADE delete. Engine is source-shared -> thin plumbing.

Fix: add a `reconstruct` MCP tool (schema/table/pk/at/history, baseline_dir/baseline_s3 or env fallback) reusing FindBaseline/ApplyAt/BuildHistory, gated as the console gates it (baseline configured, AllowGaps=false default); optionally a `recover_cascade` tool. Update the "four tools" framing. Distinct from #707 (forensics MCP tools).

---

nethalo:
## What

Adds a fifth read-only MCP tool, `reconstruct` — single-row time travel — so an
agent can answer *"what did this row look like at 3pm?"*.

`recover` is **delta-only**: it reverses events it actually has, so it emits an
incomplete script for a row nobody touched in the retained window (the FK
`ON DELETE CASCADE` case in the issue). The console and the CLI both ship
single-row PITR; this wires the same engine (`FindBaseline` →
`ReadBaselineRow` → `ApplyAt`/`BuildHistory`) to the AI surface.

Parameters: `schema`/`table`/`pk` (required), `at`, `history`, `allow_gaps`,
`baseline_dir`/`baseline_s3`. Annotated `ReadOnlyHint: true` +
`IdempotentHint: true` like its siblings.

## Registration is opt-in per surface

The tools moved to the shared `internal/mcptools` since this issue was filed, so
registering unconditionally in `NewServer` is a fork in the road, not plumbing.
New `Config.Reconstruct` gates it: a surface that cannot supply a baseline lookup
would otherwise advertise a tool that can only ever error. **Both** surfaces opt
in.

| | Baseline source | Gate |
|---|---|---|
| `bintrail-mcp` (standalone) | `baseline_dir`/`baseline_s3` per call → `BINTRAIL_BASELINE_DIR`/`BINTRAIL_BASELINE_S3` — mirroring how `index_dsn` falls back to `BINTRAIL_INDEX_DSN` | a source resolved at all |
| console `/mcp` | the bundle's own `findBaseline`; the parameters are **rejected** exactly like `index_dsn` | `baselineConfigured`, same signal as `/api/reconstruct` |

Two things worth reviewer attention on the console side:

- It binds `b.findBaseline`, **not** `reconstruct.FindBaseline` — the method
  carries the #766 local→S3 fallback, and substituting the package function is
  precisely the regression #1102 just fixed for cascade recovery.
- `/api/reconstruct` additionally refuses a session carrying a data profile
  (#1075). That has **no analogue** on `/mcp`: the endpoint authenticates by
  token only (`policyCtxKey` is set solely in `tokenMiddleware`'s session
  branch, which `/mcp` bypasses), so no session policy can ever reach these
  handlers. `Target.ProfileActive` is still refused as a belt.

## Correctness posture: stricter than `/api/reconstruct`, on purpose

An agent cannot eyeball a suspicious fold the way a human reading CLI output
can, so the tool takes the strictest available posture — **this is deliberate
divergence, not drift**:

- `allow_gaps` defaults **false** (the opposite of `query`/`recover`): a
  coverage gap yields a state that never existed, not a partial one.
- Adds `CheckDestructiveDDL` (#764) and `CheckCaptureGap` (#765), which the
  console endpoint does not run. Both are hard refusals — no archive can refill
  either, and the fold would resurrect rows that are gone.
- The `> MaxReconstructEvents` overflow check stays **before**
  `TrimPartialTailTransaction` (#783): trimming shrinks the set, so the reverse
  order lets an over-cap window slip through.
- `GapError`/`SourceEmptyError` are rewritten with remediation an MCP client can
  act on (`allow_gaps: true`, `bintrail archive reconcile`) instead of the CLI's
  `--allow-gaps` flag advice — same precedent as the `recover` tool's
  `ScriptBudgetError` branch. Note `FetchMerged` resolves archives from
  `archive_state`, not the `BINTRAIL_ARCHIVE_S3` env pair the standalone `query`
  tool consults, because gap detection is planner-driven off that same table;
  that is why the remediation names `reconcile`. Identical to the CLI.
- `StaleWarning` (#466) is surfaced as a `stale_baseline:` warning rather than
  dropped, alongside coverage-gap warnings.
- A missing baseline row whose earliest delta is not an INSERT **warns**
  (`pk_change_suspected`, #782) instead of refusing as the CLI does: this surface
  must also serve the legitimate row-created-after-the-baseline case, which the
  console handles and which is indistinguishable except by that first event type.

## Open-core boundary

The payload is reconstructed **column state**, built field-by-field from
`reconstruct.StateEntry` / the folded row map — never a serialized
`query.ResultRow`. So `connection_id`, `query_text` and `query_hash` (#699, the
three the console's `eventDTO` withholds) structurally cannot appear, and
`Target.RedactStatementText` has nothing to strip. This matches the existing MCP
`query` tool's behaviour on the console surface rather than inventing a new rule:
there, the same three fields are stripped post-fetch to match `eventDTO`.

## Deliberately left out

- **`recover_cascade`** — the issue marked it optional and it is not thin: it
  needs the cascade engine, a `cascadebaseline.Provider`, a resolver, and its own
  distinct gate (the console refuses it under a profile because victim synthesis
  cannot honor redaction — a *different* rule from reconstruct's). Worth its own
  issue.
- **CLAUDE.md** — its MCP section still says "four read-only tools". The file is
  untracked (`git ls-files` has no match) and exists only in the primary
  checkout, which this worktree must not modify; it needs a manual one-line edit.

## Tests

- Unit (`reconstruct_test.go`, no MySQL): opt-in registration + annotations +
  input schema; `index_dsn`/`baseline_dir`/`baseline_s3` rejected before any
  connection is resolved; required-parameter and bad-`at` validation; profile
  refusal; missing-baseline actionability on both surfaces; standalone source
  precedence (param > env, dir > s3); composite-PK arity; warning assembly
  including the INSERT-first case that must **not** warn.
- Integration (`//go:build integration` + `NewInMemoryTransports`, protocol
  `2025-06-18`): point-in-time fold across baseline/UPDATE/UPDATE/DELETE plus
  never-existed; history mode (baseline entry + 3 transitions, DELETE flagged);
  **gap refused by default** with the remediation naming `allow_gaps: true` and
  asserting it does *not* hand back a CLI flag; missing baseline; event-cap
  refusal.
- `internal/console/mcp_integration_test.go` updated 4 → 5 tools, plus asserting
  the ungated call and the rejected `baseline_dir` param both error.

`go build ./...`, `go vet -tags integration ./...`, and `go test ./... -count=1
-race` all pass (0 failures), and the five new integration tests pass against
Docker MySQL.

closes #953


nethalo:
**Merge-order note: overlaps with #1113** (`ext: extension seam for MCP tools`), which was open in parallel. Three files in common:

- `internal/mcptools/mcptools.go` — #1113 adds the extension seam to the registration path, this PR adds `Config.Reconstruct` to the same struct/registration.
- `internal/console/mcp.go` — both change how the console builds its tool set.
- `docs/mcp-server.md` — **both rewrite the "four read-only tools" framing**, from different directions.

Not duplicated work (a seam vs a tool), but whichever merges second needs a rebase, and the docs conflict is the one most likely to resolve badly — a mechanical merge could leave the tool count wrong in one direction or the other. Worth resolving deliberately rather than accepting either side.

nethalo:

### Critical — `allow_gaps: true` silently swallows a PERMANENT capture gap, and the refusal message leads an agent straight into it

`internal/mcptools/reconstruct.go:307` passes `args.AllowGaps` into `reconstruct.CheckCaptureGap`. Trigger: a stamped `stream_state.gap_lost_at` inside `(snapshotTime, atTime]` — events lost at the source, which no archive can refill.

1. With `allow_gaps: false`, `internal/reconstruct/capturegap.go:56` returns text ending *"pass `--allow-gaps` to proceed with a known-incomplete reconstruction"*, returned verbatim via `ErrorResult`. That is the CLI-flag-leak anti-pattern `reconstructFetchError` (`:435-447`) exists to eliminate — but it only covers the planner `GapError` path, so this one is unguarded.
2. The agent translates that to `allow_gaps: true`, the only override it has.
3. That reaches `capturegap.go:49-54`, which returns `nil` after a `slog.Warn`. `reconstructWarnings` (`:452`) takes no capture-gap input, so **nothing reaches the payload**.
4. The planner does not compensate: `gap_lost_at` is a source-side loss, so live partitions stay continuous and `plan.GapHours` is empty.

Result: `found: true`, a full `state`, zero `warnings`, over a fold missing permanently-lost events. `docs/mcp-server.md:275` asserts the opposite as fact ("refuses outright — no override"). The CLI surfaces this on stderr; MCP is the one surface where the warning is invisible.

Fix at the call site: either don't pass `AllowGaps` to `CheckCaptureGap` (making the doc true), or wrap it like `reconstructFetchError` does — rewrite `--allow-gaps` → `allow_gaps: true` on refusal, and push a `capture_gap:` entry into `Warnings` when overridden.

### Important

1. **The #765 guard is inert on exactly the indexes the console serves.** `CheckCaptureGap` keys on `ss.GapLostAt.Valid` and never consults `status.StreamStateInfo.GapColumnsPresent` (`internal/status/status.go:100,371`). On a legacy index missing the `gap_lost_*` columns it silently returns `nil` — and the console sets `EnsureSchema: false` (`internal/console/mcp.go:143`) and never migrates registry servers. `status` carries `GapColumnsPresent` precisely to report *unknown* rather than *ok*; this should too.
2. **`reconstructFetchError` has no MySQL 1054 arm.** The fetch reaches `Engine.Fetch` with `EnsureSchema: false` on the console. `internal/console/api.go:663-681` maps 1054 to an actionable 422; `reconstruct.go:446` returns it raw. Pre-existing for the `query`/`recover` tools too — consistency gap, not a regression.
3. **`packaging/mcpb/manifest.template.json:33-49` still declares four tools** — `reconstruct` missing from the `.mcpb` bundle metadata. `scripts/validate-mcpb.sh` doesn't check it against the server, so nothing catches the drift. Every *assertion* was correctly updated; this is the one miss.
4. **Three prose enumerations still list four tool names**: `docs/mcp-server.md:65`, `docs/connect-ai.md:92`, `docs/guide.md:370`.
5. **Managed-MCP-token scope now covers unredacted baseline reads.** `POST /api/mcp-token` requires only `PermSettingsRead` (`internal/console/authz.go:92-94`) while `/api/reconstruct` requires `PermReconstructExecute` (`:56`). The same widening already exists for query/recover since #1052, so this PR extends an existing design rather than opening a new hole — but it extends it onto the surface whose reads bypass redaction. Worth an explicit decision plus a line in `docs/console.md`.

### Verified as claimed

- **The RBAC claim holds, independently checked.** `policyCtxKey` is set at exactly one site (`internal/console/auth.go:115`, inside `tokenMiddleware`'s session branch); `/mcp` is mounted on the root mux (`internal/console/server.go:534-535`) outside that middleware. No session policy — and therefore no session data profile — can reach these handlers. `Target.ProfileActive` is a genuine belt and covers the zero-rule case.
- **No open-core leak.** `reconstructResult`/`reconstructStateEntry` are built field-by-field and mirror the console's DTOs exactly; `connection_id`/`query_text`/`query_hash` structurally cannot appear.
- Gating is enforced at the endpoint, not just advertised; `AllowBaselineParams: false` closes the arbitrary-storage hole; the console binds `b.findBaseline`, preserving the #766 local→S3 fallback (the #1102 regression is not repeated).

### Note

PR #1113 touches three of the same files; expect mechanical conflicts in the `Config` struct + registration block, `newMCPServer`'s config literal, and the `docs/mcp-server.md` tool table. Whoever merges second must re-check the tool-count assertions still match.

PR #1115, opened 2026-07-25
verify: add a recover-input check that walks the before-image chain (#1001)

### LINKED ISSUE #1001: verify: add a recover-input verification mode (before-image chaining)
## Background

Split out from #796 (closed: the docs over-promise was corrected). `docs/verify.md` now states plainly that `verify` **does not exercise the `recover` path** — it reconstructs full-table state from the *latest* event per PK (`row_after`, `LimitPerPK=1`) and compares at baseline anchor points. This issue tracks the deferred **code** enhancement that would close that coverage gap.

## Problem

`recover`'s reversal SQL depends on data that a full-table content match never touches:

- **`row_before` images** — used to undo `UPDATE`s and to re-`INSERT` `DELETE`d rows.
- **DELETE row images** — the pre-image of a deleted row.
- **superseded intermediate events** — an event later overwritten by a newer event on the same PK.

A corrupt `row_before`, a corrupt DELETE image, or a corrupt intermediate event all pass today's `verify` cleanly, yet `bintrail recover` would emit wrong reverse SQL from those same rows. So a green `verify` does not imply a correct `recover`.

## Proposed enhancement

Add a recover-input verification mode (opt-in flag, e.g. `--check recover` or a distinct subcommand) that walks the per-PK event chain and asserts the before/after images are internally consistent — for example, per PK: `UPDATE_n.row_before == state_{n-1}` (the reconstructed state just before event *n*), and that DELETE pre-images match the last known state. This is buildable with data already in `binlog_events` (no new capture), unlike the current anchor-point full-table check.

## Scope / notes

- Purely additive; the existing baseline-anchored `verify` stays the default.
- No source or index schema changes expected.
- `internal/verify/verify.go` is the home; `internal/recovery` defines the exact image dependencies to mirror.

nethalo:
> **⚠️ This PR stacks on #1109 and must merge AFTER it.**
> Base branch is `worktree-agent-aa73a670535ad9d80` (#1109, `verify --format json`), not `main`. This PR emits through #1109's `Report`/`TableReport` types and its `Report.ExitError()`; merging it first would break the build.

closes #1001

## Audit first (STEP 0)

The issue's claims hold against the code as it stands — nothing was already done:

- `internal/verify/verify.go:158-170` and the baseline-pair path both fetch with `LimitPerPK: 1`, i.e. the latest event per PK only.
- `internal/recovery/recovery.go` confirms the dependency set exactly: `buildInsert` (reversing a DELETE) dereferences `row_before`; `buildUpdate` dereferences **both** images (`row_before` → SET, `row_after` → the PK WHERE); `buildDelete` (reversing an INSERT) dereferences `row_after`. Each refuses on a nil one, and #784 makes one refusal abort the whole script.
- `docs/verify.md:22` states plainly that verify does not exercise the recover path.

So a corrupt `row_before`, a corrupt DELETE pre-image, or a corrupt superseded intermediate event all pass verify today while `recover` would emit wrong reverse SQL from those same rows.

## What this adds

`bintrail verify --check recover` walks each primary key's event chain in ascending time order and asserts the images are internally consistent. Index-only — no baseline, no live source. Per event:

1. **The nil-image contract, mirrored from `internal/recovery`** — DELETE needs `row_before`, UPDATE needs both, INSERT needs `row_after`. (An INSERT with a nil `row_before` is normal and is *not* flagged.)
2. **`event.CheckUnresolvedToast` on both images** — the same up-front refusal `GenerateSQLFromRows` performs before it renders anything.
3. **The core assertion:** every UPDATE/DELETE `row_before` equals the state the previous event on that key left behind.

Superseded events are visited, which is precisely the class `LimitPerPK=1` skips.

## Getting "cannot check" vs "checked and wrong" right

A false MISMATCH on a truncated window would make the mode untrustworthy, so these are all **INCONCLUSIVE, never mismatch**:

- a chain whose first in-window event is not an INSERT (no predecessor state — the normal case for any retention window);
- a **coverage gap** (`AllowGaps: false` → `query.GapError`): a chain with an interior hole cannot be asserted against;
- a window that exceeded the event budget (only its oldest events were walked);
- a table with no PK, or an unavailable index DB name (gap detection could not run).

A mismatch found *inside* a truncated window **is** still conclusive — the events that were walked are real. And a table where no assertion could be made at all reports inconclusive rather than "match": an unproven run must never read as success.

## Normalization traps

Comparison reuses this package's existing canonicalizer (`renderCellNormalized`) and its deferred-type gate (`deferredValueUnresolved`) rather than a fresh `reflect.DeepEqual` — that keeps the JSON key-order, zero-date-vs-NULL, TIME-fraction, FLOAT/DOUBLE-exponent and BLOB/base64 fixes in force. Two extra traps specific to this mode:

- **Epoch drift.** Both operands are event images, but two events on one PK can carry different `SchemaVersion` — event N-1's `row_after` may hold a raw ENUM ordinal while event N's `row_before` holds the label. Same data, different bytes. Images go through `MapEventEnumLabels` + `DecodeEventBinaries` first (as `VerifyTable` does), and a residual unresolvable deferred value degrades to inconclusive.
- **PK-changing UPDATE.** The parser files an UPDATE under the **before-image** PK, so the assertion is valid — but the row has *moved*. State becomes unknown afterwards, or the real `UPDATE pk 1→2; INSERT pk=1` sequence (both stored under key `1`) would false-mismatch. Detection recomputes both keys from the same event's images, the same construction `internal/reconstruct` uses (#782).

`EventSnapshot` rows are filtered out — recovery rejects them outright ("baseline rows are read-only"), and folding one in as a delta would manufacture a mismatch from nothing.

## Memory bound

**`--max-events` (default 200,000) events per table, plus one state entry per distinct PK seen.** `Order: "ASC"` with `Limit: maxEvents+1` — the `+1` makes truncation *detectable* rather than silent.

The correctness argument for capping this way: a prefix of the globally time-ordered stream is simultaneously a prefix of *every* PK's chain, so no walked chain has an interior hole; findings in the prefix stay conclusive and a clean-but-truncated run degrades to inconclusive.

Honest limitation: the cap is a **row count, not a byte budget**, so a BLOB/TEXT-heavy table is still large at a given event count (documented in `docs/verify.md`). This deliberately does not add a new instance of #1097's class, but it does not fix it either.

## Report integration (extending #1109, not paralleling it)

- New `verify.ModeRecoverInputs` constant.
- Three **additive `omitempty`** fields — `events_checked`, `chains_checked`, `chains_inconclusive` — threaded through the existing `TableResult` → `NewReport` copy loop → `TableReport`. Why additive rather than reusing the existing shape: `source_rows`/`reconstruct_rows` mean "rows in a table" to every current consumer, and this mode compares no table content; overloading them would be a silent contract break. A test asserts the content modes' JSON is byte-identical (the new keys are absent).
- **No second report constructor and no second exit path** — the run goes through `NewReport` and `Report.ExitError()`, so a recover-input mismatch fails a CI gate exactly like any other. `writeVerifyText` is untouched; the counts ride in the DETAIL cell.

## CLI

`--check content|recover` (default `content`), `--lookback` (default `30d`, matching `recover-cascade`'s convention), `--max-events`. The `--baseline-dir`/`--baseline-s3` requirement is relaxed for `--check recover` (it never opens a baseline), and `--source-dsn` is **rejected** there rather than silently ignored — accepting it would imply the live source was consulted when it was not.

## Tests

All pure (no Docker), so they actually run under the `-race` suite CI uses:

| Case | Expected |
|---|---|
| clean chain | MATCH, with assertions counted |
| corrupted `row_before` | MISMATCH, naming column + event |
| corrupted DELETE pre-image | MISMATCH |
| window starting mid-history | INCONCLUSIVE (never mismatch) |
| superseded intermediate event | MISMATCH — plus the premise proven: the latest-event-only view hides it |
| `recoverInputsFetchOptions` | guard: `LimitPerPK` unset, `Order` ASC, `Limit` = max+1 |

Plus truncation (clean → inconclusive; mismatch → still conclusive), the four nil-image cases, INSERT-without-before-image is normal, PK move + key reuse, event-after-delete, per-PK independence, JSON key-order, column-set difference across vs within a schema version, and snapshot-row filtering.

`go build ./...`, `go test ./... -count=1 -race`, and `go vet -tags integration ./...` all pass.

## Deliberately left out

- **No integration test** for the fetch wiring — the value is in the pure walk, and the one property the DB path could regress (`LimitPerPK` creeping in) is pinned by a unit guard instead.
- **`--explain`-style row-level drill-down** for this mode. The mismatch detail already names the event, PK and column; a full drill-down is a separate surface.
- **No `driftedEmitted` mirror.** That check is about whether reversal SQL applies to *today's* schema — a schema-drift concern, not image consistency.
- **Not wired into `bintrail status` or the console.** CLI-only for now.
- **Console `/api/verify`** is untouched (still #677).

### DIFF
--- a/docs/verify.md
+++ b/docs/verify.md
@@ -146,6 +146,14 @@ oldest events were walked, so a clean result would be a partial check. A
 mismatch found inside a truncated window *is* still conclusive: the events that
 were walked are real.
 
+> **The `30d` default assumes your window is actually covered.** If partitions
+> older than your retention were rotated out of MySQL and never archived to
+> Parquet, a 30-day lookback covers hours that no longer exist, and *every*
+> table reports `inconclusive` with a coverage-gap reason. That is correct
+> behavior, not a bug — but the fix is to **shorten `--lookback`** to your real
+> retention (or enable archiving), not to ignore the result. Since an
+> all-inconclusive run exits non-zero, a cron gate will tell you immediately.
+
 Memory bound: the walk loads at most `--max-events` events per table (default
 200,000) plus one row of state per distinct primary key seen. The cap is a
 **row count, not a byte budget** — a BLOB/TEXT-heavy table is still large at a
--- a/docs/verify.md
+++ b/docs/verify.md
@@ -150,8 +150,10 @@ Three other conditions also report `inconclusive` rather than risk a false alarm
   continuity record cannot be read at all (one predating those columns) reports
   `inconclusive` for the same reason: *unknown* is not *no gap*;
 - a window that **exceeded `--max-events`** — only its oldest events were
-  walked, so a clean result would be a partial check. A mismatch found inside a
-  truncated window *is* still conclusive: the events that were walked are real.
+  walked, so a clean result would be a partial check.
+
+A mismatch found inside a truncated window *is* still conclusive: the events
+that were walked are real.
 
 **A chain break tells you the chain broke, not why.** A `mismatch` here means
 `row_before` did not match the state the previous event on that key left. Two
--- a/docs/verify.md
+++ b/docs/verify.md
@@ -139,12 +139,54 @@ that is the normal case for any retention window, not corruption. Later events
 on the same key *are* asserted, so a table is still proven as long as at least
 one comparison was made; a table where *none* was reports `inconclusive`.
 
-Two other conditions also report `inconclusive` rather than risk a false alarm:
-a **coverage gap** in the window (a chain with an interior hole cannot be
-asserted against), and a window that **exceeded `--max-events`** — only its
-oldest events were walked, so a clean result would be a partial check. A
-mismatch found inside a truncated window *is* still conclusive: the events that
-were walked are real.
+Three other conditions also report `inconclusive` rather than risk a false alarm:
+
+- a **coverage gap** in the window (a chain with an interior hole cannot be
+  asserted against);
+- a **recorded permanent loss** inside the window — `bintrail status` shows it
+  as `GAP LOST`, and it is stamped in `stream_state.gap_lost_at`. Those events
+  are gone from live MySQL *and* from every archive, so the chains that cross
+  the loss have a hole nothing can be asserted across. An index whose
+  continuity record cannot be read at all (one predating those columns) reports
+  `inconclusive` for the same reason: *unknown* is not *no gap*;
+- a window that **exceeded `--max-events`** — only its oldest events were
+  walked, so a clean result would be a partial check. A mismatch found inside a
+  truncated window *is* still conclusive: the events that were walked are real.
+
+**A chain break tells you the chain broke, not why.** A `mismatch` here means
+`row_before` did not match the state the previous event on that key left. Two
+explanations produce byte-identical evidence, and the check cannot choose
+between them:
+
+1. **The events in between were never captured.** Coverage detection is
+   *partition-existence* based, so a hole *inside* an hour whose partition
+   exists is invisible to it — an `ALTER` without a re-`snapshot` (the indexer
+   logs a column-count warning and *skips that table's events*), a mid-history
+   `--tables`/`--schemas` filter change, a `stream --reset`, or an outage
+   shorter than the pre-created partition horizon.
+2. **The stored before-image is stale or corrupt.**
+
+To tell them apart, check `bintrail status` continuity and the indexer/stream
+logs for skipped tables, filter changes, resets or downtime covering the
+window, and compare against the source's own binlog history. Do not start from
+the assumption that the images are corrupt — a capture hole is at least as
+likely.
+
+**Partial coverage annotates a verdict, it does not erase it.** A table is
+proven as soon as *one* before-image comparison was conclusive; everything that
+stayed unproven is carried as a note on that verdict (visible in `reason`)
+rather than collapsing the table:
+
+| Not checked | Effect |
+|---|---|
+| chains beginning mid-history | noted |
+| a value whose representation could not be normalized (an unmapped `ENUM` ordinal, a JSON document the event image cannot render faithfully) | noted |
+| **drift rows** — events the index holds with no primary key (`pk_values` NULL) | noted; they belong to no chain and are never walked |
+| the entire tail of the window (`--max-events` exceeded) | **collapses** the table to `inconclusive` |
+
+Only a table where *nothing* was conclusive reports `inconclusive`. This
+matters for CI: one unresolvable JSON value in a 200,000-event window used to
+discard every clean assertion beside it and fail the gate on a healthy index.
 
 > **The `30d` default assumes your window is actually covered.** If partitions
 > older than your retention were rotated out of MySQL and never archived to
@@ -168,7 +210,8 @@ Results are **per table**, one of:
 - **inconclusive** — verify could not prove the table either way, and this is
   **never reported as a failure**. Causes include: no predecessor baseline yet,
   the index is behind the baseline anchor, an unsupported primary key, a coverage
-  gap, a digest-contract skew between a pre-pin baseline and the current scan
+  gap, a recorded permanent loss (`gap_lost_at`) inside the window, a
+  digest-contract skew between a pre-pin baseline and the current scan
   (regenerate the baseline — see below), or a value class this version cannot yet
   compare.
 
--- a/docs/verify.md
+++ b/docs/verify.md
@@ -19,14 +19,28 @@ It is read-only and never writes to your source or your index.
   to write rows to the table being compared. A MATCH is strong evidence the
   capture-and-reconstruct pipeline works; it is not tamper-evident or
   forensic-grade proof.
-- **It does not exercise the `recover` path.** `verify` reconstructs full-table
-  state from the *latest* event per PK (`row_after`, `LimitPerPK=1`).
-  `recover`'s reversal SQL additionally depends on `row_before` images,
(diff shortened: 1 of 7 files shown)


nethalo:

### Critical — an INTRA-PARTITION capture hole produces a conclusive false MISMATCH

`internal/verify/recover_inputs.go:97-109` relies entirely on `query.FetchMerged`'s strict-mode `*GapError` to deliver this PR's central promise (a chain with an interior hole cannot be asserted against). That guard is **partition-existence-based**: `buildPlan` (`internal/query/planner.go:142-152`) classifies an hour as covered iff a named `p_YYYYMMDDHH` partition exists. It never checks whether the partition received events.

Every common hole shape leaves partitions fully present:

- The one CLAUDE.md documents as expected behaviour — *"Column count mismatch: indexer logs warning and skips table's events — does not fail."* An ALTER without a re-snapshot silently drops that table's events for as long as the drift lasts.
- A mid-history `--tables` / `--schemas` filter change on `stream`/`index`.
- `stream --reset`, which per #1079 does not even stamp `gap_lost`.
- A daemon outage shorter than the pre-created future-partition horizon (`init --partitions 48` creates 48h forward, `internal/indexer/schema.go:76-80`), so the empty hours still look live.

The verdict produced (`recover_inputs.go:440-441`):

> `row_before does not match the state the previous event on this primary key left (column X); recover would generate reversal SQL from a stale or corrupt before-image`

The images are neither stale nor corrupt — the events between them were never captured. The detail string actively rules out the correct explanation and sends the operator hunting for corruption that isn't there.

Two remedies; the second blocks on its own:

1. Read `stream_state.gap_lost_at`/`gap_lost_detail` (`internal/status/status.go:351`, with the `GapColumnsPresent` legacy fallback at `:358-371`) and degrade to `inconclusive` when a stamped gap falls inside `[Since, Until]`.
2. **Regardless of (1)** — change the detail at `:440-441` so it stops asserting the cause. One line.

### Important

1. **Drift rows collapse into one shared chain.** `recover_inputs.go:289` keys on `ev.PKValues` with no empty-string guard. `internal/query/merge.go` already defends against exactly this (#318: NULL `pk_values` arrive as `""`; `LimitPerPK` gives each its own `\x00drift:<event_id>` bucket). Here two unrelated drift rows become one chain and the second's `row_before` is compared against the first's `row_after` → guaranteed MISMATCH on any index that merely *contains* drift rows. Mirror merge.go: treat `PKValues == ""` as unwalkable.
2. **One unresolvable value discards up to 199,999 clean assertions.** `recover_inputs.go:375-401` — the `unresolved > 0` branch collapses the whole table to `inconclusive` regardless of how many assertions passed. `deferredValueUnresolved`'s `json` case rejects any document containing a numeric literal, so a single differing JSON value nukes the verdict → `VerdictUnproven` → non-zero exit → a permanently red CI gate on a clean index. (The `truncated` branch at `:387-389` is a different, defensible case.)
3. **A legacy index turns a real historical DDL into a conclusive MISMATCH.** `recover_inputs.go:462-470` uses `prevVer != curVer` as the "this is DDL, not corruption" escape hatch, but `SchemaVersion` is `0` for pre-migration data (`internal/query/query.go:233`). On such an index every event carries 0, so an ADD/DROP COLUMN inside the window yields `"qty" is present in one image and absent from the other` as a MISMATCH. Treat `0` on either side as unknown-epoch → `unresolved`.
4. **Fresh-index gap reason is false.** Partitions span from the current hour forward, so on an index younger than `--lookback` every pre-install hour is classified as a gap with the text `query covers hours with no data (rotated and not archived)`. Those hours were never rotated; the index did not exist. Verdict is `unproven` (correct) — the defect is the reason string.

### Suggestions

- `mismatches` (`:281`) is unbounded but only `[0]` and `len()` are consumed (`:380-384`) — a third memory term the PR's bound section doesn't name. `first string` + `count int` closes it.
- `compareImages` returns equal for two empty images (`:457-499`); `checkImagesPresent` tests `nil`, not emptiness. Unreachable in practice (the #493 partial-row-image guard hard-errors on non-FULL `binlog_row_image`), cheap to close.
- `ChainsInconclusive` counts more than `report.go:97-99` and the docs say — it also absorbs chains restarted after a nil-image or TOAST finding (`:305`, `:312`).
- `--explain` is silently ignored under `--check recover`; `--lookback`/`--max-events` silently ignored under `--check content` (`internal/cli/verify.go:143-149`). The PR rejects `--source-dsn` on exactly this reasoning.
- A backfill `bintrail index` after live streaming assigns higher `event_id`s to earlier events — handled by the timestamp key except for events sharing a second.

### Verified as claimed

- **The truncation/prefix argument holds** — traced through `Engine.Fetch`'s `sortResults` after the late-materialization JOIN, `MergeAndTrim`/`MergeResults`, both Parquet builders' `ORDER BY`, and the S3 whole-file path (trimmed globally afterwards). The merged prefix really is a global time-prefix with no interior hole, and `truncated` is exact.
- Exit-code integration is single-path; content-mode JSON byte-identity is pinned by an actual absent-keys assertion, not just claimed.
- The `LimitPerPK` guard exercises the real `recoverInputsFetchOptions`, not a copy.
- **"Both images are normalized" is true**: `MapEventEnumLabels` and `DecodeEventBinaries` both map `RowBefore` *and* `RowAfter`; MySQL and Parquet rows both go through `UnmarshalRowImage` with `UseNumber`, so no `json.Number`/`float64` asymmetry across the archive boundary.
- PK-change detection recomputes both keys from the same event's images rather than trusting stored `pk_values` — the #782 construction.

Issue #833, opened 2026-07-05
cascade: children with a cross-schema FK are silently excluded from the graph (filter by child schema_name, not parent)


nethalo:
`LoadCascadeFKs(ctx, db, []string{rcSchema})` filters `fk_constraints` by `schema_name` (the CHILD's schema) = the parent's schema. A child in schema B with an ON DELETE CASCADE FK to a parent in schema A (legal in MySQL, typical in multi-tenant/reporting setups) isn't loaded: its cascade-deleted rows aren't synthesized, no caveat, exit 0 "complete".

Location: `internal/cli/recover_cascade.go:332`

Same pattern in `internal/console/recover_cascade.go:292` and in `cascadeParentDetect` (which additionally reports `isParent=false`, hiding the auto-detect). The console comment acknowledges it as deliberate scoping but it's not in `docs/` nor surfaced to the operator.

Fix: filter by `referenced_schema_name` (and load the child schemas transitively for multi-level recursion), or at least detect discarded cross-schema edges and emit an Incomplete caveat.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #1116, opened 2026-07-25
cascade: synthesize child UPDATEs to reverse ON UPDATE CASCADE / SET NULL (#1002)

### LINKED ISSUE #1002: recover-cascade: synthesize child UPDATEs to reverse ON UPDATE CASCADE / SET NULL
## Background

Split out from #836 (closed: the docs over-promise was corrected). `docs/indexing.md` and `docs/query-and-recovery.md` now state that `recover-cascade` reconstructs **`ON DELETE CASCADE` / `ON DELETE SET NULL` only**, and that `ON UPDATE CASCADE` / `ON UPDATE SET NULL` are not synthesized. This issue tracks the deferred **code** enhancement to also reverse `ON UPDATE` cascades.

## Problem

InnoDB (through 8.x, and MariaDB) also runs cascaded **updates** of child FK rows when a parent's referenced key is `UPDATE`d — and, like the delete cascades, these child updates are not row-logged as their own events. The current synthesis engine (`internal/cascaderecover`) deliberately gates on `DeleteRule` only. As a result, reverting a parent-key `UPDATE` leaves the child FKs pointing at the new value (dangling) with no warning: the console auto-detect only fires on `rowsContainDeleteOn`, and the CLI only processes `EventDelete`.

## Proposed enhancement

Extend the same synthesize-don't-read technique to `ON UPDATE` cascades: when reverting a parent-key `UPDATE`, synthesize child `UPDATE`s that set the FK column back `new → old` for the affected child rows, gated on the FK's `UpdateRule` (`CASCADE` / `SET NULL`), mirroring the existing `DeleteRule` path. Surface the same auto-detect + `--allow-incomplete` semantics.

## Scope / notes

- Home: `internal/cascaderecover` (+ the console auto-detect and CLI event-type filter).
- Keep the existing correct choice of not conflating delete/update rules — this adds a distinct update path, it does not merge the two.
- As an interim safety net, consider a WARN when a parent-key `UPDATE` is reversed on a table that has `ON UPDATE CASCADE` / `SET NULL` children, so the dangling-child risk is at least signalled until synthesis lands.

nethalo:
Extends `recover-cascade`'s synthesize-don't-read technique to the **ON UPDATE** foreign-key cascades.

InnoDB (≤ MySQL 8.x, and MariaDB) also runs cascaded child `UPDATE`s when a parent's referenced key is `UPDATE`d, and — exactly like the delete cascades — never row-logs them. Reverting a parent-key `UPDATE` therefore left every child FK dangling on the new value **with no warning**: the synthesis gated on `DeleteRule` only, the console auto-detect fired only on `rowsContainDeleteOn`, and the CLI fetched only `EventDelete` roots.

## What changed

- **`internal/cascade`** — `SynthesizeVictims` now dispatches per root `EventType`: a `DELETE` consults `byParentDelete` (delete_rule), a parent key `UPDATE` consults `byParentUpdate` (update_rule). The two rules are still **never conflated** — this is a distinct path, not a merge. Candidate discovery (Phase-1 window, Phase-2 baseline via the #1106 provider, per-parent truncation, the DDL-skew probe, the archives-present gate) is factored into one shared `scanChildren` closure so both paths raise identical coverage caveats.
- **The gate is the changed-key intersection, not "the event is an UPDATE".** `refKeyChanged` compares NULL-ness and then the rendered images, so an `UPDATE` of unrelated columns synthesizes nothing **and** is not reported as a parent to reverse (`Result.KeyUpdateParents`) — `recover-cascade` must not undo column edits the operator never asked about.
- **Both variants handled.** `ON UPDATE CASCADE` → children carry the new key, restore the old one guarded by `AND fk = <new key>`. `ON UPDATE SET NULL` → children are NULL, restore the old key guarded by `AND fk IS NULL`. `recovery.FormatFKCascadeRestore` is the general form; `FormatSetNullRestore` now delegates to it (byte-identical output). CASCADE recurses one level down; `--max-depth` / `--lookback` / the Phase-2 baseline fallback behave exactly as on the delete path.
- **Surfaces updated** — CLI fetches `UPDATE` roots alongside `DELETE`s (two scans, since `query.Options.EventType` is a single type and an all-types fetch would let `INSERT`s eat the `--limit` budget); the console auto-detect now matches the FK rules to the event types actually being reversed (`rowsContainCascadeTriggerOn` + `metadata.CascadeParentRulesInIndex`), and derives its parent set from `baseRows` as before (#772). New counts (`key_restores` / `key_restore_count`) so a script full of FK restorations is never read as "0 rows recovered".

## Never-silent, per this file's existing invariants

- A **repeated parent-key UPDATE** inside the window hides children: the first cascade rewrote them below the binlog, so their last indexed image predates it and the scan for the second update's old key matches nothing. `checkKeyChain` probes for that and flags `INCOMPLETE` rather than report a clean zero.
- `loadCascadeClosure`'s referenced-schema frontier now widens on `update_rule` too — scoping it on `delete_rule` alone would silently lose edges in a multi-level cross-schema ON UPDATE cascade (#833, one rule over).
- `metadata.CascadeParentRulesInIndex` splits the parent probe by action, so a `DELETE` undo never routes on an ON UPDATE-only edge (and vice versa) and surfaces a misleading "0 victims".
- Depth exhaustion, composite FKs, no-baseline, truncation and DDL-skew keep their caveats, reworded to be path-neutral.

## Tests

New `internal/cascade/cascade_update_integration_test.go`: CASCADE-on-update, SET NULL-on-update, an UPDATE that leaves the referenced key alone (synthesizes nothing, and is not reversed), both no-conflation directions, multi-level depth, `--max-depth` exhaustion flagged, the repeated-key-update caveat plus its negative control. Plus `refKeyChanged` and update-rule-frontier unit tests, `FormatFKCascadeRestore` unit tests, and `EmitSQL` golden + all-or-nothing tests for the new section.

**Verified end-to-end against real InnoDB** (`TestKeyUpdateCascade_endToEnd`): the test asserts the cascade is genuinely absent from the binlog, then that `CHECKSUM TABLE` of parent+child after applying the generated script equals the pre-update checksum.

`go build ./...`, `go test ./... -count=1 -race`, `go vet -tags integration ./...` and the integration suites for the touched packages all pass. `recover-cascade` still never executes SQL.

closes #1002

## Follow-up in this PR

The console **frontend** banner was rendering only `victim_count` / `set_null_count`, so an auto-detected recover of a parent-key `UPDATE` would have shown "0 related row(s)" over a script full of FK restorations — the misleading zero the API-side `key_restore_count` exists to prevent, one layer up. `assets/app.js` now renders all three counts and composes the banner sentence from the non-zero parts. README's feature bullet named ON DELETE only.

Checked and left alone: `metadata.buildFKCascadeQuery` (the source-side pre-flight/doctor cascade probe) already matches `DELETE_RULE = 'CASCADE' OR UPDATE_RULE = 'CASCADE'`, so it does not under-advertise what `recover-cascade` can now fix.


nethalo:

Two claims below were verified empirically against MySQL **8.0.46** (scratch DB, created and dropped), not by inspection.

### Critical 1 — a child whose PK contains the FK column gets a self-contradictory restore: silent no-op, reported as a completed recovery

`internal/recovery/recovery.go:1317-1337` (with `internal/cascade/cascade.go:828-833`, `internal/cascaderecover/emit.go:116`).

`FormatFKCascadeRestore` builds the PK predicate from `row[c.Name]` where `row` is `FKKeyRestore.Row` = `cev.RowAfter`, the child's **pre-cascade** image, then appends the guard `fkCol = <cascadedValue>`. When the FK column is itself part of the child's PK — the identifying relationship, and one of the main reasons to declare `ON UPDATE CASCADE` — the same column appears twice with contradictory values:

```sql
UPDATE `db`.`line` SET `pid` = 1 WHERE `pid` = 1 AND `seq` = 1 AND `pid` = 99;
```

Verified on 8.0.46 with `parent(id)` / `line(pid,seq) PRIMARY KEY (pid,seq), FK pid→parent.id ON UPDATE CASCADE`, key moved `1→99`:

```
after cascade:  pid=99 seq=1
rows_the_restore_touched: 0
after restore:  pid=99 seq=1     <-- child still orphaned, parent already back at id=1
```

Meanwhile `SynthesizeVictims` returns `Complete() == true` with one `KeyUpdates` entry, the CLI logs `key_restores=1`, and the console banner says "fixes 1 reference(s)". The operator applies the script, commits cleanly (FK checks are off), and is told the recovery is complete. This is the exact failure class the package's `Incomplete` machinery exists to prevent. `ON UPDATE SET NULL` cannot hit it (a PK column can't be NULL) — only the common variant is affected.

Fixes: build the PK predicate from the *post-cascade* image (the engine already computes it — `nextKeyUpdateItem`, `cascade.go:712-713`), or detect `fkCol ∈ pkCols` and skip with an `Incomplete` caveat.

Discriminating test: rerun `TestKeyUpdateCascade_endToEnd` (`cascade_update_integration_test.go:348`) with the child declared `PRIMARY KEY (pid, seq)` — fails today, must pass after.

### Critical 2 — `checkKeyChain` is blind exactly when the referenced column is the parent's PK, so the never-silent guarantee doesn't hold for `REFERENCES parent(id)`

`internal/cascade/cascade.go:639-679`, specifically `PKValues: pev.PKValues` at `:649`.

The probe scopes the search for an earlier key-changing UPDATE to the root's own `pk_values`. But `internal/parser/parser.go:692` stores `PKValues: BuildPKValues(pkCols, before)` — the **before-image** PK. So when the referenced column *is* the parent's PK, every link lands under a different `pk_values`:

- `parent.id: A → B` at t1 → event `pk_values = "A"`
- `parent.id: B → C` at t2 → event `pk_values = "B"` (the root)

The probe queries `"B"` and never sees t1. `keyChainChecked` memoizes, no caveat is raised, the candidate scan for old key `B` matches nothing (children's last indexed image still carries `A`), and the run reports **0 restores, `Complete() == true`** — the silent zero this PR claims to have eliminated.

Trigger: a window-scoped run (`--since/--until` without `--pk`, or `--pks 1,99`) over a parent whose PK moved twice inside the lookback. Note `--pk 1` does **not** reproduce it — only the first update is fetched and no chain forms.

`TestSynthesizeKeyUpdate_repeatedKeyUpdateFlagged` (`:283`) passes only because it references a non-PK column `code`, so `pk_values` stays `"1"` across both links. Give the A→B event `PKValues: "1"` and the root `PKValues: "99"` and it fails.

This also makes **`docs/query-and-recovery.md:328-336` over-promise**: it states the repeated-key-update case is flagged `INCOMPLETE RECOVERY` and that "a zero-child result there is not proof there were no children". That guarantee does not survive `REFERENCES parent(id)`.

Suggested fix, immune to `pk_values` moving: probe by referenced-column **value** — fetch parent-table UPDATEs in the window with `ColumnEq{fk.ReferencedColumn: parentOldKey}` and flag when any has `row_after[refCol] == parentOldKey` and `row_before[refCol] != parentOldKey`.

### Important

1. **Mixing DELETE and UPDATE roots destroys the ordering `GenerateSQLFromRows` depends on.** It has no sort — it relies on input order and does `slices.Reverse` (`recovery.go:228`). The engine gets this right (`GroupParentDeletesByFKGraph` sorts roots ascending, correctly interleaved), and both call sites throw it away by re-segregating: `internal/cli/recover_cascade.go:317-321` and `internal/console/recover_cascade.go:170-175`. Input that breaks: parent P key-UPDATEd at t1, DELETEd at t2, both in window → the UPDATE-undo runs first against a row that doesn't exist yet (0 rows), then the INSERT restores the post-update image. Fix: stable merge by `(EventTimestamp, EventID)` at both sites. The console auto-detect path is unaffected and its comment is right.
2. **Auto-detect wraps ordinary UPDATE undos in `SET FOREIGN_KEY_CHECKS=0/1` and claims CASCADE when nothing cascaded.** `rowsContainCascadeTriggerOn`'s UPDATE arm is deliberately coarse, so any UPDATE undo on a table with an ON UPDATE child routes through `cascadeRecover` → `EmitSQL`. When `refKeyChanged` rejects it: `cascade_detected: true`, all counts zero, banner "no related rows needed repairing", and FK validation silently disabled on a script the operator expected checked. Fix: when `len(victims)+len(setNull)+len(keyUpdates) == 0`, fall back to the plain script and response.
3. **The delete-path analog of Critical 2 is still silent** (`cascade.go:883-899`): parent key-UPDATEd at t1 then DELETEd at t2 → the delete-cascade scan uses the pre-delete key while children carry the pre-update key → zero victims, `Complete() == true`, no caveat. Pre-existing in kind, but newly *detectable* now that `update_rule` is in the graph.
4. **`keyChainProbeLimit` truncation isn't reported.** `cascade.go:267, 646-655` fetches the 8 newest UPDATEs and returns silently. A parent with >8 unrelated column updates after the earlier key change hides the chain with no caveat — the "bounded probe, unbounded conclusion" shape that `skewSampleLimit` carefully avoids. Flag `Incomplete` when `len(prior) == keyChainProbeLimit` and no chain was found.

### Suggestions

- Empty script + exit 0 + no explanation when a parent UPDATE touched no referenced key (`internal/cli/recover_cascade.go:205-235`). Now that the command advertises parent UPDATEs, say so.
- Stale comments still saying "ON DELETE CASCADE / SET NULL": `internal/console/recover_cascade.go:26`, `:78`, `internal/metadata/metadata.go:1421`.
- `buildFKCascadeQuery` (`metadata.go:1270-1272`) matches `DELETE_RULE = 'CASCADE' OR UPDATE_RULE = 'CASCADE'`, so `doctor`/pre-flight silently miss both `SET NULL` variants — pre-existing, worth aligning with the `CascadeConstraintsInIndex` change made here.
- `cascade.go` is now 1480 lines; the shared-scan refactor is a natural seam for splitting the update path out.

### Verified as claimed

- **`SET FOREIGN_KEY_CHECKS=0` does NOT re-run an `ON UPDATE CASCADE`** (probed on 8.0.46) — the emitted child restores are load-bearing, which is what makes the E2E test a real assertion.
- **The no-conflation invariant holds** — traced `byParentDelete`/`byParentUpdate` through the initial build, both dispatches, the recursion, `loadCascadeClosure`'s frontier (over-inclusive only, harmless) and `CascadeParentRulesInIndex`.
- **The guards cannot over-reach.** Every restore is PK-anchored, so `AND fk = <new key>` / `AND fk IS NULL` can only decline to touch the identified row, never widen the blast radius. `QuoteName`/`FormatSQLValue` applied to every identifier and value on both sides.
- `refKeyChanged` compares NULL-ness *before* rendering — the one ordering that matters given `valToString(nil) == valToString("")` — and pins the `NULL → ""` case.
- `EmitSQL` pre-builds every statement before writing a byte, preserving all-or-nothing across the new block.
- `TestKeyUpdateCascade_endToEnd` asserts the cascade is absent from the binlog *before* asserting the checksum — the premise is guarded, not assumed.

Issue #1117, opened 2026-07-25
MariaDB 11.4+: LogPos=0 corrupts start_pos/end_pos and makes the resume dedup delete every indexed row


nethalo:
## Summary

On a **MariaDB 11.4+** source, every intra-transaction row event reaches the
parser with `LogPos = 0`. `binlog_events` therefore stores:

```
start_pos = 18446744073709551574   (2^64 - EventSize, an underflow)
end_pos   = 0
```

for **every** captured row. The resume-time dedup then matches all of them and
deletes them.

## Impact — data loss on every resume

`deleteEventsSinceCheckpoint` cuts on `binlog_file = ? AND start_pos >= ?`.
Because the stored `start_pos` is ~1.8e19, that predicate is true for every
MariaDB row in the checkpoint's binlog file. On **every** stream restart
(`bintrail stream`, `bintrail up`, a supervisor restart under
`bintrail-console watch`), the whole file's worth of already-indexed rows is
deleted — including rows the source will never re-send.

Both resume modes are affected: position mode (`deleteEventsSinceCheckpoint`)
and GTID mode (`deleteEventsSinceCheckpointGTID`, which delegates the positional
cut to the same function).

Beyond the dedup, the corrupt `start_pos`/`end_pos` values are wrong wherever
they are read — forensics, the `UntilPos` cut in `internal/query`, and anything
that orders or bounds by binlog offset.

## Reproduction

Observed while writing the #500 acceptance tests, against
`mariadb:11.4` started with the CI flags
(`--log-bin=mariadb-bin --binlog-format=ROW --binlog-row-image=FULL
--server-id=2 --gtid-strict-mode=ON`), indexing into MySQL 8.0.46:

```
durable checkpoint: file="mariadb-bin.000213" pos=11149 gtid="0-2-1401" mode="gtid"
row pk=1 file=mariadb-bin.000213 start=18446744073709551574 end=0 gtid=0-2-1397
row pk=2 file=mariadb-bin.000213 start=18446744073709551574 end=0 gtid=0-2-1398
row pk=3 file=mariadb-bin.000213 start=18446744073709551574 end=0 gtid=0-2-1399
row pk=4 file=mariadb-bin.000213 start=18446744073709551574 end=0 gtid=0-2-1400
row pk=5 file=mariadb-bin.000213 start=18446744073709551574 end=0 gtid=0-2-1401
```

The next resume logged `rows_deleted=5` and the index came back empty.

[...]

PR #1118, opened 2026-07-25
stream: pin exactly-once resume with a real crash→restart test (#500)

### LINKED ISSUE #500: Exactly-once stream indexing: dedup re-streamed transactions after an ungraceful crash
## Summary

`bintrail stream` is **at-least-once**: after #491 (PR #498) the durable GTID checkpoint advances only at transaction commit boundaries, so no transaction is ever lost. But on an **ungraceful crash** (kill -9 / OOM / power loss), transactions that were flushed to `binlog_events` but not yet checkpoint-saved are re-streamed on restart, producing **duplicate rows**. The duplicates get new auto-increment `event_id`s, so `query.MergeResults` (which dedups by `event_id`) does **not** collapse them — inflating `query` counts and causing `recover` to emit duplicate reversal statements.

This issue tracks moving from at-least-once to **exactly-once**.

## Why it's not trivial here

1. **GTID resume is transaction-atomic.** MySQL GTID auto-position resends whole transactions; you can't resume mid-transaction. So true no-dup requires either (a) buffering whole transactions before checkpointing them — which risks unbounded memory for a single huge transaction (bulk load), since the current code flushes mid-transaction precisely to bound memory; or (b) restart-time cleanup of partial-transaction rows.
2. **`binlog_events` is range-partitioned** by `TO_SECONDS(event_timestamp)`. The idempotent-insert alternative (a unique key on the event's binlog identity + `INSERT ... ON DUPLICATE KEY`/`INSERT IGNORE`) collides with MySQL's rule that a unique key on a partitioned table must include the partition column (`event_timestamp`), making the dedup key awkward.

## Options to evaluate

- **Transaction-aware batching + atomic flush+checkpoint:** flush the batch and UPSERT `stream_state` in one DB transaction, only at commit boundaries; batch-size becomes a soft trigger (flush at the next commit after it fills). Exactly-once except a single transaction larger than the in-memory batch buffers fully — document/cap that.
- **Idempotent inserts:** add a unique key over the event's stable binlog identity (e.g. `(server_id, binlog_file, end_pos, row_ordinal)`, including `event_timestamp` for the partition constraint) and use `INSERT IGNORE`. Bounded memory, no batching redesign, but a schema migration + `EnsureSchema` change.
- **Restart cleanup:** on restart, delete `binlog_events` rows whose GTID is not in the resumed `gtid_executed` set before streaming.

## Acceptance
- No duplicate `binlog_events` rows after an ungraceful crash + restart (covered by a kill-mid-stream integration test).
- Memory behavior under a large single transaction is documented/bounded.

## References
- Follows #491 / PR #498 (which converts silent loss → at-least-once). The duplicate-not-deduped-by-MergeResults detail is the concrete downstream cost.

nethalo:
The dedup **code** for #500 landed in #874. This PR adds what the issue's own
acceptance criteria still lacked: real-database regression tests, and an
explicit decision on the one path where dedup is deliberately skipped.

## 1. Kill-mid-stream, restart, no duplicates, no loss

`TestIntegrationCrashRestartExactlyOnce_position` runs three phases through the
same `streamrun.One()` entry point the CLI uses:

| phase | what happens |
|---|---|
| run 1 — clean | stream pks 1-5 into the partitioned `binlog_events`, checkpoint durably |
| run 2 — crash | index pks 6-10 while **every** checkpoint write fails, so the durable checkpoint stays at run 1's and the new rows sit *beyond* it |
| run 3 — restart | `One()` must delete those stragglers before `StartSync` and let replication re-deliver them |

Final assertion is per-PK (`GROUP BY pk_values`), not a total: a bare `COUNT(*)`
cannot tell one duplicate plus one lost row from a clean run. It waits on pk 12
— written only during run 3 — so the sample can't be taken before the dedup ran.

**How the crash is staged.** A crash is only observable to the next process
through the database: rows flushed, checkpoint stale. Cancelling the context
isn't enough — `streamLoop`'s `ctx.Done` branch flushes *and* checkpoints, which
is a graceful shutdown. So run 2 installs a `BEFORE UPDATE` trigger on
`stream_state` that `SIGNAL`s: every `saveCheckpoint` fails (warn-only by
design, see `TestStreamLoop_saveCheckpointFailureDoesNotAbort`) while the
previously saved row survives untouched. Only production code paths, and
deterministic — unlike a timing-based kill.

**Verified as a guard, not a description.** Disabling the position-mode dedup
branch makes it fail with `pk 6: DUPLICATED … pk 10: DUPLICATED`.

## 2. The `start_pos >= pos` boundary, pinned against a real DB

`TestIntegrationDedupBoundaryIsEventStart` (+ its `…GTID` sibling) constructs
two adjacent events straddling the checkpoint offset. The invariant is an
equality — delete-set ≡ resend-set — and **both** edges are asserted:

- the row **ending** at `P` (`start_pos < P`) is never re-sent, so it must
  **survive**; widening the cut silently destroys data,
- the row **starting** at `P` is re-sent, so it must be **deleted**; narrowing
  `>=` to `>` silently duplicates it.

Flipping `>=` to `>` fails the test. The GTID sibling additionally pins the
open-transaction sweep: a row below `P` whose GTID the saved set does *not*
contain must be deleted; one it *does* contain must survive.

## 3. GTID-mode coverage — and an honest limit

`TestIntegrationResumeGTIDDedupsBeforeStartSync` pins, against a real database:
mode selection from a real `stream_state` row, the real
`deleteEventsSinceCheckpointGTID` against the partitioned table, and the
**ordering** (dedup runs before `StartSyncGTID`). Also verified by mutation.

It is **not** a full crash→restart→re-stream cycle, because no GTID-capable
source is reachable from this suite:

- the MySQL container runs `gtid_mode=OFF` both locally and in the CI
  integration matrix, and flipping that global mid-suite would perturb every
  other test sharing the server;
- MariaDB (always GTID-capable) is disqualified by a separate bug found while
  writing these tests — see below.

## Decision on the GTID gap-auto-advance window: **document, don't close**

After an unfillable-gap auto-advance in GTID mode, dedup is skipped
(`streamrun.go`). Closing that window would mean deleting on the *pre-advance*
checkpoint coordinates, which sit below a purge floor the code cannot express as
a `(file, pos)` cut — destroying rows the source has already purged and can
never re-send. The asymmetry decides it: **a duplicate is visible and
repairable; a deleted pre-floor row is the only surviving copy.** Same
reversibility logic as the architectural red lines.

It is now explicit in all three places: the code comment, the runtime
`slog.Warn`, and a new *"Known and accepted: duplicate rows after a GTID-mode
auto-advance"* subsection in `docs/streaming.md` (what the operator sees, why it
exists, and that a re-baseline or `--no-gap-fill` is the way out).

## Also

- CI false-green tripwire for `TestIntegrationCrashRestartExactlyOnce_position`,
  matching the existing e2e / caching_sha2 tripwires.
- The existing sqlmock unit tests are **unchanged** — this is additional
  coverage, not a replacement.

## Found while writing these tests (filed separately)

On MariaDB 11.4+, intra-transaction row events arrive with `LogPos=0`, so
`binlog_events.start_pos` lands as `18446744073709551574` (2^64 − EventSize) and
`end_pos` as `0`. The resume dedup then matches **every** row in the
checkpoint's binlog file and deletes it. Filed as #1117 — it needs its own
fix (`FillZeroLogPos` plus a matching change to the wraparound guard, which the
naive one-line enable trips) and a MariaDB regression baseline, so it is
deliberately out of scope here.

closes #500


nethalo:

Since this PR is almost entirely a test, the review was done by **mutating the production code** in a throwaway copy and checking the tests fail when they should — not by inspection.

| mutation | result |
|---|---|
| disable the position dedup branch (`streamrun.go:1980`) | crash test FAILS: `pk 6..10: DUPLICATED — expected exactly 1 indexed row, got 2` |
| `start_pos >= ?` → `> ?` (`streamrun.go:205`) | boundary FAILS on 2 of 3 subtests, GTID boundary FAILS, `…ResumeGTIDDedupsBeforeStartSync` FAILS |
| widen the cut (`start_pos >= ? - 100000`, over-delete) | crash test FAILS **and** all three boundary subtests FAIL |

Under-delete, over-delete and off-by-one all caught. The third is the one the PR body doesn't claim and it matters: it proves the **no-loss** half is real, not just the no-duplicate half. Also ran the package `-tags integration -race` (PASS, 20.8s) and the crash test `-count=5` (stable).

**No critical findings.** The crash simulation is faithful, assertions are genuinely per-PK, the tripwire is fail-closed, and the MariaDB job's `-run` filter excludes all four new tests so the `LogPos=0` bug filed as #1117 cannot bite them.

### Important

1. **Leftover DEBUG scaffolding with a nil-deref path** — `crash_restart_integration_test.go:321-331`. `Query`'s error is discarded, so on any failure `rs` is nil and `rs.Next()` panics, turning a diagnosable DB error into a stack trace; `rs.Scan`'s error is dropped too, so a scan failure prints zero-valued garbage as if it were data. It also emits `t.Logf` on every CI run with no assertion attached. Delete it, or gate it behind `t.Failed()` in a `t.Cleanup` and check both errors.

2. **The GTID test's stated ordering mechanism is factually wrong** — `:398-402`:

   > *"the ORDERING: dedup runs BEFORE StartSyncGTID. With gtid_mode=OFF the sync attempt fails, and the rows are already gone when it does…"*

   Instrumented, that is not what happens. `StartSyncGTID` **succeeds**: the log shows `begin to sync binlog from GTID set`, `Connected to server flavor=mysql version=8.0.46`, `Streaming started`, then `checkpoint saved file="" pos=0` — `streamLoop`'s ticker, which only runs after `One` is fully past syncer setup. go-mysql sends `COM_BINLOG_DUMP_GTID` and returns a streamer without reading the reply; the 1236 arrives asynchronously on the first `GetEvent`.

   The conclusion survives for a different reason: what pins the GTID branch is the **surviving-row set**. Had `One` taken the position branch, `deleteEventsSinceCheckpoint` would have removed `beyond-checkpoint` but left `open-transaction` (start_pos 3000 < 5000) alive, and `assertPKs` at `:487` would fail. That's a real and sufficient proof — it just isn't the one written down. In a file whose entire value is the accuracy of its claims, a wrong mechanism in the comment *is* the defect.

   Same block: the `strings.Contains(…, "GTID")` check at `:477` is described as proving the GTID branch was selected. It doesn't on its own — `detectGTIDGap` (`streamrun.go:978-1011`) returns several GTID-flavoured errors from *before* the dedup. Safe (such an error leaves rows undeleted and `assertPKs` fails), but the comment overstates what that assertion carries.

### Suggestions

- `survivingPKs` (`:506-525`) is unscoped, and because `StartSyncGTID` genuinely succeeds a real stream attaches to that index DB for up to 20s with `Checkpoint: 1`. Safe only because the source is fresh and no DML runs; any future change putting rows there turns `:487` into a flake. One `WHERE schema_name = ?` fixes it.
- `assertExactlyOnce` (`:199-223`) groups by `pk_values` alone, so "one row per PK" is right only because every source row is INSERT-only and touched once. The first person to add an `UPDATE` gets a false `DUPLICATED`. Worth one line in the doc comment.
- The tripwire guards only `…_position` (`ci.yml:213`). Defensible — build-tag/path drift drops all four at once and the position test is the only one with an extra skip that could fire independently — but adding the other three names is two cheap lines.
- `testStreamDeps()` (`:96-109`) is currently field-for-field identical to `streamdeps.Default()` and the import-cycle rationale is correct. Note the asymmetry: a *new* `Deps` field fails closed via `validate()`, but swapping an *implementation* drifts silently.
- `docs/streaming.md:164-166` quotes the WARN without its `; see docs/streaming.md` tail — a prefix, so grep still works, but an exact-match log search won't hit.

### Strengths

- **This is the first test in the repo that drives `streamrun.One` end-to-end against a live MySQL source.** Every other `One(` caller in a test is `pgstreamrun`. The MySQL streaming entry point had zero real-source coverage before this — the exact gap #500 named.
- **The crash staging is faithful on every axis that matters**, checked one by one: `checkpoint()` flushes first and only then saves, so rows land and the checkpoint doesn't; `saveCheckpoint` failure is warn-only with no retry/escalation/budget and feeds nothing back into stream behaviour; `serverID` is written but never compared on resume. And the header's core claim is right — `streamLoop`'s `ctx.Done` branch calls `checkpoint()`, so cancellation alone is graceful and would prove nothing. The trigger is the right tool.
- It also **fails closed**: if a future change made `One` write `stream_state` during a plain resume, that write would error under the trigger and `One` would return non-nil → `t.Fatalf`. It cannot silently mask a new write path.
- **The race-free sampling argument at `:358-362` is sound and load-bearing** — waiting on pk 12 (written only in run 3, therefore sequentially after 6-10) is what stops the final assertion from being taken before the dedup ran. Waiting on "10 rows present" would have passed trivially.
- **Preconditions are asserted, not assumed** (`:314`, `:317`, `:344-350`) — without them the test could pass vacuously; with them it can't.
- The GTID scope caveat is stated honestly in the PR body, the file header and the test's doc comment, and the test *name* claims only what it proves.

### On the GTID-window decision: endorsed, and provable from the code

At `streamrun.go:1867-1886` the auto-advance updates `startGTIDStr`/`accGTID` and sets `gtidAdvanced` but leaves `startFile`/`startPos` **untouched**; `persistGapAutoAdvance` then writes the advanced GTID set alongside the stale `(file, pos)`. That is the concrete demonstration that no post-advance binlog coordinate exists to key a delete on — contrast the position branch at `:1858-1865`, which *does* reassign them, which is exactly why the dedup still runs there.

Running it anyway would be worse than the PR argues: `deleteEventsSinceCheckpointGTID` (`:230-280`) also sweeps rows *below* the cut whose GTID isn't in the saved set — and after an advance that set is `@@gtid_purged`, which by construction does not contain recently-captured surviving GTIDs. Both halves of the delete would destroy irreplaceable rows.

Disclosure is adequate for a stronger reason than "there's a warn and a doc": the auto-advance is already durably stamped (`gap_lost_at`/`gap_lost_detail`) and surfaced by `bintrail status` with a `⚠ GAP LOST` banner and `--fail-on-gap`, so an operator can discover it after the fact from persistent state, not only from a log line that scrolled away.

Issue #949, opened 2026-07-10
CI: MySQL integration matrix (8.0/8.4) can go green-via-skip — no BINTRAIL_REQUIRE_MYSQL guard/tripwire, unlike MariaDB/PG


nethalo:
`SkipIfNoMySQL` (internal/testutil/testutil.go:42-52) unconditionally `t.Skipf` on any dial/ping failure. ci.yml:109-167 (integration 8.0/8.4) sets only `BINTRAIL_TEST_DSN` — no equivalent of `BINTRAIL_REQUIRE_MARIADB` (:184-189) or `BINTRAIL_REQUIRE_POSTGRES` + the `grep '--- PASS'` tripwire (:267-343), both added expressly because a build-tag/path drift "would still exit 0 green with ZERO coverage."

The flagship MySQL job — the required check proving the whole capture->index->query->recover pipeline and the caching_sha2 seam — is the one left unguarded.

Fix: add `BINTRAIL_REQUIRE_MYSQL=1` to the integration job, honor it in `SkipIfNoMySQL` (`t.Fatal` not `t.Skip`, mirroring postgres.go:29-36), and/or add the same `--- PASS: TestEndToEnd` tripwire the PG job uses.

---
_Surfaced by the 2026-07-10 multi-persona team review; verified against `main`._

PR #1119, opened 2026-07-25
audit: wire ext.AuditSink at every historical-read surface, pinned by contract tests

### LINKED ISSUE #945: ext.AuditSink seam invoked only from CLI query/recover + MCP — shim/reconstruct/recover-cascade/verify --explain/who-changed emit nothing; no contract test
The only non-test `ext.Record` sites are internal/cli/query.go:683, internal/cli/recover.go:408, cmd/bintrail-mcp/main.go:427,534. Zero calls from:
- internal/shim/* (network `_flashback`/`_snapshot` time-travel, which has per-tenant authenticated users to attribute, internal/shim/auth.go:36)
- internal/cli/reconstruct.go (full-row PITR + `--history`)
- internal/cli/recover_cascade.go (emits reversal SQL — the exact class recover exists to record)
- internal/cli/verify.go `--explain`, internal/cli/whochanged.go, internal/console/*

The seam's own contract (ext/audit.go:8-16) names "historical data access (query, shim time-travel, reconstruct)" and enumerates `shim`/`console` as Surfaces — the wiring contradicts it. No test pins which surfaces must emit, so a refactor can silently drop emission with CI green.

Fix: add `ext.Record` at reconstruct, recover-cascade, verify(`--explain`), the shim query handler (actor = authenticated tenant), who-changed, and console; add a table-driven test asserting the exact audited-surface set (mirror uifree_test.go). Correct the ext/audit.go docstring for any surface deliberately out of scope. (Console coverage depends on the app-entrypoint export issue.)

---

nethalo:
## What

`ext.AuditSink` only ever fired from the CLI's `query`/`recover` and the MCP tools, while `ext/audit.go`'s own contract named the shim, reconstruct and the console as audited surfaces. Nothing pinned the wiring, so an emission could be dropped in a refactor with CI green.

This wires the missing surfaces and — the main deliverable — pins the exact audited-surface set with **behavioural** contract tests.

## Surfaces wired

| Surface | Action | Where |
|---|---|---|
| `cli` | `reconstruct.run` | `internal/cli/reconstruct.go` — single-row, `--history`, full-table (one event per table), `--sql` |
| `cli` | `recover.cascade` | `internal/cli/recover_cascade.go` — both text and JSON modes, before `cascadeExit` (the script is durable even when the recovery is partial) |
| `cli` | `verify.explain` | `internal/cli/verify.go` — one event per drilled-down table |
| `shim` | `timetravel.query` | `internal/shim/handler.go` `HandleQuery`, covering `_flashback` / `_snapshot` / `_diff`, single-row and full-table |
| `console` | `query.run`, `recover.generate`, `recover.cascade`, `reconstruct.run`, `verify.explain` | `internal/console/{api,reconstruct,recover_cascade,verify_trigger}.go` |

Already wired before this PR (kept, now pinned): `cli/query.run`, `cli/recover.generate`, `mcp/query.run`, `mcp/recover.generate`, `console/authz.denied` and `console/profile.denied` — the two console refusals, which are not data reads. `profile.denied` was live but undeclared; the contract now lists it and a console unit case drives a profiled session into the cascade gate.

**Shim actor = the authenticated tenant.** New `Handler.BindActor(user)`, bound by both serving layers: `internal/cli/shim.go` (standalone) and `consoleapp/flashback.go` (the console's flashback port, which routes *by* username). An unbound handler reports `mysql:unbound` rather than `""`, so a serving layer that forgets the bind is visible in the trail instead of indistinguishable from a real event. Console actor = the session's verified identity, or a `token` sentinel for static-token requests — never `ext.ProcessActor` (a daemon's OS owner says nothing about who asked).

## Contract test

`internal/audittest` holds the canonical `Required` set of `(surface, action)` pairs plus the recording sink. One table-driven test per surface exercises the **real code path** with that sink installed and calls `audittest.CheckCoverage`, which fails on a *missing* emission and on an *undeclared* one:

- `internal/shim/audit_contract_test.go` (unit, sqlmock index)
- `internal/mcptools/audit_contract_test.go` (unit, sqlmock index)
- `internal/console/audit_contract_test.go` (unit) + `audit_contract_integration_test.go` (reconstruct + cascade)
- `internal/cli/audit_contract_integration_test.go` (all five CLI commands, reusing the existing seeds)

Behavioural, not source-level, on purpose: a grep or an AST walk for the call would still pass if a refactor left the call on a branch the command never takes. Verified by deleting the console `query.run` emission — the test fails with `must emit console/query.run, but exercising it recorded nothing`. The CLI/console-integration tiers run under `-tags integration`, which CI runs on **every pull request** as a required check.

Extra pins: refused requests/queries emit nothing (a refusal read no rows); the shim's `AllocsPerRun` check keeps the no-sink hot path at zero allocations.

## Failure semantics and hot path

`ext.Record` and `AuditSink.Record` both return nothing, so a sink cannot fail a user's query by construction — emissions sit on the success path, after the artifact is produced. That matches the three pre-existing sites (they don't handle sink errors because there are none to handle).

For the shim — a network daemon emitting once per client round trip — the new `ext.Auditing()` guards event construction, so the OSS build (no sink) pays a nil check and never allocates the `Detail` map.

## Deliberately out of scope (now stated in `ext/audit.go`, so contract and wiring agree)

- Metadata-only reads that return no row images: the shim's `SHOW TABLES FROM <virtual schema>`, and the console's status/schemas/capabilities/storage/baselines endpoints.
- `verify` **without** `--explain`: it compares fingerprints and reports per-table verdicts; only the `--explain` drill-down prints rows.
- The capture plane (index/snapshot/stream/agent/rotate/archive): those write the index, they don't read historical rows back out.
- `bintrail-pg` reuses `internal/cli`, so it reports `Surface: "cli"` — no new surface invented.

## One behavioral change outside the issue's scope

`recordConsoleDeny` now takes its actor from the new `consoleActor(r)` instead of `identityFrom(r.Context())`, so a **token-authenticated** denial records `"token"` where it used to record `""`. Deliberate: an empty actor reads like a session whose identity went missing, and the console's data-read emissions needed the same fallback. Session-authenticated denials are unchanged (the verified login).

## Notes on the issue's line references (stale since 2026-07-10)

- The MCP tools moved to `internal/mcptools`, shared by standalone `bintrail-mcp` and the console's `/mcp` endpoint (which tags `Surface: "console"` via `Config.AuditSurface`).
- `internal/cli/whochanged.go` no longer exists in the core — who-changed moved to the EE wrapper (`internal/console/forensics_removed_test.go` is the guard), so it is not wired here.
- The console did have one `ext.Record` (an authz **denial**, not a data read), which is why "zero calls from internal/console/*" read as true in spirit.
- Console needed no `consoleapp` plumbing: the sink is a process global, and EE's `main` installs it before calling `consoleapp.Main`.

## Testing

- `go build ./...`, `go vet ./...`, `go vet -tags integration ./...` clean
- `go test ./... -count=1 -race` green
- `go test -tags integration ./internal/{cli,console,shim,mcptools} ./consoleapp -count=1` green against MySQL 8.0 on :13306

closes #945


nethalo:

**No critical findings.** The wiring is correct where it exists, every emission is on the success path outside locks and transactions, and the contract test is a real contract rather than a tautology. The problems are two unwired surfaces that reproduce the exact contract/wiring disagreement #945 was filed about, plus one panic-safety claim the seam doesn't deliver.

### Important

1. **`bintrail-pg flashback` is a network time-travel surface that emits nothing — and is not declared out of scope.** `internal/pgshim/session.go:87-123` never calls `Handler.HandleQuery`; it calls `PKColumnCheck` → `ResolveSnapshotRow`/`ResolveFlashbackRow` → `ColumnsFor` directly, so the new emission at `internal/shim/handler.go:467-483` is unreachable from it. `internal/pgshim/pgshim.go:196-207` authenticates a real per-user credential and never calls `BindActor`.

   Trigger: `psql -h host -p 5433 -U tenant_a -c "SELECT * FROM _flashback.orders AS OF '2026-05-23 18:20:13' WHERE id = 1"` against `bintrail-pg flashback` returns a full row image and records zero audit events.

   `ext/audit.go:24-27` now states the shim emits "from the standalone `bintrail shim` AND the console's embedded flashback port" — an enumeration that silently omits the third serving layer, which exists in the core and returns row images. Either wire it or name it in the "Deliberately NOT audited" block. The PR's whole thesis is that the docstring and the wiring must agree.

2. **`reconstruct --baseline-only` reads and prints baseline row images with no emission.** `internal/cli/reconstruct.go:255-268` returns after `writeReconstructOutput` — before the `auditReconstruct` call at `:478`, whose own docstring (`:723-733`) claims it covers "the four shapes the one command serves". `--baseline-only` is a fifth, and it prints the raw row straight out of the Parquet snapshot.

   The contract test cannot catch this: `CheckCoverage` requires *at least one* `cli/reconstruct.run` and the test case supplies it via the single-row path. That is the generalization limit of "I verified it bites by deleting one emission" — deleting the only emission for a pair bites; adding a fifth unaudited mode to a pair that already has four audited ones does not.

3. **`ext.Record` is not panic-safe, and the new docstring says it is.** `ext/audit.go:88-96` calls `sink.Record` with no `recover()`, while `:73-77` claims "a sink cannot fail a user's query by construction". Returning no error prevents error *propagation*, not a panic, a block, or an `os.Exit` — and an EE sink is third-party code. Consequences, all after the artifact was produced:
   - CLI (`internal/cli/recover_cascade.go:787`): the reversal script is already written to `--output`, then the process panics and exits non-zero. A cron/CI wrapper gating on exit code concludes the recovery failed and may re-run.
   - Shim (`internal/shim/handler.go:468`): the panic unwinds through go-mysql's per-connection goroutine — best case the client's connection dies mid-resultset, worst case it takes the daemon.
   - Console (`internal/console/api.go:256`): `net/http` recovers, but the 200 body is already flushed, so the client sees a truncated body on a request that succeeded.

   One line in the seam covers all nine new call sites plus the four pre-existing: `defer func() { _ = recover() }()` in `Record`. Given the PR's own framing, worth doing here.

4. **`consoleActor` misattributes an identity-less session as `"token"`.** `internal/console/audit.go:19-25` decides by testing whether `identityFrom(ctx)` is empty, but the context already carries an unambiguous discriminator — `authKindFrom` (`internal/console/auth.go:56-58`, set at `:110`/`:114`). `identityCtxKey`'s docstring says the identity is `""` for "sessions minted with no identity", so such a request is now recorded as if the shared automation token was used.

   That population is exactly the one hitting `recordConsoleDeny`: `authzMiddleware` (`internal/console/authz.go:163-166`) returns early for a nil policy, so **every** `authz.denied`/`profile.denied` comes from a policy-carrying (EE session) request. Keying on `authKind` costs nothing; a third sentinel (`"session:unidentified"`) would mirror the `mysql:unbound` reasoning applied correctly on the shim side.

   This also answers the `""` → `"token"` question: safe for a sink treating the actor as an opaque string, but a *semantic* regression for one distinguishing "identity unknown" from "shared token". The old `""` was honest about not knowing.

5. **`BindActor(user)` on the console flashback port records a routing key, not a principal.** `consoleapp/flashback.go:246-252`, vs the `ext/audit.go:26-27` claim "Actor is the authenticated MySQL user". On this port `flashbackCreds` accepts *any* username and authenticates on the shared console token alone; the username is then resolved as a server selector. So Actor is the target server's registry id/name — identical for every human holding the token, and already present in the audit trail's server dimension. Not a hole (values are constrained to registry names), but a sink treating Actor as a principal will attribute every console-flashback read to a database name. Prefix it (`server:<name>`) or state the limitation. The standalone shim is fine — there `GetUser()` is a per-tenant credential.

6. **Emissions use cancelable contexts, so a client disconnect drops the record for a read that completed.** `internal/console/audit.go:39` passes `r.Context()`; `internal/shim/handler.go:511-514` uses `h.baseCtx`, cancelled on disconnect and SIGTERM. In both cases the rows are already fetched and serialized when `Record` fires. A sink doing ctx-aware I/O — the realistic EE implementation — drops precisely the records for aborted-mid-response reads, the population an auditor most wants. `context.WithoutCancel(...)` preserves trace/tenant values without the cancellation.

### Suggestions

- **`CheckCoverage`'s "undeclared emission" arm is weaker than its docstring claims** (`internal/audittest/audittest.go:376-415`): it fires only for a pair no owner declares, and only if some owner's test happens to execute the call. An emission added to a handler no contract case drives is invisible. Commit 2 of this PR is the proof — `profile.denied` had been live from five gates and was caught only because a case reaching it was added by hand. Soften to "fails on an undeclared emission *from an exercised path*", or add a source-level backstop (AST/grep count of `ext.Record` sites asserted against `len(Required)` + known helpers).
- The `AllocsPerRun` guard is sound and measures the real `auditTimeTravel` — but it depends on `ext.sink` being nil process-wide and never calls `audittest.Install`, so it silently becomes a no-op if any sibling test in `internal/shim` leaks a sink. A `if ext.Auditing() { t.Fatal(...) }` precondition makes that loud.
- `ext.Auditing()` (`ext/audit.go:75-77`) reads a package var unsynchronized, now once per shim round trip from every connection goroutine. Benign under the set-once-from-`main()` contract, but `audittest.Install` swaps it at test time and the PR's own doc has to warn contributors about it. `atomic.Pointer[AuditSink]` removes the footgun.
- `GET /api/events` with no `schema`/`table` returns up to 1000 row images recorded as `Schema: "", Table: ""` (`internal/console/api.go:256-259`) — the record then can't answer "who read which table", which is the stated reason the emission exists. A `detail["scope"]="all_tables"` marker would mirror the shim's `full_table`/`single_row` handling.
- CLI reconstruct puts the PK in `Detail` (`:478`), console omits it (`internal/console/reconstruct.go:437-442`). Both defensible; pick one.

### Strengths

- **The contract test is genuinely behavioural and genuinely bites.** `observed` is accumulated from what the recording sink captured, never derived from `Required`. Each case asserts `len(evs) == 1` before appending, so a subtest failing to reach the emitting code fails twice — locally and as a missing pair. Splitting by `Owner` so each package pins only what it can actually drive beats a central test that would have to fake half the paths.
- **Placement is right everywhere it exists.** All nine emissions sit after the artifact is durable and outside any lock or transaction. The shim's `FullTableGate` is released by the `defer` inside `runSnapshot`, so the audit call never runs while holding a gate slot. The cascade emission deliberately precedes `cascadeExit` in both output modes so a provably-partial recovery is still recorded — right call, well argued.
- **The `mysql:unbound` sentinel is the correct design**, pinned by a test asserting on the sentinel rather than merely on non-emptiness. Finding 1 is the sentinel doing its job — except `pgshim` never reaches the emission at all.
- **The refusal-silence pins are the non-obvious half of the contract** and both exist. Auditing unserved queries would inflate every trail; pinning that they *don't* is what stops a refactor from moving an emission above the error check.
- **Open-core boundary is clean.** No `Detail` map carries row values, column contents, `query_text`, `query_hash` or `connection_id` — only counts, timestamps, filters and output paths, with the reasoning stated explicitly.
- `MySQLRequired()` means the CLI tier isn't vacuous — a skip becomes `t.Fatalf` under `BINTRAIL_REQUIRE_MYSQL=1`, so this contract can't silently vanish the way the e2e did before #949.
- **Commit 2 is the honest move**: discovering `profile.denied` was live-but-undeclared and fixing it by adding the `Required` row, the docstring line *and* a case that reaches the gate — rather than deleting the assertion — is what makes the rest of the contract believable.

Issue #760, opened 2026-07-05
capture: post-DDL snapshot failure skips following row events while checkpoint advances — permanent, unreplayable, silent loss


nethalo:
The synchronous DDL hook takes `TakeSnapshot` best-effort: if it fails (source connection blip, revoked permission, new table), it only logs Error and "streaming continues with the previous schema". The parser then skips ALL row events of the altered/created table with `column count mismatch — skipping` / `table not in snapshot — skipping` (`parser.go:274,285`) — but `streamLoop` keeps updating `state.binlogPos`/`gtidSet` and the checkpoint **advances past the skipped events**.

Location: `internal/streamrun/streamrun.go:1653`

Scenario: `CREATE TABLE orders` + burst of INSERTs, snapshot fails on a transient timeout → hours of INSERTs discarded; when a later snapshot heals the resolver, the lost events are already behind the checkpoint and are never re-streamed. `docs/ddl-tracking.md` documents "may produce column count mismatches" but does not say the loss is **permanent and unmarked** (unlike gap auto-advance, which stamps `gap_lost_at`).

Fix: give the hook a fail-loud path (abort the stream: restart replays from the checkpoint and retries the snapshot — same philosophy as the #652 flush fix), or freeze checkpoint advance / retry the snapshot with backoff while the resolver is stale, and stamp a durable `gap_lost_at`-style marker if it proceeds.

---
_From the full-codebase audit (excl. `internal/forensics`), 2026-07-04. Each finding was adversarially verified against the code by a second agent. See sibling audit issues for related items._

PR #1186, opened 2026-07-31
stream: degrade DDL auto-snapshot on invalid tables instead of crash-looping

### LINKED ISSUE #1051: stream: DDL auto-snapshot crash-loops when any unrelated table lacks a PK (validateTables all-or-nothing + #760 fail-loud)
## Summary

The stream's DDL auto-snapshot hook calls `metadata.TakeSnapshot`, which hard-fails if **any** base table in scope lacks an explicit primary key or is not InnoDB (`validateTables`, `internal/metadata/metadata.go`). Since #760 the DDL hook is fail-loud: a snapshot error aborts the stream before the DDL event is emitted, the checkpoint stays put, and a supervisor restart re-reads the same DDL and retries the snapshot.

Combined, these two behaviors turn one no-PK table anywhere in the captured schemas into a **stream crash-loop** triggered by any DDL: the restart re-runs the snapshot, `validateTables` fails again (the offending table still exists), the stream aborts again — indefinitely, until the unrelated table gains a PK or is dropped. No events are lost (that is exactly what #760 guarantees), but capture is down and the failure mode looks like a bintrail crash rather than a source-schema problem.

## Reproduction

1. Run `bintrail stream` (or `bintrail-console watch`) against a source.
2. `CREATE TABLE scratch.t (id INT);` — no PK. (Row events for it are already skipped, as documented.)
3. Execute any DDL — e.g. `CREATE TABLE other.ok (id INT PRIMARY KEY);`.
4. The auto-snapshot fails with `snapshot validation failed — bintrail requires all tables to use InnoDB with an explicit primary key`, the stream exits non-zero, and every restart repeats steps 3–4's failure.

This is not hypothetical: transient no-PK utility/temp tables are common in the wild (e.g. WordPress plugins create and drop scratch tables as part of normal operation). Pre-#760 the same situation was silently degraded instead: the snapshot failure was logged, the DDL recorded with `snapshot_id = NULL`, and the stream continued with a stale resolver — observed on a live index as an "uncovered" DDL whose CREATE TABLE had no PK.

## Analysis

`validateTables`'s strictness is right for `bintrail snapshot` / initial setup — refusing to start capture against an unsupported schema is a good preflight. But in the DDL hook the failure is disproportionate: the snapshot is rejected wholesale because of a table that may be completely unrelated to the DDL that triggered it, and whose row events bintrail was already skipping.

## Suggested fix

In the DDL-hook path, degrade instead of failing the whole snapshot:

- Exclude no-PK / non-InnoDB tables from the snapshot and log a prominent warning naming them (they contribute no recoverable events anyway), **or**
- keep `validateTables` strict only for the table the DDL touched, warning on unrelated offenders.

Either way the hook keeps its #760 fail-loud contract for real snapshot failures (source unreachable, index write error) — the change is only that a validation failure caused by an unrelated, already-uncaptured table no longer takes the stream down.

`bintrail snapshot` (manual) and initial `up`/`EnsureResolver` behavior can stay strict.

nethalo:
## Summary

Fixes the stream crash-loop where one no-PK / non-InnoDB table anywhere in the captured schemas turned any DDL into an indefinite abort-restart cycle: the DDL auto-snapshot hook used the strict `metadata.TakeSnapshot`, whose all-or-nothing `validateTables` failed on the unrelated table; the #760 fail-loud abort kept the checkpoint on the DDL, so every supervisor restart re-read the same DDL and failed the same validation, forever.

- New `metadata.TakeSnapshotExcludingInvalid`: same snapshot pipeline, but validation violations (non-InnoDB / no explicit PK) EXCLUDE the offending tables from the snapshot instead of failing it. Their `fk_constraints` rows are dropped too, keeping FK metadata consistent with `schema_snapshots` (FKs from kept tables to an excluded parent remain). Excluded tables are reported in `SnapshotStats.ExcludedTables` (sorted).
- The DDL hook (`internal/streamrun`) now calls the new variant and logs a prominent `WARN` naming the excluded tables and the snapshot id. Rationale: those tables' row events are already skipped, so they contribute no recoverable data — degrading loses nothing, while failing loses capture entirely.
- Unchanged strictness everywhere else: `bintrail snapshot` and initial setup (`index` auto-snapshot, `EnsureResolver`) keep the strict `TakeSnapshot`. The #760 fail-loud contract still applies to REAL snapshot failures — source unreachable, index write errors, and the "every table in scope is invalid" case, which refuses to write an empty snapshot.

## Test plan

- `go test ./... -count=1` — pass.
- `go vet ./...` and `go vet -tags integration ./...` — clean.
- `go test -race ./internal/metadata/... ./internal/streamrun/...` — pass.
- Integration (Docker MySQL 8.0 on 13306): `go test -tags integration ./internal/metadata/... ./internal/streamrun/... -count=1` — pass, including three new tests:
  - `TestTakeSnapshotExcludingInvalid_degrades` — mixed scope (clean parent+FK child, no-PK FK child, MyISAM table): strict mode still fails; hook mode succeeds, records exactly the two invalid tables as excluded, writes snapshot/FK rows for only the valid ones, and the resolver loaded from that snapshot resolves the kept tables and not the excluded one.
  - `TestTakeSnapshotExcludingInvalid_allInvalidStillFails` — all-invalid scope stays a hard error (no empty snapshot).
  - `TestTakeSnapshotExcludingInvalid_cleanScope` — no violations ⇒ behaves exactly like `TakeSnapshot`, no exclusions reported.

closes #1051

### DIFF
--- a/internal/cascade/cascade.go
+++ b/internal/cascade/cascade.go
@@ -69,6 +69,13 @@ type CascadeFK struct {
 	ReferencedColumn string // parent referenced column (usually its PK)
 	DeleteRule       string // CASCADE, RESTRICT, SET NULL, NO ACTION
 	UpdateRule       string // ON UPDATE rule
+	// ChildAbsentFromSnapshot marks an edge whose child table has NO rows in
+	// the schema snapshot the FK graph was loaded from — a degraded snapshot
+	// excluded it (no explicit PK / non-InnoDB, #1051). Such a child's row
+	// events were never captured, so synthesis can find nothing for it; the
+	// walk must report the recovery as provably partial instead of presenting
+	// the inevitable zero-child scan as a clean Complete.
+	ChildAbsentFromSnapshot bool
 }
 
 // BaselineRow is one child row from a baseline snapshot, with its primary key
@@ -847,6 +854,19 @@ func SynthesizeVictims(
 					}
 					parentOldKey := valToString(oldVal)
 					cascadedHere = true
+					if fk.ChildAbsentFromSnapshot {
+						// #1051: same capture gap as the delete path — the child's
+						// events were never captured, so the scan below is a
+						// guaranteed zero. cascadedHere stays true (the parent's own
+						// reversal is still real and emitted); only the child half
+						// is missing, and that must never be silent.
+						addIncomplete("childabsent:"+fk.Schema+"."+fk.Table, fmt.Sprintf(
+							"%s.%s has cascading FK %q but is absent from the schema snapshot "+
+								"(tables without an explicit primary key or not using InnoDB are excluded "+
+								"and their row events never captured); its cascade-affected rows could NOT be reconstructed",
+							fk.Schema, fk.Table, fk.ConstraintName))
+						continue
+					}
 					if depth == 0 {
 						checkKeyChain(fk, pev, parentOldKey, item.rootTS, rootKey,
 							"some ON UPDATE cascade children may NOT be reconstructed")
@@ -936,6 +956,20 @@ func SynthesizeVictims(
 			}
 
 			for _, fk := range byParentDelete[pev.SchemaName+"."+pev.TableName] {
+				if fk.ChildAbsentFromSnapshot {
+					// #1051: the FK snapshot knows this cascading edge, but its
+					// child was excluded from the schema snapshot (no PK /
+					// non-InnoDB) so its row events were never captured. The
+					// candidate scan below can only ever return zero — a capture
+					// gap, not proof of no children — so skip it and report the
+					// recovery as provably partial instead of a silent Complete.
+					addIncomplete("childabsent:"+fk.Schema+"."+fk.Table, fmt.Sprintf(
+						"%s.%s has cascading FK %q but is absent from the schema snapshot "+
+							"(tables without an explicit primary key or not using InnoDB are excluded "+
+							"and their row events never captured); its cascade-affected rows could NOT be reconstructed",
+						fk.Schema, fk.Table, fk.ConstraintName))
+					continue
+				}
 				refVal, ok := parentRow[fk.ReferencedColumn]
 				if !ok {
 					// The FK graph (latest snapshot) names a referenced column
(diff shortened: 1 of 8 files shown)


nethalo:
Review follow-up (silent-failure finding): dropping an excluded child's `fk_constraints` rows let `recover-cascade` report a clean **Complete** over a real cascade — a no-PK InnoDB child can carry a genuine `ON DELETE CASCADE` FK, and with the edge erased the synthesis engine never even looked.

Verification also showed that merely keeping the edge is not enough: with no baseline configured, the excluded child's guaranteed zero-event candidate scan is indistinguishable from "no children existed" (the skew probe samples zero rows and deliberately stays inconclusive), so the run would still read Complete.

Fix in 9974e1c:
- `TakeSnapshotExcludingInvalid` now **keeps every `fk_constraints` row** (both directions).
- The cascade FK loaders mark edges whose child table has no `schema_snapshots` rows under the same snapshot (`CascadeFK.ChildAbsentFromSnapshot`).
- `SynthesizeVictims` skips such edges with an `Incomplete` caveat naming the child, on both the ON DELETE and ON UPDATE paths — honest partiality with or without a baseline. The `recover` cascade-parent hint is restored too.

New integration test `TestSynthesizeVictims_excludedChildFlagged` drives the real degraded-snapshot → `LoadCascadeFKsForParent` → `SynthesizeVictims` path: the excluded child's edge loads marked, the parent-DELETE run is NOT Complete, the caveat names the child, and a valid sibling stays unflagged.

Tests: full unit suite, `go vet` (both tag sets), `-race`, and the integration suites for `internal/metadata`, `internal/streamrun`, `internal/cascade` are green. Note: the cascade integration suite flakes when run in parallel with the other two packages (shared-server binlog rotation races); reproduced identically on unmodified main, so it is pre-existing — green in isolation on this branch.

nethalo:
Second review round addressed in 8a5a4d7:

- **Unit-tier guard coverage** (`internal/cascade/excluded_child_test.go`, no build tag — runs under plain `go test ./...` with no MySQL): both synthesis sites driven with hand-built `ChildAbsentFromSnapshot` edges and a **nil-DB engine**, proving the skip fires before any index scan. The ON UPDATE test also asserts the parent lands in `KeyUpdateParents`, pinning the cascadedHere-before-skip ordering.
- **ON UPDATE caveat differentiated**: it now states that the parent's key reversal in the emitted SQL is still applied (FK checks are off during apply, so nothing re-cascades), leaving the uncaptured child rows referencing a key that no longer exists. Separate dedup key (`childabsentupd:`), so a child absent under both a delete edge and an update edge surfaces both caveats; the unit test pins the wording.
- **Loader drift guards**: `TestLoadCascadeFKsFromIndex` asserts no edge is marked under a strict snapshot; the degraded-snapshot integration test now also loads through the schema-scoped `LoadCascadeFKs` (the exported, otherwise-uncovered duplicate of the `child_absent` SELECT) and asserts it marks the same edge.

**Known limitation — pre-fix snapshots**: snapshots taken before this fix physically dropped the excluded child's `fk_constraints` rows, so against such a snapshot the silent Complete persists. It self-heals on the next snapshot — a manual `bintrail snapshot` or the next DDL's auto-snapshot — which re-captures the FK rows under the new behavior.

Tests: `go test ./... -count=1` green (new unit tests run in this tier), `go vet` (both tag sets) clean, `-race` on cascade green, integration suites for `internal/metadata` and `internal/cascade` (isolated, per the pre-existing parallel binlog-rotation flake) green against Docker MySQL 13306.

nethalo:
CI red (integration 8.0/8.4) root-caused and fixed in 4c3d59a — a **false positive** of the new excluded-child caveat, the cry-wolf class this design promised to avoid.

**Root cause**: the flag was *inferred* — "child has `fk_constraints` rows but no `schema_snapshots` rows under the same snapshot_id ⇒ excluded". That shape is producible with nothing excluded: every hand-seeded recover-cascade fixture (cli/console/mcptools seed an FK row plus snapshot rows for the parent only), and in production a table created between `TakeSnapshot`'s columns query and its FK query yields it too. So `TestRecoverCascade_endToEnd` and friends got a spurious "provably partial" verdict on Complete recoveries, deterministically on both MySQL versions.

**Fix — explicit over inferred**: degraded snapshots now *record* their exclusions in a new `snapshot_exclusions` table (`snapshot_id, schema_name, table_name, reason`), written by `TakeSnapshotExcludingInvalid` in the same transaction as the snapshot rows. The cascade FK loaders flag `CascadeFK.ChildExcludedFromSnapshot` (renamed from `ChildAbsentFromSnapshot`) only from that record, via a shared `childExcludedExpr` so the two hand-written SELECTs cannot drift. The writer knows exactly what it excluded, so the flag can never fire on a state a strict writer produced — false positives are structurally impossible; false negatives (legacy index without the table) degrade to the pre-#1051 silent behavior, which is the correct direction under the Result-contract doctrine. The table is lazily created by the writer (`snapshot_id_seq` pattern) and eagerly provisioned by `indexer.EnsureSchema`; all readers tolerate its absence (missing table = no exclusions, mirroring the `fk_constraints` tolerance).

**Tests**: `go test ./... -count=1` green; `go vet` (plain + `-tags integration`) clean; `-race` on cascade/metadata/indexer green; integration against Docker MySQL 13306 green for `internal/cli` (the previously failing suite, in full), `internal/metadata` (now also asserts the exclusion rows + reasons), `internal/mcptools`, `internal/console` cascade tests, `internal/streamrun`, `internal/indexer`, and `internal/cascade` (isolated, per the pre-existing parallel binlog-rotation flake).

Issue #1050, opened 2026-07-18
mcp: list_schema_changes omits snapshot_id — uncovered DDLs cannot be identified from the MCP surface


nethalo:
## Summary

The MCP `list_schema_changes` tool does not return `snapshot_id` (or any coverage indicator), so it is impossible to tell **which** DDLs lack a schema snapshot from the MCP surface — even though `status` (also exposed via MCP) reports the aggregate count of uncovered DDLs and warns that recovery across them may require a manual snapshot.

## Current behavior

`MakeSchemaChangesTool` (`internal/mcptools/mcptools.go`) selects:

```sql
SELECT id, detected_at, schema_name, table_name, ddl_type, ddl_query, binlog_file, binlog_pos, gtid
FROM schema_changes ...
```

`snapshot_id` is omitted from both the SELECT and the response DTO.

## Why it matters

The intended triage flow is: `status` warns "N DDL(s) detected without auto-snapshot" → operator (or an AI agent using the MCP server) lists the schema changes to find which ones are uncovered and decide whether a manual `bintrail snapshot` is needed. That flow dead-ends today: `list_schema_changes` returns all changes indistinguishably, and the only way to identify the uncovered ones is direct SQL against the index (`WHERE snapshot_id IS NULL`), which defeats the purpose of the read-only MCP surface.

## Suggested fix

- Add `snapshot_id` (nullable) to the `list_schema_changes` response.
- Optionally add an `uncovered_only` (bool) filter parameter so an agent can go straight from the status warning to the exact rows.

If the `snapshot_state` distinction from the status-warning issue lands (taken / not_needed / failed), expose that field here too so TRUNCATE rows (deliberate NULL) aren't misread as missing coverage.

nethalo:
Related: #1049 — if the snapshot_state (taken / not_needed / failed) distinction lands there, expose it here too.

PR #1187, opened 2026-07-31
status: exclude by-design TRUNCATE rows from the uncovered-DDL warning and stop mislabeling it as file mode

### LINKED ISSUE #1049: status: 'DDLs without auto-snapshot' warning counts by-design TRUNCATE rows and mislabels stream-mode failures as file mode
## Summary

`bintrail status` reports a misleading warning about DDLs without auto-snapshot. The `UncoveredDDLs` counter is computed as:

```sql
SELECT COUNT(*) FROM schema_changes WHERE snapshot_id IS NULL
```

(`internal/status/status.go:239`), and rendered as:

> Warning: N DDL(s) detected without auto-snapshot (file mode) — recovery across these DDLs may require manual snapshot

(`internal/status/status.go:730`). Both the count and the label are wrong in common cases.

## Problem 1: TRUNCATE rows are counted as "uncovered"

TRUNCATE deliberately records `snapshot_id = NULL` — it does not change table structure, so both the stream DDL hook (`internal/streamrun/streamrun.go`, `SetSyncDDLHook`) and the file-mode handler (`cliapp/index.go`) skip the snapshot by design ("DDL detected (no snapshot needed)"). The status query cannot distinguish "snapshot deliberately not needed" from "snapshot missing/failed", so every TRUNCATE permanently inflates the warning.

Observed on a live index: 4 "uncovered" DDLs, of which 3 were TRUNCATEs (by design, recovery across them needs nothing) and only 1 was a genuinely failed auto-snapshot. The warning is mostly noise, which trains operators to ignore it.

## Problem 2: the "(file mode)" label is hardcoded

The text claims file mode, but `snapshot_id IS NULL` rows are also produced by:

- a **failed** auto-snapshot in stream mode (the hook records the change with NULL before erroring — see the fail-loud path added in #760, and the pre-#760 warn-and-continue path), and
- TRUNCATE in **any** mode (see above).

An operator running only `bintrail stream` sees "(file mode)" and reasonably concludes the message cannot apply to them.

## Suggested fix

- Exclude `ddl_type = 'TRUNCATE TABLE'` from the `UncoveredDDLs` count (minimal fix), **or** persist the distinction explicitly (e.g. a `snapshot_state` enum: `taken` / `not_needed` / `failed` / `unavailable`) so status can report "N failed, M not needed".
- Reword the warning to name both origins ("file mode without --source-dsn, or a failed auto-snapshot") or branch on the recorded state.
- Ideally list the affected `schema_changes.id`s (or the first few) so the operator can locate them without querying the index by hand — today the count is not actionable.

JSON output (`uncovered_ddls`) has the same semantics and should follow the same fix.

nethalo:
closes #1049

## Summary

`bintrail status` counted every `schema_changes` row with `snapshot_id IS NULL` as a "DDL without auto-snapshot" and attributed all of them to "(file mode)". Both were wrong:

- **TRUNCATE is by design.** TRUNCATE TABLE changes no table structure, so every capture path deliberately records it with `snapshot_id = NULL` ("DDL detected (no snapshot needed)"): the stream DDL hook (`internal/streamrun`), the file-mode handler (`cliapp/index.go`), and the PostgreSQL TRUNCATE audit marker (`internal/pgstreamrun`). Each TRUNCATE permanently inflated the warning. `LoadCoverage` now excludes `ddl_type = 'TRUNCATE TABLE'` from `UncoveredDDLs` — the minimal fix from the issue's suggested options. The JSON `uncovered_ddls` field uses the same counter, so it follows automatically.
- **The "(file mode)" label was hardcoded.** NULL rows are also produced by a failed auto-snapshot in stream mode, so the warning now names both origins: *"file-mode indexing without --source-dsn, or a failed auto-snapshot"*.

Of the issue's suggested fixes, this takes the minimal ones (count exclusion + reword). The richer options — a persisted `snapshot_state` enum and listing affected `schema_changes.id`s — are left out on purpose: they overlap with the coverage-state surfacing tracked in #1050, and the minimal change is what makes the warning truthful today. #1051 (the stream-mode crash-loop origin) is untouched.

## Test plan

- New `TestLoadCoverage_uncoveredDDLs_integration` drives the real `LoadCoverage` SQL against seeded `schema_changes` rows: a TRUNCATE/NULL row no longer counts, an ALTER/NULL row still counts, an ALTER with a `snapshot_id` never counted. **Ran against Docker MySQL 8.0 on 13306: PASS.**
- Tightened `TestWriteStatus_withCoverage` to pin the new wording and reject the old "(file mode)" label.
- `go test ./... -count=1` — pass (two `consoleapp` flashback/monitor failures appeared only in the full parallel run, are unrelated to this change, and pass in isolation — the known full-parallel flake).
- `go vet ./...` — pass (also with `-tags integration` for `internal/status`).
- `go test -tags integration ./internal/status/... -count=1` — 71 pass, 0 fail, 0 skip.

### DIFF
--- a/docs/ddl-tracking.md
+++ b/docs/ddl-tracking.md
@@ -14,7 +14,7 @@ DDL tracking solves three problems:
 
 1. **Detection**: The parser identifies DDL statements (`ALTER TABLE`, `CREATE TABLE`, `DROP TABLE`, `RENAME TABLE`, `TRUNCATE TABLE`) and emits them as events instead of just logging warnings.
 2. **Auto-snapshot**: When a DDL is detected and a source database connection is available, dbtrail automatically takes a new snapshot and hot-swaps the resolver — no manual intervention needed. This works in both stream mode (always has source connection) and file mode (when `--source-dsn` is provided).
-3. **Restore coverage**: The `status` command shows the time range of indexed events and warns about DDLs that weren't followed by a snapshot (file mode), so you know where recovery gaps might exist.
+3. **Restore coverage**: The `status` command shows the time range of indexed events and warns about DDLs that weren't followed by a snapshot — whether from file-mode indexing without `--source-dsn` or from a failed auto-snapshot — so you know where recovery gaps might exist.
 
 ---
 
@@ -85,7 +85,7 @@ Key fields:
 |---|---|
 | `ddl_type` | One of `ALTER TABLE`, `CREATE TABLE`, `DROP TABLE`, `RENAME TABLE`, `TRUNCATE TABLE` |
 | `ddl_query` | The full DDL statement from the binlog |
-| `snapshot_id` | The snapshot taken after this DDL, or NULL if no source connection was available |
+| `snapshot_id` | The snapshot taken after this DDL. NULL when none was taken: file mode without `--source-dsn`, a failed auto-snapshot, or `TRUNCATE TABLE` (which changes no table structure, so no snapshot is needed — by design, in every mode) |
 
 This table is created by `bintrail init` and must exist in the index database. Older index databases (created before this feature) won't have it — the status command handles this gracefully by treating a missing table as zero schema changes.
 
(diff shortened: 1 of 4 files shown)


nethalo:
Folded in the review's LOW finding: docs/ddl-tracking.md still documented the old warning semantics. Updated the sample warning text to the actual output, documented the TRUNCATE-by-design exclusion from the count, and replaced the "(file mode)" attribution with both real origins (file-mode indexing without `--source-dsn`, or a failed auto-snapshot in any mode) — including the `snapshot_id` field description and a footnote on the stream-vs-file comparison table. Docs-only commit 2b30abe.

Issue #1022, opened 2026-07-14
console(pg): verify-trigger supports PostgreSQL (verify engine PG type/PK handling)


nethalo:
Part of #1018. **Depends on A.** Reconstruct dependency met (#593 closed).

## Problem

`internal/console/verify_trigger.go` `handleVerifyTrigger` (:152) + the verify controller run the MySQL verify engine. A PostgreSQL source currently reaches it and returns a misleading **inconclusive** — surfaced at the message level by #1009 ("PK type unsupported"). The reconstruct machinery verify relies on (baseline + delta replay) is available for PG (#593 closed), so this is the **verify engine's PG type / PK handling**, not a reconstruct gap.

## Scope

Make console verify — baseline-anchored, and live-source where a PG source DSN is configured — actually run and reach a real MATCH / MISMATCH verdict for a PostgreSQL server, at parity with MySQL for the supported type surface.

- #1009 is the **message-level symptom** (a clear not-supported verdict where a type genuinely isn't handled). Cross-ref, don't duplicate — this issue is "make verify run and pass for the supported cases"; #1009 is "say the right thing for the unsupported ones".

## Acceptance

"Run verification" on a PG server produces a real verdict (not a blanket inconclusive) for the supported type surface, same as MySQL.

## Related

Epic #1018 · verdict message #1009 · reconstruct #593 · #597

Issue #528, opened 2026-06-20
[pg] Extract parser.Event to a dependency-neutral package + generalize position fields


nethalo:
Part of #527 (PostgreSQL-as-source).

**Precondition for everything else** — do this first or every dependency boundary is illusory.

`parser.Event` currently lives in `internal/parser`, which imports `go-mysql-org/go-mysql/replication`. As a result, ~38 source-agnostic downstream files (e.g. `internal/recovery`, `internal/reconstruct`, `internal/query`) transitively link go-mysql even though they only read named row data off `Event`.

## Scope
- Extract the `Event` struct (and the `EventType`/`DDLKind` constants it needs) into a **dependency-neutral package** (e.g. `internal/event`) that imports no source driver.
- Generalize the MySQL-named position fields (`BinlogFile`/`StartPos`/`EndPos`/`GTID`) into a **source-agnostic position vocabulary** that can also express a Postgres LSN/slot. (Keep the MySQL path byte-identical in behavior.)
- Repoint `internal/parser`, `internal/indexer`, and all downstream consumers at the neutral package.

## Why
This is a refactor of existing code, shared-benefit (de-bloats current read paths that needlessly link go-mysql), and the OPPOSITE of a rewrite — it is what lets one core serve N sources without duplicating the value stack.

## Acceptance
- Downstream read-only packages no longer transitively import go-mysql via `Event`.
- MySQL/MariaDB capture + index + recover/reconstruct unchanged (full test suite green).
- The neutral `Event` can represent a Postgres LSN position.

PR #1262, opened 2026-08-07
verify: PG-native live-source consistency checksum

### LINKED ISSUE #1024: console(pg): live-source verify for PostgreSQL sources (native consistency checksum)
Follow-up from #1018 / PR #1023 (the one deliberately-deferred parity edge).

## Context

The console (and `bintrail verify`) exposes two verify modes:
- **baseline-anchored** (the drift-free default) — now fully supported for PostgreSQL (#1022 / PR #1023).
- **live-source** (`--source-dsn` / `mode=live-source`) — still MySQL-only.

Live-source verify fingerprints the live source table with `consistency.ConsistentTableChecksumNormalized`, which is MySQL SQL (`information_schema.COLUMNS`, `SHA2`, `CAST(... AS CHAR)`). A PostgreSQL source has no drop-in equivalent, so PR #1023 **refuses** a live-source verify for a PG server with a clear message at the API boundary (`internal/console/verify_trigger.go`), the CLI (`internal/cli/verify.go`), and an engine backstop (`cmd/bintrail-console/verify.go`) — strictly better than the previous misleading inconclusive, and aligned with #1009.

## Scope

Add a PostgreSQL live-source consistency-checksum path so `mode=live-source` reaches a real verdict against a live PG source, at parity with MySQL live-source verify. This needs a PG-native normalized table checksum (the pg equivalent of `ConsistentTableChecksumNormalized`) whose output is comparable to the reconstructed digest, plus wiring it behind the existing live-source branch instead of the current refusal.

## Notes

- Baseline-anchored is the primary/default drift-free path and is already at parity, so this is an enhancement of the secondary mode, not a launch blocker.
- Evidence-gate before building: confirm there is real demand for live-source (vs baseline-anchored) verify on PG sources.

Related: #1018 (epic) · #1009 (verdict messaging) · #597 (PG beta→GA)

nethalo:
closes #1024

## Summary
- **`internal/pgverifysource` (new)**: `ConsistentTableChecksumPG` — the PG-native sibling of `consistency.ConsistentTableChecksumNormalized`. One `REPEATABLE READ, READ ONLY` transaction; `pg_current_wal_lsn()` captured by the snapshot-materializing first statement; attnum-ordered non-generated columns (pgbaseline's exact column contract); values read in **protocol text format** (`pgx.QueryResultFormats{TextFormatCode}` — never `::text` casts: `boolean::text` is `'true'`, `bool_out`/pgoutput/COPY emit `'t'`) on a connection pinned to the capture plane's render GUCs; rows folded through the shared `consistency.Hasher` so digests share the version-tagged contract. It is a separate package because it links pgx, which `TestCoreBinaryIsPostgresFree` (#534) and the read-layer dep guard (#528) ban from the core binary and the read stack.
- **`verify.VerifyTablePG` (pgx-free)**: mirror of `VerifyTable` with the PG-path decisions already ratified elsewhere — no PK-type gate (text identity, as in `VerifyBaselinePair`'s pg branch), time-bounded delta window (the "PG never sets a position bound" invariant), `pgTextPK` reconstruct. Normalization is symmetric by construction: both sides normalize under the empty `data_type` `WritePGSnapshot` stores (`pgNormalizeRenderedBytes` ≡ `renderCellNormalized` on strings — pinned by a unit test).
- **Coverage verdict** (`pgCoverageVerdict`, pure + unit-tested grid): `checkpoint >= anchor` **proves** coverage; behind-anchor proceeds with a `coverage unverified` note — WAL from other databases and non-transactional activity advances the anchor without producing indexable events, so blocking there would be permanently-inconclusive cry-wolf; the same documented trade-off as a GTID-off MySQL source. `gap_lost`, missing stream state, and a zero checkpoint stay hard inconclusive (never a false mismatch).
- **Wiring**: the CLI routes by index flavor through a new seam (`cli.SetPGLiveVerifyConnect`) that only `bintrail-pg` fills — the core `bintrail` binary refuses with a pointer to `bintrail-pg` and stays postgres-free. The console API-boundary refusal and the engine backstop are removed; the supervisor routes to the PG loop by the index's recorded flavor (the daemon already links the PG stack). PG target tables come from the resolver, not the `MAX(snapshot_id)` query (a PG index stores one relation per snapshot_id — that query silently names a single table).
- **Docs**: `verify.md` gains per-source notes (PG live-source semantics, coverage caveat, `LSN:` anchors); the stale `postgres.md`/`verify.md` claims that predated baseline-anchored PG verify (#1022/PR #1023) are corrected — full-table `reconstruct` remains out of scope, `verify` is not bound by that limit.

## Test plan
- [x] Unit tests pass (`go test ./... -count=1`, full sweep) — including the previously-broken dep guards `TestCoreBinaryIsPostgresFree` and `TestReadLayerDoesNotLinkGoMySQL`
- [x] `go vet ./...` and `go vet -tags integration ./...` clean
- [x] Integration (MySQL 8.4 @13306 + PostgreSQL 16 @15432, `-race`): `TestConsistentTableChecksumPG_Integration` — digest equals a **hand-built** digest of the expected pinned-GUC renderings (`t`, `\x` hex bytea, UTC timestamptz, numeric scale, NULL≠empty), row-order independence, generated-column exclusion
- [x] Integration end-to-end: `TestVerifyTablePG_Integration` — pgbaseline snapshot + seeded index → **match** on identical data (proven-coverage branch, no note), then conclusive **mismatch** for an uncaptured UPDATE (content) and DELETE (row count)
- [x] Full `-tags integration` suites for `internal/verify`, `internal/consistency`, `internal/console`, `consoleapp`, `internal/cli` green

### DIFF
--- a/cmd/bintrail-pg/main.go
+++ b/cmd/bintrail-pg/main.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/dbtrail/dbtrail/internal/cli"
 	"github.com/dbtrail/dbtrail/internal/observe"
+	"github.com/dbtrail/dbtrail/internal/pgverifysource"
 	"github.com/dbtrail/dbtrail/internal/telemetry"
 )
 
@@ -73,6 +74,10 @@ func init() {
 	// shared verbatim with the core binary. The PostgreSQL capture command
 	// (stream) is registered in stream.go's init().
 	cli.AddReadCommands(rootCmd)
+	// PG live-source verify provider (#1024): this binary links the PG driver
+	// stack, so it fills the seam the shared verify command leaves empty in
+	// the postgres-free core binary.
+	cli.SetPGLiveVerifyConnect(pgverifysource.LiveSource)
 	// The index-side maintenance plane (rotate, archive reconcile) — also
 	// source-agnostic, so a PostgreSQL-only install can bound its index growth
 	// with `bintrail-pg rotate` instead of needing the core MySQL binary against
--- a/cmd/bintrail-pg/verify_seam_test.go
+++ b/cmd/bintrail-pg/verify_seam_test.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/dbtrail/dbtrail/internal/cli"
+)
+
+// TestPGLiveVerifySeamIsFilled is the wiring guard for #1024: this binary's
+// init() must install the PostgreSQL live-source verify provider
+// (cli.SetPGLiveVerifyConnect(pgverifysource.LiveSource)). Delete that call
+// and every other suite stays green while `bintrail-pg verify --source-dsn`
+// refuses at runtime — with a message telling the operator to run the very
+// binary they are already running. Mutation-detecting by construction: the
+// test observes the seam's state after init, not the code that fills it.
+func TestPGLiveVerifySeamIsFilled(t *testing.T) {
+	if !cli.PGLiveVerifySeamFilled() {
+		t.Fatal("the PG live-source verify seam is empty — cmd/bintrail-pg's init() must call cli.SetPGLiveVerifyConnect(pgverifysource.LiveSource)")
+	}
+}
--- a/consoleapp/verify.go
+++ b/consoleapp/verify.go
@@ -256,7 +256,7 @@ func (s *verifySupervisor) run(req console.VerifyRequest, baselineSrc string) {
 			runErr = s.runLiveSource(req, db, resolver, dbName)
 		}
 	case console.VerifyModeRecoverInputs:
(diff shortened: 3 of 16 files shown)


15432:
ok

nethalo:
Review round applied (4 agents: comment-analyzer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer — every finding verified before acting): LSN==0 guard, coverage note appended on mismatch instead of dropped, gap_lost scoped to the comparison window, **PG recover-inputs enumeration fixed** (both CLI and console walked ONE relation via MAX(snapshot_id)), empty-enumeration guard in the console, pgverifysource composition unexported (pin invariant now structural), and mutation-detecting wiring tests for all three delivery seams plus the missing delta-fold leg in the e2e (see 168d408).

Issue #1009, opened 2026-07-12
verify: PostgreSQL source reports a misleading 'PK type unsupported' inconclusive instead of a clear not-supported verdict


nethalo:
## Summary

`bintrail-pg verify` (baseline-anchored mode) on a PostgreSQL source returns an **`inconclusive`** verdict whose reason blames the primary-key type, which reads like a fixable config error when the real cause is that the check is not wired for PostgreSQL at all.

## Repro (live, RDS PostgreSQL 16.14)

Two baselines taken, then:

```
$ bintrail-pg verify --index-dsn '<mysql-index>' --baseline-dir /baselines
TABLE          STATUS        ROWS(src/recon)  DETAIL
public.orders  inconclusive  0/0              primary-key column "id" has type "" unsupported by the baseline canonicalizer

0 match, 0 mismatch, 1 inconclusive, 0 error
no tables were verified (1 inconclusive); nothing proven
```

## Root cause

Baseline-anchored `verify` reconstructs the previous baseline forward via the **full-table** reconstruct path, which gates on `reconstruct.SupportedPKType(dataType)`. For a PostgreSQL source `schema_snapshots.data_type` is the empty string `""` (PG records `pg_type_oid`, not a MySQL type token — same shape behind #1006), so the gate reports the PK type as unsupported. Full-table reconstruct for PostgreSQL is deliberately out of GA scope (the PG baseline omits the `CREATE TABLE` metadata full-table reconstruct needs — see #597), so `verify` cannot run for a PG source today. That is a known limitation; the **message** is the defect.

## Impact

The `id`-column message implies a per-table PK-type problem the operator could fix (change the PK, re-snapshot), when nothing they do will make it pass — PG verify is simply not wired. A `verify` positioned for cron/CI that returns `inconclusive`/`exit 0` with a misleading reason is worse than an explicit refusal.

## Suggested fix

Detect the PostgreSQL source flavor (`query.SourceFlavor`) at the top of the verify path and emit an explicit verdict — e.g. `verify is not supported for PostgreSQL sources yet (baseline-anchored verify uses full-table reconstruct, out of GA scope — see #597)` — instead of running into the canonicalizer and reporting an opaque per-column inconclusive. Same class as #1006, but in the full-table/verify path rather than the single-row `_snapshot` path.

PR #1269, opened 2026-08-07
reconstruct/verify: refuse generated PK members up front (MariaDB system versioning)

### LINKED ISSUE #1266: reconstruct: full-table PK canonicalization fails when a system-versioning period column is a PK member
Follow-up from #863 / PR #1263 (found in review, confirmed empirically; out of that PR's scope — the parse-level fix there is a strict improvement regardless).

MariaDB extends a system-versioned table's PRIMARY KEY with the `ROW END` period column (`PRIMARY KEY (`id`,`row_end`)` — observed on MariaDB 11.4 for both visible and INVISIBLE explicit period columns). After #863, the baseline correctly excludes `row_start`/`row_end` (mydumper never dumps their values), but the schema snapshot keeps them: `metadata.TableMeta.PKColumnMetas()` selects on `IsPK` only, so the full-table reconstruct join key still includes `row_end`, which is absent from every baseline row.

Observed (MariaDB 11.4, mydumper v1.0.3-1, real end-to-end run):

- `bintrail baseline` — **works** (post-#863).
- Single-row `reconstruct --pk 1 --pk-columns id` — **works**, returns the correct current row.
- Full-table `reconstruct --output-format mydumper` — fails loud: `canonicalize baseline PK for svdb.sv_explicit: PK column missing from row: column "row_end" not in baseline row` (`internal/reconstruct/pk_canonicalize.go`'s `MissingPKColumnError`).
- `verify --source-dsn` (live-source mode) — same error, reported as table status `error` (not a false mismatch), exit non-zero.

Not silent, not corruption — but the refusal for these tables now lives at reconstruct/verify time instead of baseline time.

Likely fix direction (needs design care): exclude generated columns (`IsGenerated`, which MariaDB sets for period columns via `GENERATION_EXPRESSION = 'ROW START'/'ROW END'`) from the reconstruct-side PK join key, since for **current-state** rows the remaining PK columns are unique (the period column exists in the PK only to disambiguate history rows, which neither the baseline nor the binlog current-state fold carry). The same consideration applies to an ordinary STORED generated column inside a PK — rare in MySQL, but MariaDB produces the versioned-table shape by default, and the event-side `pk_values` (built from the parser's PK column list) must stay consistent with whatever the baseline-side key becomes.

nethalo:
closes #1266

## Summary
- **Empirical finding that reshaped the fix** (MariaDB 11.4, real binlog): the issue's suggested direction — exclude generated columns from the reconstruct-side PK join key — would corrupt silently. The binlog carries **history rows as Write_rows** and **versioned DELETEs as row_end Update_rows** under the same remaining key, so a reduced key folds history over current state, appends orphan history rows as duplicate live rows, and resurrects deleted rows. Full support needs versioning-aware fold semantics (follow-up material; evidence recorded on `GeneratedPKColumn`'s doc comment).
- New `reconstruct.GeneratedPKColumn` + `GeneratedPKGateReason`: detect a STORED/VIRTUAL generated PK member (the MariaDB system-versioning shape, `PRIMARY KEY (id, row_end)`) and render the shared refusal text.
- `ReconstructTable` refuses **up front** on both full-table paths — baseline merge (previously died per-row with `MissingPKColumnError` deep in the probe) and the **binlog-only fallback** (previously *silent corruption*: no baseline probe to fail loudly, history inserts emitted as duplicate live rows).
- `verify` (live-source `VerifyTable` and baseline-anchored `VerifyBaselinePair`) reports the shape as **`inconclusive`** instead of `error` — a permanent table property must not fail every verify run forever (same stance as "table has no primary key"). Gate sits inside the `!pg` branch; `ExplainBaselinePairMismatch` needs no sibling (only runs after a MISMATCH, which a gated table never reports).
- Unchanged and still working: `bintrail baseline`, single-row reconstruct with explicit PK columns, query, recover.

## Test plan
- [x] Wiring tests through the real call paths (nil DBs; deleting a gate makes each test die on the nil DB downstream): `TestReconstructTable_generatedPKRefusal_baselinePath` / `_binlogOnlyPath`, `TestVerifyTable_generatedPKInconclusive`, `TestVerifyBaselinePair_generatedPKInconclusive`, `TestVerifyBaselinePair_pgBypassesGeneratedPKGate`
- [x] `go test ./... -count=1` green; `internal/reconstruct` + `internal/verify` also green under `-race`
- [x] `go vet -tags integration ./...` clean

### DIFF
--- a/docs/query-and-recovery.md
+++ b/docs/query-and-recovery.md
@@ -537,6 +537,14 @@ The resolver is loaded best-effort in the `recover` command — a failure logs a
 
 Generated columns (`STORED` or `VIRTUAL`) are computed by MySQL and cannot be set explicitly, so the generated script skips them in `INSERT`/`UPDATE` SET clauses — the script won't fail trying to assign a value MySQL owns.
 
+### System-versioned tables (MariaDB)
+
+MariaDB silently extends a system-versioned table's `PRIMARY KEY` with its `ROW END` period column (`PRIMARY KEY (id, row_end)`), and that column is generated — baselines never carry its values, and the binlog records history rows and versioned `DELETE`s under the same surviving key. Rendering such a table as a whole would risk emitting history rows as live data, so bintrail refuses instead:
+
+- **Full-table `reconstruct`** (baseline merge and binlog-only fallback alike) and the shim's **full-table `_flashback`/`_snapshot`** views refuse with an error naming the generated PK column.
+- **`verify`** reports the table as `inconclusive` — a permanent property of the table's shape, not a failure. A run scoped only to such tables still exits non-zero (nothing was proven).
+- **Still available**: `query` and `recover` are not gated, and single-row `reconstruct` works with an explicit column list, e.g. `--pk 1 --pk-columns id`.
+
 ### Column type encodings (BLOB/TEXT, GEOMETRY, VECTOR)
 
 Columns that MySQL delivers as raw bytes are stored base64-encoded in the index, so `recover` decodes them back to a loadable literal using the column type from the schema snapshot:
--- a/internal/baseline/schema.go
+++ b/internal/baseline/schema.go
@@ -115,10 +115,13 @@ var generatedRe = regexp.MustCompile(`(?i)\bAS\s*\(.*\)\s*(?:VIRTUAL|STORED|PERS
 // Known scope limit (#1266): MariaDB extends the table's PRIMARY KEY with the
 // ROW END column, and the schema snapshot keeps it (PKColumnMetas selects on
 // IsPK only), so FULL-TABLE reconstruct — and verify, which reconstructs —
-// still refuse such a table loudly at the PK-canonicalization step
-// (MissingPKColumnError: the baseline no longer carries the column). The
-// baseline itself, single-row reconstruct with --pk-columns, and the query
-// paths all work.
+// refuse such a table UP FRONT (reconstruct.GeneratedPKColumn gates both
+// full-table paths; verify reports it inconclusive rather than error).
+// Excluding the column from the join key instead would corrupt silently —
+// the binlog carries history rows and versioned deletes under the same
+// remaining key; see GeneratedPKColumn's comment for the MariaDB 11.4
+// evidence. The baseline itself, single-row reconstruct with --pk-columns,
+// and the query paths all work.
 var rowPeriodRe = regexp.MustCompile(`(?i)\bGENERATED\s+ALWAYS\s+AS\s+ROW\s+(?:START|END)\b`)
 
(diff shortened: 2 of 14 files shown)


nethalo:
Review round applied (commit 617df38):

**Extended** — same defect class found by review in three more surfaces, now gated: shim `_snapshot` baseline merge (died per-row with `MissingPKColumnError` as a generic wire error), shim `_flashback`/`runFullTable` binlog-only fold (**silent corruption**: history rows rendered as duplicate live rows, tombstoned deletes stayed live), and recover-cascade's Phase-2 baseline fallback (unsatisfiable "run `bintrail snapshot`" remediation). All refuse with the versioning-aware cause; the shim refusal deliberately does NOT steer to `_flashback`.

**Refuted with evidence, kept as-is** (reasoning documented in code): gate ordering after the `SupportedPKType` loop (preserves the #1009 PG wrong-path verdict) and `verify --check recover` (its chain walk already resets state on PK-changing UPDATEs, so a versioned tombstone can't false-mismatch).

**Known, pre-existing, not addressed here**: `baseline refresh` is all-or-nothing, so one versioned table among the refreshed set still blocks publication (it did before this PR too, with the deep error). Giving refresh skip semantics for permanent table shapes would be a separate design change to its todo-o-nada contract.

nethalo:
Second review round (4 specialized agents: code / comments / tests / silent-failures) applied in c1f241e:

- **Fixed — false operator steer**: the shim refusal said "single-row lookups are unaffected", but the versioned PK is composite `(id, row_end)` and `PKColumnCheck` refuses the `WHERE pk = v` shape on exactly these tables. Now steers to the CLI single-row path (where `--pk-columns` is mandatory, so the claim is true).
- **Fixed — prose promoted to guards**: gate-ordering tests at all four sites (`{IsGenerated, DataType:""}` must keep the #1009 PG wrong-path verdict), and a MariaDB integration test that creates a REAL versioned table, snapshots it through the production pipeline, and asserts the gate fires (run green against live mariadb:11.4 locally; runs in the `integration-mariadb-source` CI job).
- **Fixed**: Debug line on the shim gate's blind branch (no snapshot metadata → fold proceeds unchecked); nil-Gate test-comment mechanism corrected; steer assertion hardened; "unaffected" → "not gated"; docs section in query-and-recovery.md.
- **Filed separately**: #1272 — IMPLICIT system versioning is a silent sibling (snapshot can't see the hidden period columns, indexer skips every event, reconstruct returns baseline-only state with no refusal). Different fix layer (capture/snapshot), evidence in the issue.
- **Known, unchanged**: cascade Phase-2's refusal lands in the `--allow-incomplete`-suppressible caveat bucket labeled `baselinefail:` (loud by default; the full cause string survives), and cascade Phase-1 (no baseline configured) has no metadata seam to gate at all.

Issue #700, opened 2026-07-02
Detect schema drift at capture time: cross-check resolver against TABLE_MAP column names (binlog_row_metadata=FULL)


nethalo:
## Problem

Column mapping depends entirely on the `schema_snapshots` resolver. When DDL lands between snapshots, today's behavior is the documented gotcha: **"column count mismatch → warn and skip the table's events"** — silent event loss until an operator notices the warning and re-runs `bintrail snapshot`. Worse, a same-count drift (e.g. `DROP COLUMN a` + `ADD COLUMN b` in one ALTER, or a column rename) is not detected at all: events index with values attributed to the **wrong column names**, which is corruption, not loss.

## Opportunity

With `binlog_row_metadata=FULL` on the source (MySQL 8.0+, also MariaDB 10.5+), every TABLE_MAP event embeds the real column names. go-mysql already decodes them: `TableMapEvent.ColumnNameString()` (`replication/row_event.go`). We currently never look.

That gives us a per-event ground truth to cross-check the resolver against — turning both silent failure modes into loud, actionable ones.

Prior art: [avaitla/mysqlbinlog2parquet](https://github.com/avaitla/mysqlbinlog2parquet) relies on this variable exclusively (no snapshot state at all). We don't want to replace the resolver (the variable is opt-in and its metadata lacks what snapshots carry), but as a **cross-check** it's free fidelity.

## Sketch

1. **Parser/indexer**: when `TableMapEvent.ColumnNameString()` is non-empty, compare against the resolver's column list for that table:
   - names match ⇒ proceed as today (zero cost beyond the compare; cache per table@table-map to avoid re-comparing every event)
   - mismatch ⇒ **fail loud**: structured error/warn naming the table, the two column lists, and the remediation (`bintrail snapshot` re-take). Whether to skip-with-loud-error or hard-stop the stream should follow the precedent set by the schema-drift fail-loud work in #601 (detection == emission by construction).
2. **Optional trust mode (follow-up, maybe out of scope)**: when metadata is FULL and the snapshot is missing/stale, consider indexing directly from TABLE_MAP names instead of skipping — recovers the events instead of dropping them. Needs care: TABLE_MAP metadata has names/types but not everything a snapshot carries (PK info comes via `binlog_row_metadata=FULL` too, but charset/collation subtleties need review).
3. **Doctor**: report `binlog_row_metadata` with the trade-off spelled out (slightly larger TABLE_MAP events; enables drift detection) — validate, never set.
[...]

PR #1276, opened 2026-08-07
snapshot: synthesize hidden period columns of implicitly system-versioned MariaDB tables

### LINKED ISSUE #1272: MariaDB implicit system versioning: hidden period columns are invisible to the snapshot, so every row event is skipped and full-table reconstruct silently returns baseline-only state
Sibling of #1266 (explicit period columns), found while reviewing PR #1269. The explicit form now refuses loudly; the IMPLICIT form — `CREATE TABLE t (...) WITH SYSTEM VERSIONING` without declared period columns, the shortest and arguably most common spelling — fails **silently**. Observed on MariaDB 11.4:

- `information_schema.COLUMNS` does not list the implicit `row_start`/`row_end` at all, so the schema snapshot records `PRIMARY KEY (id)` and no generated members — the #1266 gate structurally cannot fire.
- `information_schema.KEY_COLUMN_USAGE` **does** show `PRIMARY KEY (id, row_end)`: the server extends the key regardless of visibility.
- The binlog ROW image carries the period columns (4 columns against the snapshot's 2), so the indexer's column-count-mismatch guard warns and **skips every row event of the table**.

Net effect: zero deltas ever indexed for the table. A full-table `reconstruct` returns the baseline's as-of state with no deltas applied and **no refusal anywhere**; `verify` sees no events either. The failure mode is exactly the silent one #1266's explicit-column fix exists to prevent — but one layer earlier, at capture.

Fix direction (needs design): detect the versioned shape independently of `information_schema.COLUMNS` — `KEY_COLUMN_USAGE` for the true PK, or the table's versioning flag — and either refuse/warn loudly at snapshot/index time, or record synthetic period-column metadata so the existing generated-PK gate fires downstream. The indexer's silent per-table skip (a `slog.Warn` today) is the load-bearing part: for this shape it should be an operator-visible verdict, not a log line.

nethalo:
closes #1272

## Summary
- **The silent failure**: implicit system versioning hides `row_start`/`row_end` from `information_schema.COLUMNS`, so the snapshot undercounted the table by two columns and the indexer skipped **every** row event as a column-count mismatch — zero deltas, reconstruct silently baseline-only.
- **Fix**: `TakeSnapshot` detects the implicit form (`TABLE_TYPE = 'SYSTEM VERSIONED'`, no `ROW START`/`ROW END` generated columns listed) and synthesizes the two hidden columns into the snapshot, making implicit == explicit everywhere downstream: capture works, `pk_values` carries the extended key, and the #1266 generated-PK gates fire.
- Every synthesized fact is empirical (MariaDB 11.4), recorded in the function's doc comment: names match the TABLE_MAP FULL metadata (so the #700 drift guard agrees), name collision is impossible (`ER_DUP_FIELDNAME`), the columns sit last even after `ALTER ... ADD COLUMN`, and `KEY_COLUMN_USAGE` confirms the PK extension. MySQL sources never report that `TABLE_TYPE`, so the function is a no-op there.
- Runs after the #1051 exclusion filter on purpose — an excluded table cannot re-enter the snapshot through synthesis.

## Test plan
- [x] Unit (sqlmock): synthesis shape (ordinals, types, PK membership, generation expressions), explicit/plain tables untouched, MySQL no-op path
- [x] Integration, run green against live MariaDB 11.4 + MySQL index: end-to-end `TestParseFile_implicitSystemVersioning_mariadb` — snapshot has the 4 columns, `GeneratedPKColumn` fires, the parser **captures** the events (2 inserts / 2 updates / 0 deletes: history INSERT + tombstone UPDATE, never Delete_rows) with composite `pk_values`
- [x] `go test ./... -count=1` green; `go test -tags integration` green for metadata/parser/verify; `go vet -tags integration ./...` clean


nethalo:
Review round applied (522d576):

**Load-bearing (both confirmed, both fixed)**: MariaDB reports versioned tables as `TABLE_TYPE 'SYSTEM VERSIONED'`, so `invalidTables`' `= 'BASE TABLE'` filter let them bypass the InnoDB/no-PK validation entirely — inert while capture silently skipped them, load-bearing the moment the synthesis lets them capture. The widened scan validates them like any table and doubles as the versioned-table detection (the synthesis's own round trip is gone). And the synthesis no longer stamps `PRI` unconditionally: a PK-less versioned table (legal MariaDB) would have received a fabricated one-column generated PK whose sentinel `row_end` collapses every live row onto one `pk_values` — validation now refuses it upstream (pinned live against MariaDB: `TestTakeSnapshot_pkLessVersionedTableRefused`), with a belt in the synthesis (MariaDB extends an existing PK, never creates one).

**Also applied**: the agent BYOS path (which builds its resolver straight from information_schema and reproduced the silent skip verbatim) now runs `metadata.AddImplicitPeriodColumns` — exported sibling sharing the same shape constants; the pure rework of the synthesis (one struct-valued scan, no parallel maps, no sqlmock needed to unit-test it); docs now state the re-snapshot requirement and the baseline-refresh transition.

**Verified stable by review across 10.6/11.4/11.8**: the names/types/ordinal facts and the `TABLE_TYPE` signal; the #700 drift-guard concern was REFUTED (count-mismatch no-ops the name check).

nethalo:
Toolkit review round (4 specialized agents: code / comments / tests / silent-failures) applied in a16280f. Verdict across agents: clean on silent failures ("error handling notably loud and well-labeled"), agent wiring already mutation-pinned by sqlmock, comments "model practice". Applied: docs now name the NEW validation refusal (PK-less/non-InnoDB versioned tables fail strict snapshot where they previously slipped through unvalidated) with remediation; `invalidTables` doc names its third return + the ordering guarantee the synthesis rides; round-trip count corrected (1-2, not 2); the agent-surface EXTRA parity for explicit period columns stated (verified on the same 11.4 run). Verified and left as-is: CI enforces the MariaDB integration wiring (`BINTRAIL_REQUIRE_MARIADB=1`); the #700 drift-guard claim is bounded (guard no-ops on name-count mismatch; with synthesis both count and names match); degraded-mode exclusions land in snapshot_exclusions, not silence.

PR #1277, opened 2026-08-07
cascade: gate system-versioned child edges in both phases via a metadata probe

### LINKED ISSUE #1273: recover-cascade Phase-1 has no metadata seam, so a system-versioned child table is synthesized from history-polluted candidates with no warning
Follow-up of #1266 / PR #1269, found by review. PR #1269 gated cascade's **Phase-2** baseline fallback (`internal/cascadebaseline/provider.go`): a generated PK member — the MariaDB system-versioning shape — now fails loud with the real cause. But **Phase-1** (no `--baseline-dir`/`--baseline-s3` configured, the default) has no gate and structurally cannot have one today: the cascade engine (`internal/cascaderecover`) carries no schema resolver — the baseline provider is its only metadata-aware seam.

Consequence for a system-versioned child table (explicit period columns, the #1266 shape): the candidate-victim scan works per `pk_values` with `LimitPerPK: 1`, and a versioned table's binlog carries history rows as INSERTs under their own full `pk_values` (distinct `row_end`) plus versioned DELETEs as `row_end`-update tombstones (evidence recorded on `reconstruct.GeneratedPKColumn`'s doc comment, verified on MariaDB 11.4). Each history row therefore looks like an independent candidate row, and tombstones don't look like deletes — the synthesized child restores can be built from history-polluted state with **no warning anywhere**. The restore guards (`AND fk = <new key>` / `AND fk IS NULL` + PK-anchored predicates) bound the blast radius — restores can decline, not widen — but the emitted SQL and the reported victim set can still be wrong for these tables.

Fix direction (needs design): give the cascade engine an optional metadata probe (a resolver seam like the baseline provider's) so the CLI and console call sites can pass one, and refuse — or at minimum stamp an incomplete-caveat — when a child table's PK contains a generated member. Related labeling nit from the same review: the Phase-2 refusal currently lands in the `--allow-incomplete`-suppressible caveat bucket keyed `baselinefail:`, labeled as a transient lookup failure, although it is a permanent table property; a distinct caveat key would let operators (and the console) tell the two apart.

nethalo:
closes #1273

## Summary
- **The gap**: cascade Phase-1 (no baseline configured — the default) had no metadata seam, so a system-versioned child's candidate scan treated each history-row INSERT as an independent candidate and missed tombstoned deletes — synthesized restores from history-polluted state, silently. Phase-2's refusal (added in #1269) was also mislabeled as a transient `baselinefail:` and still fell through to the same polluted Phase-1 scan.
- **`cascade.Options.PKMetas`** — optional PK-metadata probe (adapted from the schema resolver via `PKMetasFromResolver`, wired at all three call sites: CLI, console, MCP). A child edge with a generated PK member is skipped in **both** phases with a permanent `generatedpk:` caveat, before any index or baseline work. Exit semantics unchanged: non-zero without `--allow-incomplete`; suppressing it consciously yields a safe skip (nothing synthesized), never wrong SQL.
- **`reconstruct.ErrGeneratedPK`** sentinel — wrapped by the cascadebaseline provider's refusal so probe-less callers get the same permanent classification (no string matching, per the sentinel-verdict convention), and the edge skips Phase-1 too.
- Docs: one bullet in the system-versioned tables section.

## Test plan
- [x] Nil-DB engine unit tests (excluded_child_test.go style — a reached fetch panics, proving the gates fire before any scan): probe path and provider-sentinel backstop path, including the assertion that the sentinel never lands in the `baselinefail:` bucket
- [x] `go test ./... -count=1` green; `go test -tags integration ./internal/cascade/ ./internal/cascadebaseline/` green against live MySQL; `go vet -tags integration ./...` clean

### DIFF
--- a/docs/query-and-recovery.md
+++ b/docs/query-and-recovery.md
@@ -543,6 +543,7 @@ MariaDB silently extends a system-versioned table's `PRIMARY KEY` with its `ROW 
 
 - **Full-table `reconstruct`** (baseline merge and binlog-only fallback alike) and the shim's **full-table `_flashback`/`_snapshot`** views refuse with an error naming the generated PK column.
 - **`verify`** reports the table as `inconclusive` — a permanent property of the table's shape, not a failure. A run scoped only to such tables still exits non-zero (nothing was proven).
+- **`recover-cascade`** skips any FK edge whose **child** table is system-versioned, reporting it as a permanent incompleteness caveat (its cascade side effects are not synthesized) — a candidate scan over such a child would rebuild rows from history-polluted state.
 - **Still available**: `query` and `recover` are not gated, and single-row `reconstruct` works with an explicit column list, e.g. `--pk 1 --pk-columns id`.
 
 ### Column type encodings (BLOB/TEXT, GEOMETRY, VECTOR)
--- a/docs/query-and-recovery.md
+++ b/docs/query-and-recovery.md
@@ -543,7 +543,7 @@ MariaDB silently extends a system-versioned table's `PRIMARY KEY` with its `ROW 
 
 - **Full-table `reconstruct`** (baseline merge and binlog-only fallback alike) and the shim's **full-table `_flashback`/`_snapshot`** views refuse with an error naming the generated PK column.
 - **`verify`** reports the table as `inconclusive` — a permanent property of the table's shape, not a failure. A run scoped only to such tables still exits non-zero (nothing was proven).
-- **`recover-cascade`** skips any FK edge whose **child** table is system-versioned, reporting it as a permanent incompleteness caveat (its cascade side effects are not synthesized) — a candidate scan over such a child would rebuild rows from history-polluted state. Detection reads the schema snapshot, so keep it current (`bintrail snapshot`); without one the gate cannot see the table's shape.
+- **`recover-cascade`** skips any FK edge whose **child** table's primary key contains a generated column (typically MariaDB system versioning), reporting it as a permanent incompleteness caveat (its cascade side effects are not synthesized) — a candidate scan over such a child would rebuild rows from history-polluted state. Detection reads the schema snapshot, so keep it current (`bintrail snapshot`); without one, the gate only fires when a configured baseline covers the table.
 - **Still available**: `query` and `recover` are not gated, and single-row `reconstruct` works with an explicit column list, e.g. `--pk 1 --pk-columns id`.
 
 ### Column type encodings (BLOB/TEXT, GEOMETRY, VECTOR)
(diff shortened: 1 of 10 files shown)


nethalo:
Review round applied (3e78d45) — 8 finder angles, every finding verified against the code before acting:

**Applied**: shared caveat frame + per-table memo (the `skewChecked` pattern); probe now calls `reconstruct.GeneratedPKColumn` (no inline divergence risk); caveat states the whole truth (ON UPDATE parent reversal still emitted; descendants through the skipped edge not visited); gate hoisted above the `checkKeyChain` probes — after `cascadedHere = true`, the ordering the excluded-child test pins — so a skipped edge emits no redundant `keychain:` caveat; backstop message de-stuttered; "most commonly" hedge; docs and `Options.PKMetas`/`BaselineProvider` docs no longer overclaim (snapshot required; backstop only fires when a baseline covers the table; `ErrGeneratedPK` wrap contract documented on the interface).

**Refuted, with the reasoning recorded**: removing the sentinel backstop (defense for probe-less callers — the wired-callers-make-it-unreachable argument couples every future caller to the probe); the "stale latest-snapshot" corruption scenario (a conversion to system versioning AFTER a recovered delete leaves the polluted events outside the `[rootTS − lookback, rootTS]` scan window); the console bundle's load-once resolver (pre-existing, bundle-wide behavior, out of scope); a warning when the probe is nil (would cry wolf for every snapshot-less deployment — and with no snapshot the indexer captured nothing to scan anyway).

nethalo:
Toolkit review round (4 specialized agents) applied in 6903e0a — the test-analyzer's findings were the load-bearing ones and all are closed:

- **Real-provider sentinel now pinned** (`errors.Is` assertion in the #1266 provider test) — reverting the classification previously stayed green while production refusals fell back to `baselinefail:` and Phase-1 scanned the polluted candidates.
- **CLI wiring pinned by an integration test** and mutation-checked LIVE: deleting the CLI's `PKMetas:` line fails it (children synthesized, caveat gone); restored, it passes. Console/MCP remain one-line siblings of the same pattern.
- **Both hoists panic-observable** (DELETE fixture arms `checkKeyChain`; new ON UPDATE root test also pins the after-`cascadedHere` placement by asserting the parent reversal still lands in `KeyUpdateParents`).
- **Backstop memoizes its verdict** (silent-failure finding): no repeated provider scans per parent key, no redundant `keychain:` caveat in backstop-only mode.
- **Sentinel without stutter**: `GeneratedPKRefusalError` (the `MissingPKColumnError` `Is()` pattern) replaces `%w`; `fullTableGeneratedPKRefusal` now classifies too, and the `ErrGeneratedPK` doc states exactly who carries it (comment-analyzer's "global truth, locally true" catch).

Issue #1126, opened 2026-07-25
verify: legacy indexes and fresh indexes produce misleading recover-input verdicts


nethalo:
Follow-ups from the review of PR #1115 (`verify --check recover`). The critical false-MISMATCH path from that review is being fixed in the PR; these two are narrower cases of the same theme — a verdict or a reason string that asserts more than the data supports.

## 1. A legacy index turns a real historical DDL into a conclusive MISMATCH

`internal/verify/recover_inputs.go:462-470` uses `prevVer != curVer` as the "this is DDL, not corruption" escape hatch. But `query.ResultRow.SchemaVersion` is documented as *"snapshot_id at index time; 0 for pre-migration data"* (`internal/query/query.go:233`).

On an index whose rows predate that column, every event carries `0`, so `prevVer == curVer` always holds, and an ADD/DROP COLUMN inside the window yields:

> `"qty" is present in one image and absent from the other`

as a **MISMATCH**, on an index that is fine.

Fix: treat a `0` on either side as unknown-epoch → `unresolved`, the same direction the version-skew branch already takes.

## 2. Fresh/short-lived index: the gap reason string asserts something false

`internal/indexer/schema.go:76-80` — partitions span from the current hour *forward*. On an index younger than `--lookback` (default `30d`), `buildPlan` classifies every pre-install hour as a gap, and `FetchMerged` returns a `GapError` reading:

> `query covers hours with no data (rotated and not archived)`

Those hours were never rotated; the index did not exist. `recover_inputs.go:106` prepends `"coverage gap in the window: "` and hands that to the operator, and `docs/verify.md`'s remedy ("shorten `--lookback`") is unactionable when *any* lookback exceeding the index age fails.

The verdict itself (`unproven`, not `mismatch`) is correct and the docs pre-empt the behaviour — the defect is the reason string. Clamping `Since` to the oldest live/archived hour is one option and the design already absorbs it (chains just start mid-history → `chainNoPredecessor`), but that is a design change rather than a wording fix.

## 3. Minor, same file

- `mismatches` (`recover_inputs.go:281`, appended at `:300`/`:310`/`:328`/`:361`) is unbounded but only `[0]` and `len()` are ever consumed at `:380-384`. On a systemically broken table with `--max-events 200000` that retains up to 200k formatted strings — a third memory term the PR's bound section does not name. `first string` + `count int` closes it.
[...]

Issue #679, opened 2026-07-01
console: Overview 'latest window' is a fixed 200-event count, not a time window


nethalo:
## Problem

The console Overview page (`/overview`) reads: *"What changed recently, and where — your starting point. **N delete(s)** in the latest window: the ones worth a look first."*

"Latest window" sounds like a time-based window (last hour, last day), but it's actually a fixed **row count**: `renderOverview()` (`internal/console/assets/app.js:591-604`) fetches `api("/api/events?limit=200&order=DESC")`, and `deletes`/`tables touched`/`Activity by table` are all computed from those 200 rows (`buildOverview`, `app.js:610-629`). On a busy source those 200 events might span seconds; on a quiet one, days — the copy doesn't say which, so it reads as "recent activity" without telling you how much history that actually covers.

## Related inconsistency

The `COVERAGE <earliest> → <latest>` line at the foot of the "Activity by table" panel (`app.js:678-680`) is meant to show what the window above spans, but it prefers `status.coverage` (`cov.earliest_event`/`cov.latest_event` from `/api/status`, `app.js:614,630-631`) whenever the status fetch succeeds — which is the **global** index coverage, not the 200-event window's actual bounds. It only happens to equal the window's real span when `status` fails to load (the fallback branch) or, coincidentally, when the whole index has fewer than 200 events (as it did in the report that surfaced this). On an index with more history than the window covers, this line would show the full index's date range while the stats above it only reflect the last 200 events — misleadingly implying the deletes/tables-touched numbers cover more history than they do.

## Proposal

- Reword the Overview subhead to name the actual scope, e.g. "N delete(s) in the last 200 events" (or compute and show the real timespan those 200 events span).
- Either scope the `COVERAGE` line to the fetched window's own `earliest`/`latest` event timestamps unconditionally (not `status.coverage`), or add a second, clearly-labeled line for "full index coverage" vs. "this window's span" so the two aren't conflated.

## Context

Surfaced 2026-06-30 while reviewing a fresh console install where total indexed events (143) happened to be under the 200-event limit, making the window and the full index coincide — the ambiguity wasn't visible until traced through the code.

PR #1307, opened 2026-08-10
console: every Overview tile states its own scope, over a server-side window aggregate

### LINKED ISSUE #1300: console: Overview tiles mix all-time and last-200-events scopes; compute aggregates server-side
## The tiles mix two denominators without saying so

A live Overview shows, in one row of identical tiles:

| tile | actual scope |
|---|---|
| `3121` changes indexed | the **whole index** |
| `53` deletes | the **last 200 events** |
| `9` tables touched | the **last 200 events** |
| `2026-08-09 17:47:32` most recent change | a point in time |

Three different scopes, one visual treatment. An operator reads "53 deletes" next to "3121 changes indexed" and reasonably concludes 53 deletes out of 3121 — when it is 53 out of the most recent 200, a ratio more than eight times worse. The page intro says "in the last 200 event(s)" once, in prose, above a card that then presents the number without that qualifier; and the Activity-by-table panel carries its own `WINDOW` footer, which is the only place any scope is stated at the point of use.

This is the kind of number an operator screenshots into an incident channel. It should not need the paragraph above it to be read correctly.

## Compute the aggregates server-side

The window-scoped tiles exist because the page derives them client-side from a 200-event fetch — the same fetch that makes the page slow. That is backwards on both counts: it pulls 700 kB of row images to compute four integers, and it forces the numbers to describe an arbitrary fetch window rather than a meaningful period.

An aggregate endpoint (counts by event type, distinct tables, over a stated period like the last hour or 24 h) would be a small indexed query, would let the tiles state a period the operator chose, and would remove the reason to fetch 200 full events on a page that only lists six of them.

Do this rather than caching the current payload: a cache would freeze the numbers on the one page whose whole purpose is "what changed recently", and would leave the underlying cost untouched everywhere else.

## Either way

Whatever the scope ends up being, **each tile must state it** — "last 24 h", "all time" — in the tile, not in a paragraph above the card.

Related: the fetch behind this page is separately slow for an unrelated reason (an S3 archive scan on a newest-N query).

nethalo:
Closes #1300

## The problem

Four visually identical Overview tiles, three different scopes: `3121 changes indexed` was the whole index, `53 deletes` and `9 tables touched` were computed client-side over the newest 200 fetched events. Side by side, that invites reading 53 as a share of 3121 when it is 53 of 200. The only scope statement on the page was one line of prose above the card.

## What changed

**Every tile states its own scope**, in the tile:

| tile | scope line |
|---|---|
| changes indexed | `all time · estimate` |
| deletes | `last 24 h` (+ ` · partial` when the window is not fully covered) |
| tables touched | `last 24 h` |
| most recent change | `point in time (UTC)` |

**New `GET /api/activity`** (`internal/console/activity_api.go`) — counts by event type, distinct tables touched, per-table breakdown, over a stated period. `?period=` accepts `1h`/`6h`/`24h`, default `24h`; anything else is a 400 naming the options. The page now fetches **8** events instead of 200, so it no longer pulls ~700 kB of before/after row images to derive four integers, and a small period picker sits beside the tiles.

No cache, per the issue: this is the one page whose purpose is "what changed recently".

## Honesty properties (the ones worth reviewing)

- **The aggregate reads live `binlog_events` only.** Aggregating Parquet would reinstate the DuckDB scan this removes. So the handler runs `query.Plan` over the window and returns `complete: false` + a note when hours have rotated into the archives. A narrower number under an *explicit* period would be strictly worse than the mislabelled one this issue started from.
- **Gap hours are deliberately not counted as undercounts.** Nothing retrievable lives in a gap, and folding them in would cry wolf on every index younger than the window (#1126) and on every lapsed future-partition horizon (whose events land in `p_future`, where this timestamp-filtered aggregate *does* count them). Data actually lost is the coverage card's continuity verdict, directly above the tiles.
- **`changes indexed` is `total_events_estimate`** — `information_schema.TABLE_ROWS`, an InnoDB estimate. The tile says so instead of sitting in the same type as three exact counts.
- **A failed aggregate renders `—`**, never a zero nobody measured; an unopened index connection is a 502, not a quiet `0`.
- **RBAC**: deny rules (startup floor ∪ session profile) are applied in the SQL, so a denied table contributes to neither the counts nor `top_tables` — a table name is exactly what a deny profile withholds elsewhere.
- **Not audited**, on purpose: counts and table names are metadata, the side of the line `ext/audit.go` draws, same as `/api/status` and `/api/coverage`.

## Cost, stated plainly

The issue calls this "a small indexed query"; it is cheaper than the old fetch but it is not index-served. `binlog_events` is `PARTITION BY RANGE (TO_SECONDS(event_timestamp))`, so the bounded, **unwrapped** predicate prunes to the window's partitions — but no index leads with `event_timestamp` and `event_type` is in none, so it is a scan of the pruned partitions plus a grouping pass. That is why the period allowlist stops at 24 h (~25 partitions): a week-long aggregate is a reporting query, not a tile that loads on every visit to the landing page.

## Drive-by fix

`GET /api/coverage` was registered in `server.go` but **never classified in `apiRoutePerms`** — so every policy-carrying (EE) session got a 403 on the Overview's own coverage card. It went unseen because `registeredAPIPatterns`, the completeness test's hand-maintained mirror, was missing it too. Both lists now carry `coverage` and `activity`.

## Tests

`test/console-e2e/console_e2e.mjs`: Scenario 15 rewritten (its #679/#684/#686 intent preserved — the window line must never fall back to `status.coverage.oldest`) plus a new live scenario that renders the real page against the running daemon. The fixture makes the two sources disagree on purpose — the events array holds 1 delete over 1 table, the aggregate says 17 over 5 — so a tile that recomputes from the events fails. Degraded arms covered: partial window, failed aggregate. **121/121 pass locally.**

### Mutations run (each broke production code, confirmed the named test fails, then reverted)

| # | mutation | caught by |
|---|---|---|
| M1 | drop the `event_timestamp >= ?` lower bound | `TestActivitySQLPrunesPartitions` |
| M2 | drop the deny-table clause from the builder | `TestActivitySQLExcludesDeniedTables` |
| M3 | drop the deny plumbing in the handler (`nil` instead of `opts.DenyTables`) | `TestActivityDenyTablesReachTheQuery` |
| M4 | never mark the window incomplete | `TestActivityArchivedWindowIsNotComplete` |
| M5 | `Tables` = length of the rendered top list | `TestActivityAggCapsRenderedTables` |
| M6 | route DELETEs into the `other` bucket | `TestActivityAPICounts` |
| M7 | count gap hours as archived | `TestArchivedHoursInWindow` |
| M8 | answer 200 with zeros on a nil index connection | `TestActivityUnopenedIndexIsNotZero` |
| M9 | unregister `/api/activity` from `apiRoutePerms` | `TestRouteTableClassifiesEveryRegisteredRoute` |
| M10 | unregister `/api/coverage` from `apiRoutePerms` | `TestRouteTableClassifiesEveryRegisteredRoute` |
| JM1 | remove the scope line from `ovStat` | 7 e2e assertions |
| JM2 | window footer falls back to `status.coverage.oldest` | 2 e2e assertions |
| JM3 | deletes tile recomputed from the fetched events | 2 e2e assertions |

`gofmt -l` clean on the touched files, `go vet ./...` clean, `go test ./internal/console/...` green, `node --check` clean.

The diff is confined to the Overview parts of `app.js` (`renderOverview`/`buildOverview`/`ovStat`) and adds no lines to `api.go`, to stay out of the way of the concurrent Events-view work.


nethalo:
Follow-up commit `3e7431a` after review:

- **Permission tier corrected.** `/api/activity` was classified `status:read`; it is now `query:execute`. It aggregates indexed row data and reports which tables changed and how often — and the handler has to apply the profile's `DenyTables` to be safe, which is what disqualifies it from the health-read floor. Costs a caller nothing: the page rendering it also reads `/api/events`. The completeness tests only prove a route is *classified*, so the tier is now pinned by `TestActivityRequiresQueryPermission` (mutation M11: downgrading it back to `status:read` fails that test). `/api/coverage` stays at `status:read` — timestamps and verdicts, and it self-gates its table-name half on `sessionRestricted`.
- **Period picker staleness.** It re-rendered without bumping `viewGen`, so two quick clicks left two renders in flight and whichever fetch landed last painted — tiles showing 6 h of data under a "last 24 h" label, i.e. this PR's own bug through the control added for it. Now bumps `viewGen` first, the same guard `renderRoute` uses.

Re-verified: `go test ./internal/console/...` green, `go vet ./...` clean, console e2e **121/121**.

Issue #1353, opened 2026-08-15
console: Events default browse must paint in seconds on an archive-heavy source, with a visible loading state


nethalo:
Sibling of #1352 (Overview). Same deployment shape, same two problems, different mechanics.

## Symptom

On a source with a short live retain — nearly all history rotated into S3 Parquet archives — opening `/events` leaves the operator staring at an empty list for tens of seconds. The only sign anything is happening is the `…` placeholder in the result-count chip; the list area itself is blank, which reads as "no events" or "broken", not "working". This is the flagship investigation surface: it is the page an operator opens mid-incident.

## Mechanics

`renderEvents` paints the frame (search bar, filters, result bar) synchronously — good — but `runEventsQuery` then blocks the whole list on a single `GET /api/events` fetch. The default browse carries no `since`/`until`, and the console's cross-source read is archive-aware, so the fetch can end up opening Parquet archives (S3 download + DuckDB) to answer a question whose result — the newest N events, `order=DESC` — is almost always satisfiable from the live index alone: archives only ever hold partitions OLDER than the live rotation boundary, so they cannot contribute anything to a newest-first page that the live index already fills.

Worth verifying with a profile before implementing, but the suspects are archive auto-discovery + the merged read running on an unbounded window for a page the live index could serve in milliseconds.

## Goal

**p95 under 5 seconds** — ideally well under — **for the default browse's first page** on an archive-heavy source, and a loading state that is impossible to miss.

Candidate directions (pick after measuring, not before):

- **Newest-first short-circuit**: when the live index fills the requested page with rows newer than the oldest live partition, skip the archive tier for that page entirely. Archives get consulted on older pages (`before` cursor past the rotation boundary) and time-filtered queries — where they are genuinely needed. The #1319 divergence check only applies where the two tiers can overlap, which a fully-live page does not.
- **Answer live first, reconcile after**: paint the live-index page immediately, and if the archive tier must run, let it revise/append with a visible note — never silently.
- Whatever ships must keep the honesty contract intact: the `warnings` array (#1311/#1321) still says when a result excluded archives, and a page that DID skip archives because the live index filled it is a correctness-preserving optimization, not a scope reduction — the distinction should be auditable in the response.

## Loading-state visibility

- The list area needs a real loading state from the moment the fetch starts: skeleton rows (preferred) or a centered spinner with label — a blank list must never be the "busy" rendering.
[...]

Issue #1325, opened 2026-08-13
merge: index/archive divergence is logged but never reaches the console or MCP response warnings


nethalo:
Surfaced by review of #1319, which added a `slog.Warn` when two copies of one `event_id` disagree between the index and a Parquet archive.

`slog.Warn` is the only output, and for two of the three surfaces that call `MergeResults` the operator can never see it:

- **Console** — the operator is in a browser; the warning goes to the `bintrail-console watch` container's stderr.
- **MCP** — same shape, and the response has a `Warnings` list built for exactly this.
- **Shim** — a MySQL-protocol client gets a resultset and nothing else. (Consistent with the shim's existing `stale_baseline` handling, which is already a documented deferred gap.)

The codebase states the rule for this class in CLAUDE.md, about capture gaps:

> `CaptureGapStatus` (the finding) is split from `CheckCaptureGap` (the strict/allow policy) so an OVERRIDDEN gap still lands in the response `Warnings` — `found: true` with zero warnings over a permanently-lost window is the failure mode this exists to prevent.

Divergence is the same class and arguably sharper: on the recover path the chosen `row_before` becomes the `SET` clause of generated reversal SQL. A silent coin-flip between two disagreeing before-images is a data-integrity event, and stderr-only under-weights it.

The blocker is that `MergeResults` returns only `[]ResultRow`. Suggested shape, mirroring the `CaptureGapStatus` split:

```go
func MergeResultsReport(rows []ResultRow, limit int, order string) ([]ResultRow, int) // rows, divergedCount
```

Keep `MergeResults` as a thin wrapper so no call site breaks, then thread the count into the console and MCP `Warnings` (`event_divergence: N duplicate events disagreed between the index and the archive`). Keep the `slog.Warn` — it is the right channel for the CLI.

Issue #1311, opened 2026-08-11
console: a profiled session silently reads live-MySQL only, and the gap warning misattributes the cause


nethalo:
## What

A console session that carries an RBAC data profile reads live MySQL only — `fetchRestricted` (`internal/console/session_profile.go`) forces `NoArchive` because the Parquet archive path does not run the redaction pass:

```go
noArchive := b.noArchive || sessionRestricted(r)
```

That is the right call and it fails in the safe direction: the session is shown less, never unredacted more. The problem is that the result does not say so, and the two surfaces it covers — `GET /api/events` and `POST /api/recover` (the two `fetchRestricted` call sites) — are exactly where an operator forms a belief about what happened to their data.

Hours that rotated out of live MySQL still exist in the archives; the profiled session simply does not read them. What the operator sees is a short or empty result, and the natural reading of that is "nothing happened in this window", not "this window is in storage your session does not open".

The same silence applies to the process-global `--no-archive` (and `--profile`, which implies it), but that one at least follows a flag the operator typed.

## What already exists, and what is missing

Both responses already carry the channel: `eventsResponse.Warnings` and `recoverResponse.Warnings` are populated today from `gapWarnings(plan)`. Two gaps remain.

**1. No time filter, no warning at all.** `buildOptions` sets `Since`/`Until` only from request params, and the planner runs only when a time range and DBName are set. The default browse — newest events, no filter — produces no plan, so `gapWarnings` returns nil. That is precisely the view where the silence bites hardest.

**2. With a time filter, the warning fires but names the wrong cause.** Under `NoArchive` the planner deliberately counts archived-only hours as gaps (documented in `FetchMergedOptions.NoArchive`), so the message rendered is:

> query covers hours with no data (rotated and not archived): … – …

For a profiled session that is false in its second clause: those hours *were* archived. An operator who reads it goes looking for a broken rotation or a lost upload, and finds a healthy archive tier. Misattributing a cause is worse than saying nothing, because it sends someone to fix what is not broken.

## Options

1. **Say it.** Emit a distinct warning whenever the fetch ran with archives off, naming the reason (session data profile vs `--no-archive`/`--profile` at startup), and suppress or reword the gap message in that case so the archived-only hours are not reported as missing archives. No permission or data change; closes the misreading with the channel that is already there. `internal/console/coverage_warnings.go` is the established shape for this.
[...]

Issue #1120, opened 2026-07-25
recovery: raw identifiers and PK values in `--` comments break out of the comment on a newline


nethalo:
Surfaced by the review of PR #1110, which found it in the new opt-in `AUTO_INCREMENT` block — but the more reachable instance is **pre-existing on `main`** and ships today.

## The defect

`internal/recovery/recovery.go:288` writes the per-statement header comment with raw values:

```go
fmt.Fprintf(&body, "-- [%d] reverse %s on %s.%s pk=%s at %s%s\n", …)
```

`event.EscapePKValue` escapes only `\` and `|` — not `\r`/`\n`. A `--` comment ends at the first newline. So a **PK value containing a newline** ends the comment and puts whatever follows it into the script as executable SQL, *inside* the `BEGIN`/`COMMIT`.

A newline inside a VARCHAR primary key is ordinary data — no exotic schema required. The value reaches the generator unfiltered: `internal/parser/parser.go` stores the row image verbatim and the indexer persists it as-is.

The same class exists for identifiers. `QuoteName` escapes backticks but not newlines, and MySQL permits any BMP character except U+0000 in a backtick-quoted identifier. A table named `orders\nDROP TABLE users; -- ` breaks out of any `--` line carrying it. PR #1110's `AUTO_INCREMENT` block is the sharpest case, because there the comment prefix is the *entire* safety mechanism and the block sits after `COMMIT`, so the injected statements run standalone:

```
-- `db`.`orders
DROP TABLE users; -- `:
--   SELECT IFNULL(MAX(`<auto_increment_column>`), 0) + 1 FROM `db`.`orders
DROP TABLE users; -- `;
--   ALTER TABLE `db`.`orders
DROP TABLE users; -- ` AUTO_INCREMENT = <N>;
```

## Why it matters here specifically

`recover` never executes — the emitted text **is** the product. An operator pastes it into a production restore, usually under incident pressure. Every other correctness guard in this package assumes the script says what it appears to say.

## Fix

One shared helper (`sanitizeForComment`: replace `\r`/`\n` with a space, or refuse outright) applied at every site that interpolates schema, table, or PK values into a `--` comment. Fix all sites together — patching only the newest one leaves the easier vector open.

[...]

PR #1367, opened 2026-08-15
console: render the archive-elision record as an info note, not a warning

### LINKED ISSUE #1365: console: the archive-elision note renders as an alarm — a benign audit fact styled and worded like an incident
Follow-up to #1353 / PR #1357, from watching a real user read the page.

The archive-elision notice on Events renders as an **amber ⚠ warning banner** — the same component every response warning gets — while its own text says "Nothing is missing from this page." A benign, correct-by-construction fact ("the live index answered this page; the archives only hold older events") is styled like an incident, and the phrasing "were not searched … because they could not change it" reads as jargon. The user's read on first contact: *something bad happened*. That is the opposite of what the notice exists to convey.

## Root cause

The console renders every entry of the response `warnings` array through one alert component. The array now mixes severities: real cautionary facts (coverage gaps, a session's archive exclusion, divergence findings) and purely informational audit notes (#1357's elision). One visual register for both means either the info notes look like incidents, or — if we mute the component — the incidents stop looking like incidents.

## Fix

1. **Split severity.** Either type the notices on the wire (`warnings` stays for cautionary facts; a new `notes`/`info` list for benign audit notes — additive, existing consumers unaffected) or classify UI-side; wire-side is the honest split since MCP and the API share the shape. The #1357 elision note moves to info. Gap warnings, profile-exclusion notices, and divergence findings stay warnings.
2. **Render info quietly.** A muted single-line note (no ⚠, no amber — the theme's ink-muted register, maybe collapsed behind the result-count line), because its job is auditability, not attention.
3. **Rewrite the copy in plain words.** Something like: "This page was answered from the live index. Archived events are all older than this page reaches — paging further back or filtering by time searches them too." Drop "because they could not change it".
4. **Keep the API contract.** The auditability requirement from #1353 stands: the response must still say the archives were skipped and why. This issue is about severity and wording, not about removing the fact. (If after the severity split the maintainer prefers hiding the info note from the default UI entirely and leaving it API-only, that's a one-line call in the PR — the wire fact must survive either way.)

## Guard

console-e2e: the elision note must render in the info register (no alert classes), real gap/profile warnings must keep the warning register, and the MCP/API responses must keep carrying the elision fact.

nethalo:
Fixes #1365.

## Summary

The #1357 archive-elision notice — "the live index answered this page; the archives only hold older events" — rendered through the same amber ⚠ alert component as every response warning, while its own text said nothing is missing. A benign, correct-by-construction audit fact read as an incident.

**Severity model (split on the wire).** The response `warnings` array now carries only cautionary facts — coverage gaps, a session's archive exclusion (#1311/#1321), merge divergence findings (#1325). A new, additive `notes` list carries benign informational audit facts; the #1353 elision record moves there. The split is wire-side because the API and the console UI share the shape: classifying UI-side would leave every other client (curl, an agent) with one undifferentiated list. `responseAdvisories` in `internal/console/coverage_warnings.go` is the single assembly point for both lists on `/api/events` and `/api/recover` (all recover response paths, cascade included), so the split cannot drift per handler. Consumers that ignore `notes` see the unchanged warnings contract.

**Quiet rendering.** The console renders `notes` in a muted single-line register (`--ink-3`, no ⚠ icon, no amber, no border) directly under the result-count line on Events, and in its own container on Recover — both the undo response and the shared preview, which keeps re-rendering rather than clearing it. Real warnings keep the alert register unchanged. (The maintainer option of hiding the note from the UI entirely was not taken: the muted register keeps the audit fact one glance away without alarming anyone.)

**Copy rewrite.** Before:

> Archived (rotated) hours were not searched for this page because they could not change it: the live index filled the page with the newest matching events, and every archived hour is older than what the live index still holds. Nothing is missing from this page. Paging further back, or filtering to a time range that reaches archived hours, does search the archives.

After:

> This page was answered from the live index; the registered archives were not read. Every archived event is older than this page reaches, so nothing is missing here. Paging further back, or filtering to a time range that reaches archived hours, searches the archives too.

The wire string still carries the #1353 audit fact (archives skipped, why that loses nothing, how to reach them) and stays flag-free per the MCP rule — a unit test pins both.

**MCP decision: keep the single `warnings` list, unchanged.** No MCP tool ever produces the elision record: the reconstruct tool's fetch is ASC and unbounded, which the newest-first short-circuit never takes (the flag is structurally false and discarded), and the query/recover tools read archive sources directly with no short-circuit path. There is nothing on that surface to reclassify, and inventing an empty `notes` field there would be dead wire. The structurally-false call sites (`internal/mcptools/reconstruct.go`, `internal/console/reconstruct.go`) now document that if a future fetch shape ever makes the flag reachable, the record belongs in an info notes list, never in warnings.

**Auditability stands.** This is severity + wording only: the response still carries the elision fact on every short-circuited page, and silencing it fails tests (see below).

## Testing

- **Unit** (`internal/console/coverage_warnings_test.go`): `TestResponseAdvisoriesSeveritySplit` drives all four facts through the real assembly — gap, exclusion and divergence land in `warnings`, the elision in `notes`, and the lists never cross; `TestArchiveElisionNote` pins the rewritten copy (audit fact present, old jargon gone, flag-free) and the quiet no-elision path.
- **Integration** (MySQL 127.0.0.1:13306): the #1357 browse short-circuit test now asserts the fact in `notes` with `warnings` clean on the fast path, no elision claimed on the slow path, and the profiled session's exclusion staying a warning with zero notes; the #1321 wiring test additionally pins that `/api/recover` never demotes the exclusion to a note; the #1325 divergence wiring test passes unchanged (the finding stays a warning). Full `internal/console` integration package green.
- **Mutation checks** (performed, then reverted): routing the elision back into `warnings` fails `TestResponseAdvisoriesSeveritySplit` and the short-circuit integration test; silencing the note fails `TestArchiveElisionNote`, `TestResponseAdvisoriesSeveritySplit`, and the integration test's auditability assertion.
- **console-e2e**: `run.sh` now builds a second archive-fixture index (real `bintrail rotate` archiving six seeded hours to Parquet, plus a manufactured rotated-not-archived gap hour), with fixture premises asserted loudly. New scenario 17 (12 PASS lines) proves on the wire and in the DOM: the elision rides `notes` with `warnings` clean, renders as one muted line under the result-count bar (no alert classes, no svg icon, transparent background, no border), and a real coverage-gap warning on the same fixture keeps the ⚠/amber/border alert register with no elision claimed. Full suite: **179/179 passed**, no silent skips.
- `go build ./...`, `go test ./... -count=1` (51 packages ok), `go vet ./...` and `go vet -tags integration ./...`, `gofmt -l` clean on touched files, `node --check` on `app.js` and `console_e2e.mjs`.


nethalo:
Review round applied in b254557 (one consolidated fix round, all findings verified against the code first):

1. **Recover notes wiring pinned positively** — the only recover-side `notes` assertion was a negative holding vacuously on nil (the #1120 class). New subtest on the short-circuit fixture: `POST /api/recover {"schema":"shop","table":"orders","limit":50}` (DESC + filled limit → elides) asserts the note in `notes`, the limit-truncation warning present, and the elision absent from `warnings`. Mutation-verified: deleting `Notes: notes` at the plain write site fails it, and reverting the split only in `handleRecover` fails it on both assertions. The divergence wiring test now decodes `notes` on events + recover with `len==0` asserts; the cascade write site carries a comment naming its untested-notes caveat.
2. **Discard-site comments corrected + guarded** — both reconstruct fetches DO carry a limit (`reconstructMaxEvents+1` / `MaxReconstructEvents+1`); ASC is the single leg preventing the short-circuit. Comments now say so, and the flag is captured with a loud `slog.Warn` instead of `_`, so a future DESC flip can't silently swallow the audit fact.
3. **Docs** — the `POST /api/recover` contract row now includes `notes`, and the "notes is the informational sibling of warnings" statement is scoped to `/api/events` + `/api/recover` explicitly, with a warning that `/api/activity`'s pre-existing `notes` is cautionary (explains `complete: false`).
4. **Verbatim copy pins** — `TestArchiveElisionNote` asserts `got != want` on the exact string per surface, on top of the semantic fragment/flag-free checks.
5. **Per-surface wording** — recover gets its own note (a reversal has a window and a limit, not pages): *"This reversal was built from the live index; the registered archives were not read. Every archived event is older than the window this request reaches, so nothing is missing here. A time range that reaches archived hours, or a higher limit, searches the archives too."* `responseAdvisories` takes the surface's string; both pinned verbatim.
6. Events now renders warnings **above** notes (alerts first, both views agree).
7. `runEventsQuery`'s catch clears both advisory containers alongside the rows.
8. Rewrapped the ragged comment in `session_profile.go`.
9. `run.sh` computes the fixture anchor hour once and passes it to both node scripts (independent clock reads could straddle an hour boundary).
10. `responseAdvisories`' doc states the lists' exclusivity is enforced by convention — handlers must not hand-assemble advisory lists.

Evidence: `go build`/`go vet` (+`-tags integration`) clean, unit suite green (51 packages), `internal/console` + `internal/mcptools` integration packages green against 127.0.0.1:13306 (new recover subtest PASS), console-e2e **179/179**, `node --check` and `gofmt` clean on touched files. All four mutation checks (elision→warnings, note silenced, recover `Notes` deleted, recover-only split revert) fail a test; all reverted.

